PR #5, opened 2021-06-07
Add new ResolverWithOpts

jdolitsky:
This will allow us to add new options to Resolver without breaking API in the future.

In the near term, this allows us to set a custom User-Agent header:
```
headers := http.Header{}
headers.Set( "User-Agent", "my-user-agent")
resolver := ResolverWithOpts(WithResolverHeaders(headers))
```

Resolves https://github.com/oras-project/oras/issues/240


deitch (commented):
A few things I noticed. 

* We used to pass a `context.Context` to `Resolver()`. Do we not need it anymore?
* In other areas, functions to set options passed as variadic parameters always were named `With*`, which would make these `WithResolverOptClient(client *http.Client)` or `WithResolverOptPlainHTTP()`. Should we be consistent in the naming?
* whether or not to use plain http is a boolean, which defaults to false. If the only thing we will do is set it to true, then it probably should just be `ResolverOptPlainHTTP()` (or `WithResolverOptPlainHTTP()`, not parameter required

Right now, just a comment.

jdolitsky:
@deitch -

> We used to pass a context.Context to Resolver(). Do we not need it anymore?

It's actually not needed. The original implementation was just throwing it away (`Resolver(_ context.Context, ...`). The beauty of this is that if it is needed later, we can add a new `WithResolverContext(context.Context)` option.

> In other areas, functions to set options passed as variadic parameters always were named With*, which would make these WithResolverOptClient(client *http.Client) or WithResolverOptPlainHTTP(). Should we be consistent in the naming?

Yes, agree. Done.

> whether or not to use plain http is a boolean, which defaults to false. If the only thing we will do is set it to true, then it probably should just be ResolverOptPlainHTTP() (or WithResolverOptPlainHTTP(), not parameter required

Done.

deitch (approved):

PR #35, opened 2021-10-20
Feat: Adding discoverer & example copy command 

juliusl:
This is the minimum viable product for the recursive copy/discover. I restored some files that were removed for `pkg/oras/copy.go` namely pull.go, since it has the semantics I'm looking for in this implementation, and added back the ProviderIngester interface.

Signed-off-by: Julius Liu <juliusl@microsoft.com>


juliusl:
Pinging @mnltejaswini

deitch:
Is the usage of `type ProvideIngester` and its embedded types of `content.Provider` and `content.Ingester`, as well as the `func Pull()`, necessary?

We worked pretty hard to unify it, make it much simpler and easier to understand. It actually was one of our key goals in the major rewrite: make this simpler to understand and write plugins for.

My understanding from talking to the containerd folks is that the existence of `remotes.Resolver`/`Pusher`/`Fetcher` and `content.Provider`/`Ingester` is largely due to legacy, i.e. 2 different sources that got merged. They also want to get rid of it and have a single much-less-confusing interface, likely `remotes.Resolver` (potentially with some additional features added).

What could we not do with the existing interface that we would need to add these?

pkg/target/docker/discoverer.go
```diff
@@ -0,0 +1,375 @@
...
+
+	resp, err := ctxhttp.Do(ctx, client, req)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to do request")
+	}
+	log.G(ctx).WithFields(responseFields(resp)).Debug("fetch response received")
+	return resp, nil
+}
+
+func (r *request) doWithRetries(ctx context.Context, responses []*http.Response) (*http.Response, error) {
```
mnltejaswini, line 283: This method is used for other requests. Do you think you can separate it in its own file like utils.go

pkg/target/docker/discoverer.go
```diff
@@ -0,0 +1,375 @@
...
+	return r.host.Scheme + "://" + r.host.Host + r.path
+}
+
+func requestFields(req *http.Request) logrus.Fields {
+	fields := map[string]interface{}{
+		"request.method": req.Method,
+	}
+	for k, vals := range req.Header {
+		k = strings.ToLower(k)
+		if k == "authorization" {
```
mnltejaswini, line 344: You may still want to find all the headers that were set and for secret value headers, the common pattern is using some form of scrubbed string like *****

pkg/target/docker/discoverer.go
```diff
@@ -0,0 +1,375 @@
...
+		r.path = r.path[:i+1]
+		q, err = url.ParseQuery(r.path[i+1:])
+		if err != nil {
+			return
+		}
+	} else {
+		r.path = r.path + "?"
+		q = url.Values{}
+	}
+	q.Add("ns", ns)
```
mnltejaswini, line 184: Curious as to why we need this ns?

pkg/target/docker/discoverer.go
```diff
@@ -0,0 +1,375 @@
...
+	hosts      docker.RegistryHosts
+	header     http.Header
+	refspec    reference.Spec
+	reference  string
+	repository string
+	tracker    docker.StatusTracker
+	remotes.Resolver
+}
+
+var localhostRegex = regexp.MustCompile(`(?:^localhost$)|(?:^localhost:\\d{0,5}$)`)
```
mnltejaswini, line 70: Will this include 127.0.0.1

juliusl:
@deitch  

The store interface only has Fetch and Push. For my purposes, I wanted something lower level that gives me a Reader and Writer. I will elaborate on this. 

The Pull() function returns the artifacts that were pulled and the parent of those artifacts in a single call, while the Copy() function only returns the parent. So I added the Pull() function back since it serves the exact semantics I was looking for.

Initially I intended to just use the oras.Copy function, but it appears there are issues with that. For context I wrote a copy command that does a recursive copy using the referrers api. This code works. When I initially started reverse integrating the code back to oras-go I had hoped to use the oras.Copy function and swap it out with the oras.Pull function. I attempted that but it runs into a ErrNotFound issue. What I'm observing is instead of talking to the Registry first and fetching the content as I expected. The code talks to the memory store first in the hybrid store.

```
> oras.land/oras-go/pkg/oras.transferContent.func2.1() /home/juliusl/go/src/github.com/juliusl/oras-go/pkg/oras/copy.go:122 (hits goroutine(19):1 total:1) (PC: 0xb08475)
   117:                                 return nil, nil
   118:                         }
   119:                         defer cw.Close()
   120:
   121:                         rc, err := f.Fetch(ctx, desc)
=> 122:                         if err != nil {
   123:                                 return nil, err
   124:                         }
   125:                         defer rc.Close()
   126:                         return nil, content.Copy(ctx, cw, rc, desc.Size, desc.Digest)
   127:                 })
(dlv) c
> oras.land/oras-go/pkg/content.(*Memory).Fetch() /home/juliusl/go/src/github.com/juliusl/oras-go/pkg/content/memory.go:93 (hits goroutine(19):1 total:1) (PC: 0xaf8997)
    88: }
    89:
    90: // Fetch get an io.ReadCloser for the specific content
    91: func (s *Memory) Fetch(ctx context.Context, desc ocispec.Descriptor) (io.ReadCloser, error) {
    92:         _, content, ok := s.Get(desc)
=>  93:         if !ok { 
    94:                 return nil, ErrNotFound
    95:         }
    96:         return ioutil.NopCloser(bytes.NewReader(content)), nil
    97: }
    98:
```

So, I ran into this issue initially when I wrote the copy command, and to fix it I needed to set the ProviderIngester store so that I could bypass the hybrid store that gets created by default. From my observations the hybrid store had too many side effects, and fighting with it was causing a bunch of friction. When I bypassed it, everything started to work perfectly, which I confirmed on the registry side logs and client side content. 

I found that the OCI store offered the best resiliency to heterogenous content. This is important because a recursive copy of artifacts involves heterogenous media, and I need the flexibility to treat content opaquely in order to minimize errors when transferring deeply nested trees of heterogenous references. 

In the latest oras-go the copy option analagous to the pull option has been updated to require a content.Store interface instead. The content.Store requires the push and fetch command. I had hoped to just use the OCI store as the content.Store but it appears that the OCI store does not implement that interface. The Push method has been separated into a different struct, so OCI type on it's own does not have the Push function. 

Adding a push function to the main OCI store wouldn't make sense because in order to create a pusher you need a reference. In my implementation a) I don't know the reference early enough in order to do that b) taking a step back I don't ever plan on pushing to the OCI store. 

In order to minimize risk and deliver working code, I chose to put back the IngesterProvider interface back, since all I needed to do to get the OCI store to conform to that interface was restoring code that already existed prior to the refactor. And the Pull function I put back since it offered the semantics I was looking for out of the box, and only requires the ProviderIngester interface to work. Since an OCI store only takes a single method to turn into a ProviderIngester that seemed like the logical choice.

pkg/target/docker/discoverer.go
```diff
@@ -0,0 +1,375 @@
...
+		return nil, errdefs.NotFound(errors.New("no discover hosts"))
+	}
+
+	ctx, err = docker.ContextWithRepositoryScope(ctx, d.refspec, false)
+	if err != nil {
+		return nil, err
+	}
+
+	v := url.Values{}
+	v.Set("artifactType", artifactType)
```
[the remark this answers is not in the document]
juliusl: Technically yes, but adding it without an artifactType is valid as well. This lets me avoid unesscary branching logic. the constraint on supporting multiple artifactTypes is on the registry side, so we could answer that by taking a look at the specification. I don't recall it mentioning that multiple artifactTypes are allowed but I might have overlooked that.

pkg/target/docker/discoverer.go
```diff
@@ -0,0 +1,375 @@
...
+		}
+	}
+
+	return hosts, nil
+}
+
+func (d *dockerDiscoverer) Discover(ctx context.Context, desc ocispec.Descriptor, artifactType string) ([]artifactspec.Descriptor, error) {
+	ctx = log.WithLogger(ctx, log.G(ctx).WithField("digest", desc.Digest))
+
+	hosts, err := d.filterHosts(docker.HostCapabilityResolve)
```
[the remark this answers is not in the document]
juliusl: The specification states that the registry MUST implement pagination, but makes no mention that clients are required to implement the pagination in order to make use of the referrers api. I feel we should add pagination incrementally in a follow up PR as I do not have the bandwidth to get that into this one.

pkg/target/docker/discoverer.go
juliusl: I'm not sure actually, it was a part of the original code that I ported from the containerd fork. I assumed that it was used for something important and didn't have enough context to remove it.

pkg/target/docker/discoverer.go
juliusl: Fixed

deitch:
> The Pull() function returns the artifacts that were pulled and the parent of those artifacts in a single call, while the Copy() function only returns the parent. So I added the Pull() function back since it serves the exact semantics I was looking for.

Correct. We thought about that as well, but it turns out to be too specific and inflexible. It assumes a lot about the structure of what you return, and doesn't leave a whole lot of room for other structures and flows.

Granted, by returning only the root descriptor, you need to take more steps to walk down the tree and get more objects, but you can handle any structure at all.

> The Pull() function returns the artifacts that were pulled and the parent of those artifacts in a single call, while the Copy() function only returns the parent. So I added the Pull() function back since it serves the exact semantics I was looking for.

Is that more correct than using the returned descriptor to find what you need? Or add the handlers in the process?

I can see a use case you are describing, i.e. it might be common to need all of the descriptors pulled down, so let's make a common `CopyOpt` that does just that, but that is a different requirement.

> When I initially started reverse integrating the code back to oras-go I had hoped to use the oras.Copy function and swap it out with the oras.Pull function. I attempted that but it runs into a ErrNotFound issue. What I'm observing is instead of talking to the Registry first and fetching the content as I expected. The code talks to the memory store first in the hybrid store.

You lost me here. Maybe the use case isn't clear to me?

> bypass the hybrid store

I am not madly in love with that hybrid store. But it serves a valid purpose (which long predates our `Copy()` API). 

When you copy from `A` to `B` (or pull and push, whatever), you walk the tree down in `A`: root descriptor to manifest to layers, in the simplest example. You have to, as that is how it exposes itself.

But you cannot load into `B` that way. For many targets of `B`, if you try to load up the manifest before the layers are loaded, you will get an error. So manifests and indexes and such are saved in memory, while layers are sent directly to `B`. When all is done, it can go back and load manifests from the memory to `B` in reverse order.

In theory, you could just reverse order _everything_, but then you would need to be storing all of the big layers, either locally on disk or in memory; that isn't efficient and is quite brittle.

I still don't quite understand why it was necessary to return the `Pull` and the `content.*` interfaces, probably because I didn't understand some of the comments above and, perhaps more basically, the use case.

We don't object to `content.Store` per se; we are trying rather hard to:

* simplify this - it was quite hard for people to wrap their heads around it with both `remotes.*` and `content.*`
* align with where the containerd folks are going anyways
* reduce dependency on containerd code wherever possible
* make it easier to support all sorts of `target`, including registry and OCI layout and etc.

All of that drove us towards, no one is pulling or pushing, they are copying from one to the other; the very idea that the registry is somehow "there" and where I am running it is "here", leading to push/pull semantics doesn't really make sense in a multi-target artifact world. 

At the same time, we did recognize that some things just naturally come with a `content.Store` interface, or might need a `ReaderAt`, so we created [ProviderWrapper](https://pkg.go.dev/oras.land/oras-go@main/pkg/oras#ProviderWrapper), and updated the various `content` structures to support `target.Target`, e.g. [OCI](https://pkg.go.dev/oras.land/oras-go@v0.4.1-0.20211014081142-293f55c08a7b/pkg/content#OCI), file, memory, etc.

juliusl:
@deitch  Here is a hack-doc I made for the design of this PR. It's a bit more organized then my general ramblings.

https://hackmd.io/@juliusl/Hk5hBec8K

juliusl:
> @juliusl - Is this similar to RX semantics of processing artifacts as you see them? 

@sajayantony actually I draw a lot of my design philosophy from RX semantics, I've been trying to write a library on the side that allows an architect to "master data in motion"

sajayantony:
Added issue for tracking copy to local targets - https://github.com/oras-project/oras-go/issues/43

deitch:
Having finally got my head around most (but definitely not all) of the problem to be solved, I do have a few comments.

Is it correct to state that _everything_ that I want to copy - from any "there" to any "there" - is essentially through a `remote.Resolver`, i.e. something that has a `Pusher` and a `Fetcher`? And that the `remotes.*` is the tx/rx modules in your RF analogy, while the `content.*` is the local connection on each end between the tx/rx module and its local buffers (or whatever it connects to locally)?

If I did understand it correctly, then I am somewhat confused as to why we are exposing things like `ReaderAt()` on the various content stores. Isn't that an internal thing? As far as. I can tell, it only is used in [Object.Move()](https://github.com/oras-project/oras-go/pull/35/files#diff-ae6515f1df584f55610c934588ddf0c64fcfe5e38a9178023749eca95a46caf8R134), which "moves" something from an `Artifact` to a `Target`.

This confuses me. Do I not always move things from one `Target` to another `Target`? Isn't that the comparative analogy to the rx/tx channel, and the channel itself has no clue what is inside the transmission, which only makes sense to the local side of each tx or rx?

Let's take this a step higher up (btw, that hackmd doc was helpful). Can we list out what activities a user would need to be able to do, including what they can do today? Here is what I know of, including some that I think are correct from the code in this, but I am not sure about those:

* copy an image from one `Target` to another `Target` (current)
* copy any generic artifact type from one `Target` to another `Target`, i.e. not just an image, but lots of things, with all of its dependencies (we should expand `Copy()` to support this). The tree example from [the very bottom of this doc](https://github.com/oras-project/distribution/blob/feature_oras_referrers/docs/referrers.md) means that if I run `Copy()` on the tag `v1.0`, then the entire tree will be replicated to the `to Target`
* find all of the referenced dependencies of a given tag or she, i.e. ref, possibly filtered. At first blush, this is not an expansion of `Copy()`, but rather getting some information and spitting it out to the user. Is this the "new thing" we are after with `Discover()`? Or something else?

For the latter, I would argue that it is precisely another iteration of `Copy()`, except that we are filtering for certain media types, using some memory store as the `to Target`, and then extracting the data. This would keep `Copy()` as clean as always, and just add a later stage.

What else did we miss? And how does `Object` play into this?

juliusl:
@deitch Hey there, great questions. I needed a weekend to recover before finishing up the rest of my examples. I will try to answer all of your questions, and hopefully I won't disappoint. 

> Is it correct to state that everything that I want to copy - from any "there" to any "there" - is essentially through a remote.Resolver, i.e. something that has a Pusher and a Fetcher? And that the remotes.* is the tx/rx modules in your RF analogy, while the content.* is the local connection on each end between the tx/rx module and its local buffers (or whatever it connects to locally)?

Yes, you've hit the nail on the head. 

> If I did understand it correctly, then I am somewhat confused as to why we are exposing things like ReaderAt() on the various content stores. Isn't that an internal thing? As far as. I can tell, it only is used in Object.Move(), which "moves" something from an Artifact to a Target.

I admit, this part is a bit more confusing, but only because it's intentionally left open-ended. Since an Artifact by my definition is completely end-user defined, I decided to use content.Store to keep the symmetry with the rest of the library, while reducing what an implementor needs to worry about. The use case this covers is the ability to not have to worry about registry-centric stuff while processing content. For example, since Discover is meant to be an extension of remotes.Resolver and by extension target.Target, it might get confusing if you started to pass around target.Target without knowing which one can actually Discover. By separating the concerns here, you can be more deliberate because if you pass around a target.Artifact to gain access to the underlying content, this would never get confused. 

By that same notion, what if you wanted the duality? That's actually an interesting case and is completely valid also. If you had a type that implemented both target.Target and target.Artifact, the Artifact would act as a backend, and the target.Target would act as the frontend. This would let you add content to the store to enrich it, without needing to expose that content with your resolver with a reference. 

For example, say you needed to download all of your signatures and sboms locally to an oci store so that you can do your validation, but you did not want these to be resolvable or attached to your registry content until you've finished that validation. With the Object/Artifact combo this is doable, you can incrementally download content into your oci store, or even download it adhoc, and once you're ready, you can call .Move on the object back to the same oci store to make it available to the resolver! Pretty neat, and this granularity has been helping us move internally.

I didn't directly address all of your bullet points, but I hope this illustrates the usages a bit more clearly. If you take a look at my hackmd once more, I've finally finished the component that ties this all together and shows this in action. 

[hackmd - adding Discover() to content.OCI](https://hackmd.io/MOqZ9mq5TlOhgCLJyPEERw?view#Adding-Discover-to-contentOCI-and-the-duality-of-targetTargettargetArtifact)

Other things that this applies to off the top of my head, are Snapshotters. I am probably going to tinker with this next, but Snapshotters are an example of something where this granularity could be helpful. They fit into the RX/TX pattern, as a snapshotter pulls content from a remote, before it can make that content available to the host, there is processing that needs to happen before the snapshotter can TX this to any consumers. 

Another example is webhooks. If you were to Push a reference, it's possible to trigger a webhook. I'm not sure how people will end up creating targets, but if you separate the concerns between RX/TX this won't even be a problem, because moving content through the RX would not activate the push webhook. 

I have some more ideas to play around with, but in general if I could come up with a way to describe it, target.Artifact would be content that is host-less, it would be an incomplete reference with just a namespace and object. 

If I were to translate this into a mock cli cmd it might look like: 

```
oras open localhost:5000/net-monitor:v1 --validator
oras copy --validator /net-monitor:v1 other.registry.io/net-monitor-validated:v1 -r 
```

At this point, this might be a brain dump but, another way to put it, target.Artifact lets us create pseudo hosts. 

On our side we're excited to use these new interfaces. As I said before we have teams internally that are consuming these new interfaces, and the added granularity has allowed them to pretty much convert all of their code to these new api's. Also, with my last commit, we can now start talking about exporting the object graph as a single image. That means we can use the new api's to build a graph locally, and then Copy to push it to registries that don't currently support the new artifacts manifest, without needing to add any complexity to Copy.

OK last analogy (since you're a fellow EE :D), target.Artifact is like the breadboard, and target.Target is like an IC, idk if that one hit any marks but it's just another thing that popped into my head.

deitch:
I started to get it, then lost it in some of the direction.

I don't object to _more_ capabilities in the API. I just still have a hard time seeing how they don't mix things and make them more complex. Maybe that complexity is a necessity 🤷‍♂️ 

It sounds like your definition of `Artifact` (the interface, not the concept) is something that sort of references a single manifest in an actual remote. So if `Target` = "acr", then `Artifact` = "this sbom in the acr target" (and I still do not know what `Object` represents).

If the examples you gave are behaviours you are looking for, then I would see a simpler flow as follows:

`Copy(ref, from, to)` where: `ref` = the reference to the DAG in the remote registry; `from` = the remote registry; `to` = your local OCI store. 

The parts that are missing?

* `ref` is a tag, the current implementation follows the root manifest and all of the things it points to, does not understand how to read the artifact-implementing references. It might pull down the config and layers of a traditional image, but not the cross-referenced signature or sbom. So it needs expansion.
* `ref` is a tag, the current implementation doesn't know how to copy _just_ the specific artifact types, like sbom or signature or whatever, rather than the actual image. I _think_ this would be an option to `Copy()`, but it might be a different func call altogether.
* we do not have handlers to filter out just the specific artifact types. Actually we do, with the various `CopyOpt` that allow you to pick media types or such.
* we do not have the ability to handle signature validation before allowing into a local store. Actually, we do, with the various `CopyOpt` handlers. Given a well-defined use case, like you have here, I would like to see a wrapper around some of that to make it easier, such as `WithValidateSignature` that would validate before letting it into the `to`

It isn't that I want everything to be built into `Copy`; it is that it is so painfully simple to understand, and abstracts out all of the complexity into individual `Target` implementations and some `CopyOpt`, that the UX for using it is straightforward.

> If I were to translate this into a mock cli cmd it might look like:

What is `oras open --validator`?

> target.Artifact lets us create pseudo hosts

Huh?

> OK last analogy (since you're a fellow EE :D), target.Artifact is like the breadboard, and target.Target is like an IC, idk if that one hit any marks but it's just another thing that popped into my head.

I kind of view it the other way. `Target` is the breadboard or my storage bin, so I do a `Copy()` from the storage bin to the breadboard, or reverse, or one breadboard to another; `Copy()` doesn't know much or care about the implementation. `ref` is the reference to the IC, the reference to the thing that I move from one `Target` to another.

If I need greater granularity than `ref` can provide, then we need to expand it; if there are options that are missing, we should provide them. And if there are things we need to do to an IC _while on the breadboard_, then the breadboard should provide those options, i.e. `Target` needs to be expanded.

That is why I can see an `Artifact` that references the IC, and perhaps an ability to tell the breadboard "get the voltage between two pins", i.e. `Target.ReadVoltage(ref)` (and if `ref` isn't good enough, let's expand it).

I really like the breadboard analogy. It highlights how simple `Copy()` is. All it knows is a reference to one breadboard (`Target`), reference to another (`Target`), way to tell the first breadboard "this is how you find the IC I want to copy from you", a way to tell the other breadboard "this is how you will name the IC I want to copy to you", and we are done.

It sort of sounds like you are saying:

* "I want to be able to do things while it is in the Target / on the breadboard", which is totally legitimate and an argument for expansion of what `Target` provides
* "I want to be able to reference the thing in the Target beyond just reference (sha or tag)", which is totally legitimate and an argument for greater flexibility than `ref`, perhaps a way to say, "ref which means its manifest" or "ref which means the things pointing to it", but I think that would be covered in the "copy but restrict just to signatures" or "copy but restrict just to artifacts that reference the hash of the root"

juliusl:
@deitch  I'm honestly having a hard time articulating this because it's just so abstract and intuitive to me and so bear with me. 

> It sounds like your definition of Artifact (the interface, not the concept) is something that sort of references a single manifest in an actual remote. So if Target = "acr", then Artifact = "this sbom in the acr target" (and I still do not know what Object represents).

Close, but actually what I mean is, Target = "acr", Artifact = "validator of sboms". So if Target is an interface for a container image registry, then Artifact is an interface for the processing of artifacts. Object is just POD that can be used to process artifacts at runtime, and as you noted at the bottom of your reply, my attempt at designing a better `ref`.

> It isn't that I want everything to be built into Copy; it is that it is so painfully simple to understand, and abstracts out all of the complexity into individual Target implementations and some CopyOpt, that the UX for using it is straightforward.

Sure, I think that's all fine, and I'm trying to add to that ecosystem. Simplicity aside, one important use case for us that I would like to discuss: 

```
var subject target.Object

... 

var manifest struct {
	Version   int                  `json:"schemaVersion"`
	MediaType string               `json:"mediaType"`
	Config    ocispec.Descriptor   `json:"config"`
	Layers    []ocispec.Descriptor `json:"layers"`
}

// Deserialize the manifest for browsing without pulling the entire image
err = subject.MarshalJson(ctx, source, &manifest)
if err != nil {
	return err
}
```

If you were to use Copy to do the same thing, it would end up looking like this: 

```
var (
  registry target.Target
  store    target.Target
)

desc, err := oras.Copy(ctx, registry, "", store, "", opts...) 
if err != nil {
	return err
}

fetcher, err := store.Fetcher(ctx, "")
if err != nil {
	return err
}

reader, err := fetcher.Fetch(ctx, desc)
if err != nil {
	return err
}
defer reader.Close()

var manifest struct {
	Version   int                  `json:"schemaVersion"`
	MediaType string               `json:"mediaType"`
	Config    ocispec.Descriptor   `json:"config"`
	Layers    []ocispec.Descriptor `json:"layers"`
}

err = json.NewDecoder(reader).Decode(&manifest)
if err != nil {
	return err
}
```
Now the weakness of doing this is that when you call oras.Copy, that will copy the entire image to the store. However, what if all you intended was to peek at the Manifest? This is an actual use case brought up to me, and while I am aware that you can do, 

```
ref, desc, err := registry.Resolve(ctx, "") 
if err != nil {
	return err
}

fetcher, err := registry.Fetcher(ctx, ref)
if err != nil {
	return err
}

reader, err := fetcher.Fetch(ctx, desc)
if err != nil {
	return err
}
defer reader.Close()

var manifest struct {
	Version   int                  `json:"schemaVersion"`
	MediaType string               `json:"mediaType"`
	Config    ocispec.Descriptor   `json:"config"`
	Layers    []ocispec.Descriptor `json:"layers"`
}

err = json.NewDecoder(reader).Decode(&manifest)
if err != nil {
	return err
}
```

That's only because I know how the remotes.Resolver interface works, but now we've steered the user away from oras.Copy. 
Aside from that, that's a bit of code for something really straightforward, so why not build it into a class? Which is how oras.Graph and target.Object work together, 

```
objects, err := oras.Graph(ctx, ref, artifactType, source, nil)
if err != nil {
  return err
}

var manifest struct {
	Version    int                  `json:"schemaVersion"`
	MediaType string               `json:"mediaType"`
	Config    ocispec.Descriptor   `json:"config"`
	Layers    []ocispec.Descriptor `json:"layers"`
}

// First element is always the root
err = objects[0].MarshalJson(ctx, source, &manifest)
if err != nil {
  return err
}
```

Then if you had a target,Artifact in play, say something that Validates on Download, then you can have: 
```
var source target.Target
var validator target.Artifact

objects, err := oras.Graph(ctx, ref, artifactType, source, graph.DownloadToArtifact(ctx, ref, source, validator))
if err != nil {
  return err
}

var manifest struct {
	Version    int                  `json:"schemaVersion"`
	MediaType string               `json:"mediaType"`
	Config    ocispec.Descriptor   `json:"config"`
	Layers    []ocispec.Descriptor `json:"layers"`
}

// First element is always the root
err = objects[0].MarsjalJsonFromArtifact(ctx, validator, &manifest)
if err != nil {
  return err
}
```
All it took to enhance the behavior here was a target.Artifact and a built-in walk function. I hope this illustrates how everything is interacting a bit better.

It's kind of hard to have an objective discussion about simplicity/complexity, because we all as humans have a different range of what we think is complex and simple. I initially was going to encapsulate a bunch of stuff into oras.Graph, but then after listening to teammates who are consuming oras-go, I've learned that less is more. They were happy with this level of complexity, because the reality is they've already invested a lot into their own code. In order to adapt to a monolithic api such as oras.Copy, it would involve re-architecting the system and extra bloat. By giving them lower level api's they were able to just drop this into their code. It provides a better launching point for those who are already up to speed with runtime code.

> I kind of view it the other way. Target is the breadboard or my storage bin, so I do a Copy() from the storage bin to the breadboard, or reverse, or one breadboard to another; Copy() doesn't know much or care about the implementation. ref is the reference to the IC, the reference to the thing that I move from one Target to another.
> 
> If I need greater granularity than ref can provide, then we need to expand it; if there are options that are missing, we should provide them. And if there are things we need to do to an IC while on the breadboard, then the breadboard should provide those options, i.e. Target needs to be expanded.

I might be nitpicking, but this seems to contradict your earlier statement: 

> It isn't that I want everything to be built into Copy; it is that it is so painfully simple to understand, and abstracts out all of the complexity into individual Target implementations and some CopyOpt, that the UX for using it is straightforward.

It sounds like you are saying anything that is missing must be built into Copy. But, if we were to build every missing feature into an CopyOpt, wouldn't that also make the UX less straightforward and Copy more complex?

> If the examples you gave are behaviours you are looking for, then I would see a simpler flow as follows:
> 
> Copy(ref, from, to) where: ref = the reference to the DAG in the remote registry; from = the remote registry; to = your local OCI store.

This is simpler if the DAG actually exists in the remote registry, but what your example glosses over is how did the DAG end up in the from in the first place? That is another thing this contribution is addressing as well.  

I hope that clarifies some things, but it seems like we're having a more productive conversation.

deitch:
> it's just so abstract and intuitive to me

I know. Some people are really good at seeing the abstract. Others are better with the concrete.

> if Target is an interface for a container image registry, then Artifact is an interface for the processing of artifacts. Object is just POD that can be used to process artifacts at runtime, and as you noted at the bottom of your reply, my attempt at designing a better ref

I didn't get this. We are in agreement that `Target` represents a place to store CAS - could be local files, or registry, or memory, or anything. What did you mean that "Artifact is an interface for the processing of artifacts"? And "Object is just POD that can be used to process artifacts at runtime"?

> but it seems like we're having a more productive conversation

Yes, indeed, we are getting closer to the more practical.

By the way, I am becoming more convinced that you and I are getting quite a bit closer to each other, as we understand the positions better.

> It sounds like you are saying anything that is missing must be built into Copy. But, if we were to build every missing feature into an CopyOpt, wouldn't that also make the UX less straightforward and Copy more complex?

Not quite. I am positing the following:

* if I want to copy a whole tree of something from one place to another, use `Copy` - exists today
* if I want to copy a whole tree of something including references from one place to another, use `Copy` - it doesn't support it correctly today, but we should expand it
* if I want to copy a subset of something from one place to another, use `Copy` with some opts - sort of supported today, could use some simple opts
* if I want to process something, that is external to `Copy`; that isn't "move this from here to there", but rather, "given that it is there, do some analysis on it" (agreeing with what you said). I might implement this under the covers as "first Copy whatever I need from there to some other more convenient in-memory there and process it", but I might not
* if I have a common use case that does a bunch of steps like that, wrap it all with a convenience function

> That is another thing this contribution is addressing as well

Huh? Why do we care how it ended up in a `Target` in the first place? It just is; the starting assumption of any of these commands is that items exist in a `Target`.

> If you were to use Copy to do the same thing, it would end up looking like this:
> Now the weakness of doing this is that when you call oras.Copy, that will copy the entire image to the store. However, what if all you intended was to peek at the Manifest?

This is a really good example of a common use cases that does a bunch of steps like that. With the right `CopyOpt`, this would not copy the whole image, just the manifest. I would 100% agree that it should be simpler, and in that case, a nice convenience function that might look like:

```go
manifest, err := oras.GetManifest(from, ref) // pick your function name
```

I think we are quiet a bit closer. The part I still don't get it is the `Artifact` and `Object`. It almost sounds like an `Artifact` represents either a root descriptor or manifest in a `Target`, sort of "ref+`Target`", or the entire tree, sort of a genericized image. But you said that isn't what it is. And I still don't get what an `Object` is, and why I need it.

shizhMSFT (approved):
LGTM to be merged to the `artifacts` branch.

Issue #63, opened 2021-11-17
Remove old implementation


shizhMSFT:
Once the new implementation is done and well tested, we can drop the old implementaion. For users who want the old implementation, they can pin the `oras-go` version to `v0.5.0`.

shizhMSFT:
Since we have already released `v1.0.0` and all `v1` related implementation are in the [v1](https://github.com/oras-project/oras-go/tree/v1) branch, we can safely drop the old implementation in the `main` branch.

/cc @sajayantony @deitch @jdolitsky

sajayantony:
Sounds reasonable. I think reshaping this for 2.0 makes sense.

Issue #54, opened 2021-11-17
Memory Target


shizhMSFT:
To make `Copy` useful, we need to implement a `Memory` target based on the current `oras.land/oras-go/pkg/content.Memory`.

Issue #49, opened 2021-11-17
[Epic] ORAS Go Library Revamp


shizhMSFT:
Targeting releasing `v2.0.0`, it is important to stabilize the `oras-go` library with the following criteria since `oras-go` is and will be used as a fundamental building block of production projects and even security products:

- Simple to use.
- Fewer dependencies.
  - More dependencies imply more maintenance work.
  - We need to release frequently just to update the dependencies if there are updates, especially security updates.
- No breaking changes on minor releases.
  - We can start with a few exported packages, and develop or export new packages slowly.
- Well tested.
- Well documented.

This issue is used to track the process of the `oras-go` revamp in order to make `oras-go` meeting the above releasing criteria. Tasks are created as follows:

- Milestone 1
  - [x] #50 
  - [x] #62 
- Milestone 2
  - [x] #52
  - [x] #53 
- Milestone 3
  - [x] #54 
  - [x] #55 
  - [x] #56 
- Milestone 4
  - [x] #57 
  - [x] #58 
- Milestone 5
  - [x] #59 
  - [x] #60 
  - [x] #61 
- Milestone 6
  - [x] #63

/cc @sajayantony @SteveLasker @jdolitsky @deitch @northtyphoon

sajayantony:
@shizhMSFT Which item do you consider things like retrieving manifests from local or remote target?

SteveLasker:
looks good. 
Where would you see unit tests being added for stability to changes?  
Is the expectation that each milestone can be validated with the oras client?

shizhMSFT:
> @shizhMSFT Which item do you consider things like retrieving manifests from local or remote target?

Manifests are modeled as blobs. Therefore, retrieving manifests can be done by retrieving the blob. Since target has a tag resolver, manifests can be retrieved by generic tags.

shizhMSFT:
> Where would you see unit tests being added for stability to changes?

The unit tests will be added in the corresponding go packages.

> Is the expectation that each milestone can be validated with the oras client?

Milestones here are like stages. We can start updating `oras` CLI when the milestone 4 is completed.

shizhMSFT:
Closing as all issues are resolved.

PR #67, opened 2021-11-18
Copy API

### LINKED ISSUE #52: Copy API
Based on #50, the `Copy` API originated by #8 need to be re-implemented without `containerd`.

That is, we need our version of

- `images.Childen`
- `images.Dispatch`

Since we are implementing the `images.Dispatch`, we can optimize the graph traversal for our ORAS scenarios.

In fact, `images.Dispatch` is designed for [trees](https://en.wikipedia.org/wiki/Tree_(data_structure)) not for [DAGs](https://en.wikipedia.org/wiki/Directed_acyclic_graph). It is less inefficient if there are multiple nodes pointing to the same node. The pointed node might be transferred multiple times at the same time based on the concurrency level.

In the current code, we also implemented something equivalent to `remotes.PushContent`. Although the layers are transferred in parallel, the manifests are transferred in sequences. It might not be efficient if we want to copy a manifest list or index.

shizhMSFT:
Resolves #52


deitch:
Two questions:

1. It looks like `copy.go` is mostly copied from `pkg/oras/copy.go`. If this replaces that, should the other one not go away?
2. I see a lot of parts in `internal`, like `docker` and `memory`. Should those not be exposed?

shizhMSFT:
> It looks like `copy.go` is mostly copied from `pkg/oras/copy.go`. If this replaces that, should the other one not go away?

Yes, the old one will go away once #49 is completed. It is tracked by issue #63.

> I see a lot of parts in `internal`, like `docker` and `memory`. Should those not be exposed?

They should not be exposed without a proper discussion. Once we expose them, we need to maintain them, and they cannot be taken back for backward compatibility considerations.

deitch:
> Yes, the old one will go away once #49 is completed. It is tracked by issue #63.

Got it. So we are going to have both for a little while? I don't love it, but as long as it is for a reasonably short period of time, not so big a deal.

> They should not be exposed without a proper discussion. Once we expose them, we need to maintain them, and they cannot be taken back for backward compatibility considerations

Some of the io stuff, sure, but what about the Memory implementation of `Target`, etc.? Those currently are exposed in `content`, and should continue to be. How else will someone use them?

shizhMSFT:
> Some of the io stuff, sure, but what about the Memory implementation of `Target`, etc.? Those currently are exposed in `content`, and should continue to be. How else will someone use them?

The `Memory` implementation in the `internal/cas` package is just an implementation of `content.Storage`. It is not qualified as a `Target`.

I am currently working the `Memory` implementation of `Target`. It will sit in the `content/memory` package. This work is tracked by the issue #54.

go.mod
```diff
@@ -2,6 +2,9 @@ module oras.land/oras-go/v2
 
 go 1.16
 
-require github.com/opencontainers/image-spec v1.0.2
-
-require github.com/opencontainers/go-digest v1.0.0 // indirect
+require (
+	github.com/opencontainers/go-digest v1.0.0
+	github.com/opencontainers/image-spec v1.0.2
+	github.com/oras-project/artifacts-spec v1.0.0-draft.1.1
```
sajayantony, line 8: Ok great. This uses the updated spec with MediaTypes.

internal/graph/graph.go
```diff
@@ -0,0 +1,112 @@
...
+// - `postHandler` is called after traversing the down edges.
+// An optional concurrency limiter can be passed in to control the concurrency
+// level.
+// A handler may return `ErrSkipDesc` to signal not traversing descendants.
+// If any handler returns an error, the entire dispatch is cancelled.
+// This function is based on github.com/containerd/containerd/images.Dispatch.
+// Note: Handlers with `github.com/containerd/containerd/images.ErrSkipDesc`
+// cannot be used in this function.
+// WARNING:
+// - This function does not detect circles. It is possible running into an
```
sajayantony, line 41: We should guard this as over time ORAS will be called in runtimes and not just CLIs. I would say we prioritize this or ensure you have an issue and let's get this fixed. We also use this in Ratify and so it would blow up the verifier if we see a graph that is messed up.

shizhMSFT: This function is guard as an [internal](https://go.dev/doc/go1.4#internalpackages) package. The library consumers are not able to import this `internal/graph` pacakge. In other words, only ORAS can invoke this function, and Ratify cannot. As long as we have proper guarded in the exported methods of ORAS, we are fine.

In this PR, `Dispatch()` is invoked by `CopyGraph()` and `CopyGraph()` avoids circles.

sajayantony: Ok sounds good. If we don’t have any more feedback would LGTM this change and keep making progress. 
I’ll ack by EOD PST.

sajayantony (approved):
LGTM.

SteveLasker (approved):
LGTM

internal/graph/handler.go
```diff
@@ -0,0 +1,76 @@
...
+	ErrStopHandler = errors.New("stop handler")
+
+	// ErrSkipDesc signals the handler chain generated by Handlers() to stop
+	// processing the rest handlers. The down edge nodes returned by the
+	// previous handlers are ignored.
+	// This error also applies to single handlers.
+	ErrSkipDesc = errors.New("skip descriptor")
+)
+
+// Handler processes conntents. This interface is equivalent to
```
Wwwsylvia, line 37: nit: typo here

Wwwsylvia (approved):
LGTM

internal/graph/handler.go
Wwwsylvia:
```suggestion
// Handler processes contents. This interface is equivalent to
```

shizhMSFT: Nice catch!

jdolitsky (approved):
great test coverage here

deitch (approved):

PR #73, opened 2021-11-19
Prep for v1.0.0 release

jdolitsky:
- Update Go modules
- Update GitHub Actions to watch v1 branch
- Remove stability warning from README


sajayantony (approved):

sajayantony:
@jdolitsky - Thank you for making this release happen.

shizhMSFT:
@jdolitsky Recently, I've learnt that we can use `GITHUB_TOKEN` to do release instead of using our own PAT.

I've updated for https://github.com/notaryproject/notation/pull/135. Do you want to also apply that to `oras` and `oras-go`?

shizhMSFT (approved):
LGTM

jdolitsky:
@shizhMSFT - I think yes, absolutely, for releases with binaries.

In this case, since Go modules uses git, I will just create the tag

PR #77, opened 2021-11-20
Update `go.mod` to `oras.land/oras-go/v2`

### LINKED ISSUE #76: Update `go.mod` in `main` to `oras.land/oras-go/v2`
Since we have `oras-go` [version 1](https://github.com/oras-project/oras-go/releases/tag/v1.0.0) released, we have to update the `go.mod` in the `main` branch to `oras.land/oras-go/v2` as documented in the [go doc](https://golang.org/ref/mod#modules-overview).

> If the module is released at major version 2 or higher, the module path **must** end with a [major version suffix](https://golang.org/ref/mod#major-version-suffixes) like `/v2`. This may or may not be part of the subdirectory name. For example, the module with path `golang.org/x/repo/sub/v2` could be in the `/sub` or `/sub/v2` subdirectory of the repository `golang.org/x/repo`.

shizhMSFT:
Resolves #76

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,7 @@
-module oras.land/oras-go
+module oras.land/oras-go/v2
 
 go 1.16
+
+require github.com/opencontainers/image-spec v1.0.2
+
+require github.com/opencontainers/go-digest v1.0.0 // indirect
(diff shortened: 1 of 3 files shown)


deitch:
Is this not going to cause issues elsewhere on `pkg/`, where it still includes the non-v2 path?

shizhMSFT:
The `pkg` folder will not be impacted. Before we releasing `2.0`, we will do #63 and eventually remove the entire `/pkg` folder.

All old implementation will still be there in the `v1` branch and tagged by `v1.x.x`. Developers can import `oras.land/oras-go v1.0.0` to access `oras.land/oras-go/pkg`.

SteveLasker (approved):
LGTM

PR #80, opened 2021-12-02
Memory Target

### LINKED ISSUE #54: Memory Target

shizhMSFT:
- Resolves #54.
- Add test for `oras.Copy()` using the `Memory` target.

### DIFF
--- a/content/memory/memory.go
+++ b/content/memory/memory.go
@@ -1,0 +1,115 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package memory
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"sync"
+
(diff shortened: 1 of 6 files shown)


content/memory/memory.go
```diff
@@ -0,0 +1,115 @@
...
+	upEdgeKey := descriptor.FromOCI(expected)
+	downEdges, err := content.DownEdges(ctx, s.storage, expected)
+	if err != nil {
+		return err
+	}
+	for _, downEdge := range downEdges {
+		downEdgeKey := descriptor.FromOCI(downEdge)
+		upEdgesValue, _ := s.upEdges.LoadOrStore(downEdgeKey, &sync.Map{})
+		upEdges := upEdgesValue.(*sync.Map)
+		upEdges.Store(upEdgeKey, expected)
```
Wwwsylvia, line 69: The `upEdgeValue` variable naming is confusing. It looks like a value of the upEdges and is paired with `upEdgeKey`, but it is not...

shizhMSFT: Let's reduce the confusion by just naming it `value`.

Wwwsylvia (approved):
LGTM

sajayantony:
I believe this addreses @deitch's earlier questions. Would be good to have an LGTM from @deitch

deitch (approved):
LGTM. I compared it to what we have, I think I rather like how this is structured. I spent some time on the internal structure, but mostly seeing functionality via the good tests.

deitch:
I was going to merge this in, but don't know if you are ready for it @shizhMSFT .

shizhMSFT:
@deitch I'm ready to merge it.

PR #81, opened 2021-12-02
Registry Interface

### LINKED ISSUE #58: Registry Interface
Since oras is not just a tool for transfering contents between storages, it is also a fully functional registry client now. Therefore, we need to design a proper user-friendly interface for the Registry.

Based on [Docker Registry HTTP API V2](https://docs.docker.com/registry/spec/api/), the registry client should have the following abilities:

1. List repositories.
2. Upload / Download / Delete layers.
3. Upload / Download / Delete manifests.
4. Tag manifest.
5. Resolve tag.
6. List tags.

Since ability 2 - 5 are already covered by the `Target` interface, the registry interface should just need two more methods:

- List repositories.
- List tags.

We can open another issue for the referrers API support

shizhMSFT:
Resolves #58

### DIFF
--- a/registry/reference.go
+++ b/registry/reference.go
@@ -16,6 +16,7 @@ package registry
 
 import (
 	"fmt"
+	"net/url"
 	"regexp"
 	"strings"
 
@@ -25,9 +26,19 @@ import (
 
 // regular expressions for components.
 var (
-	registryRegexp   = regexp.MustCompile(`^(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))*(?::[0-9]+)?$`)
+	// repositoryRegexp is adapted from the distribution implementation.
+	// The repository name set under OCI distribution spec is a subset of the
+	// the docker spec. For maximum compability, the docker spec is verified at
+	// the client side. Further check is left to the server side.
+	// References:
+	// - https://github.com/distribution/distribution/blob/v2.7.1/reference/regexp.go#L53
+	// - https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests
 	repositoryRegexp = regexp.MustCompile(`^[a-z0-9]+(?:(?:[._]|__|[-]*)[a-z0-9]+)*(?:/[a-z0-9]+(?:(?:[._]|__|[-]*)[a-z0-9]+)*)*$`)
-	tagRegexp        = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
+
+	// tagRegexp checks the tag name.
+	// The docker and OCI spec has the same regular expression.
+	// Reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests
+	tagRegexp = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
 )
 
 // Reference references to a descriptor in the registry.
@@ -99,7 +110,8 @@ func (r Reference) Validate() error {
 
 // ValidateRegistry validates the registry.
 func (r Reference) ValidateRegistry() error {
-	if !registryRegexp.MatchString(r.Registry) {
+	uri, err := url.ParseRequestURI("dummy://" + r.Registry)
+	if err != nil || uri.Host != r.Registry {
 		return fmt.Errorf("%w: invalid registry", errdef.ErrInvalidReference)
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


registry/repository.go
```diff
@@ -0,0 +1,87 @@
...
+	content.Storage
+	content.Deleter
+	content.Resolver
+}
+
+// TagPusher provides advanced push with the tag service.
+type TagPusher interface {
+	// PushTag pushes the manifest with a reference tag.
+	// It is equivalent to call `Push()` and then `Tag()` but more efficient or
+	// equal.
```
sajayantony, line 73: Is there a push manifest equivalent? Basically digest only push manifest?

registry/reference.go
```diff
@@ -0,0 +1,165 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/opencontainers/go-digest"
+	"oras.land/oras-go/v2/errdef"
+)
+
+// regular expressions for components.
+var (
+	registryRegexp   = regexp.MustCompile(`^(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))*(?::[0-9]+)?$`)
```
Wwwsylvia, line 28: Can we add some explanations for these?

shizhMSFT: Code update and this complex regexp is removed.

registry/repository.go
shizhMSFT: Yes, `Repository` is a `content.Pusher` and thus users can push manifest by digest. This method is also called in `Copy()`.

Here's the interface decomposition.
- `Repository` is an `oras.Target`.
- `oras.Target` is a `content.Storage`.
-  `content.Storage` is a `content.Pusher`.

Wwwsylvia (approved):
LGTM

deitch:
@shizhMSFT I am trying to wrap my head around it. I have a few questions, although I think I get it.

* registry as top-level: Why is the `registry/` package at the top-level and not part of content? I think you are stating that there are **two** (maybe more) high-level interfaces: `Copy()` which copies things from `Target` A to `Target` B; and `registry.Registry` which provides for interactions with a registry. Of course, I can use a `registry.Registry` to get a `registry.Repository`, which satisfies a `Target`, which is usable with `Copy()`. Is that it? If so, I think having the docs state explicitly what the key to-level entry points to oras are. I was thinking entirely in `Copy()` terms until I got the above, and then it all seemed to make sense (if I got it right, that is).
* `content` and `Target`. I understand why a `registry.Repository` would fulfill `Target`, so it can be used in `Copy()`. What are the other `content.*` interfaces that it implements, and why does it need them?

shizhMSFT:
> `registry.Registry` which provides for interactions with a registry. Of course, I can use a `registry.Registry` to get a `registry.Repository`, which satisfies a `Target`, which is usable with `Copy()`. Is that it?

@deitch Yes, exactly.

Behind the scene, `oras` and `oras-go` are envolving into a full registry client in addition to its content `Copy()` functionality. Since it is a full registry client, it would be better to promote the `content/registry` package to the top level (i.e. `registry` as it is in the PR). @jdolitsky @sajayantony Any votes on the package path?

> What are the other `content.*` interfaces that it implements, and why does it need them?

Besides `oras.Target`, the `registry.Repository` interface includes `content.Deleter` so that it can delete a blob or a manifest since it is a full registry client. Note that both `oras.Target` and `BlobStore` have `content.Storage` and `content.Resolver`.
For other interfaces like `TagPusher`, `Blobs()`, `Manifests()`, `Tags()`, I've documented them and they are there for the functionalities of a registry client.

shizhMSFT:
To get the full picture, you may take a look at [my current working branch](https://github.com/shizhMSFT/oras-go/tree/dev/registry/distribution), which I will make a PR and send it out after this PR is merged.

Here are some example usages (note: I am still implementing the auth):

```golang
src, err := distribution.NewRepository("localhost:5000/hello-world")
check(err)
src.PlainHTTP = true

dst, err := distribution.NewRepository("localhost:5001/hello")
check(err)
dst.PlainHTTP = true

desc, err := oras.Copy(ctx, src, "latest", dst, "")
check(err)
```

```golang
src, err := distribution.NewRepository("mcr.microsoft.com/hello-world:latest")
check(err)

dst := memory.New()

_, err = oras.Copy(ctx, src, src.Reference.String(), dst, "")
check(err)
```

where the full package name of `distribution` is `oras.land/oras-go/v2/registry/distribution`.

deitch:
Thanks for the explanation @shizhMSFT ; makes sense. 

I do think we need clearer docs explaining what the high level interfaces are (primarily two: `Copy` and Registry`), but that needs to be fixed [here](https://oras.land/client_libraries/go/) and can come later.

deitch (approved):

shizhMSFT:
We definitely will document the library in the `oras-www` repo.

PR #85, opened 2021-12-19
Registry Auth Client

### LINKED ISSUE #57: Registry Target
To make `Copy` useful, we need to implement a registry client for the `Registry` target.

shizhMSFT:
Registry authentication implementation for #57.

Change logs:

- Implementation authentication for the registry client.
- Resolves #57.
- Addresses comments in #82.
- Bump up golang version from `v1.16` to `v1.17` due to the use of new built-in library functions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module oras.land/oras-go/v2
 
-go 1.16
+go 1.17
 
 require (
 	github.com/opencontainers/go-digest v1.0.0
--- a/internal/httputil/seek.go
+++ b/internal/httputil/seek.go
@@ -21,9 +21,17 @@ import (
 	"net/http"
 )
 
+// Client is an interface for a HTTP client.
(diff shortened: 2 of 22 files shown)


deitch (approved):
I felt like I was reading through my reimplementation of the OCI registry client in Ruby years ago, but much better done.

I do wish we could have a shared implementation of this to everyone. Now docker has one in moby, and google in ggcr, and container has one, and...

As long as this works, it is good. Maybe at one point we all can join forces?

shizhMSFT:
> As long as this works, it is good. Maybe at one point we all can join forces?

Yes, we will 😄 and `oras-go` has lots of advantages. It is more extensible and has fewer dependencies.

Most importantly, it will be integrated into many security products.

SteveLasker:
great work @shizhMSFT, @deitch 
High profile area always have multiple implementations. The refactoring @deitch started with @shizhMSFT have been simply marvelous!

Issue #55, opened 2021-11-17
OCI Target


shizhMSFT:
To make `Copy` useful, we need to implement a `OCI` target based on the current `oras.land/oras-go/pkg/content.OCI`.

OCI Layout specification can be found at https://github.com/opencontainers/image-spec/blob/main/image-layout.md

Issue #60, opened 2021-11-17
Update Example


shizhMSFT:
Once we have the least code working, we should update the examples so that people can try themselves.

Examples (originally from @sajayantony 's list):

- [x] Download Manifest
- [x] Download a Blob given the digest
- [x] Put Manifest
- [x] Create manifest with Annotation
- [x] Upload Blob
- [x] Upload Manifest with blob
- [x] Copy Artifact
- [x] Insecure TLS / Plain HTTP
- [ ] HTTP Client Configuration

sajayantony:
Add the list of items we potentially need as examples - 

# ORAS API Usecases

The following usecase can help define the list of examples that a new developer should be able to refer. 

1. Download Manifest
2. Download a Blob given the digest 
3. Put Manifest
4. Create manifest with Annotation
5. Upload Blob
6. Upload Manifest with blob
7. Copy Artifact
8. Upload manifest with reference 
9. Download reference manifest
10. Download blob of reference manifest
11. Copy reference 
12. Copy artifact and references


## Authentication tokens 
1. Get token for operation with scopes. 
2. Validate cross mount 

### Example Permutations 
1. Auth required for pull/push 
2. Insecure planHttp

shizhMSFT:
Assigning this item to @qweeah.

qweeah:
Ack

sajayantony:
@qweeah - can you include GetManifest in the examples.

qweeah:
> @qweeah - can you include GetManifest in the examples.

@sajayantony I think it's covered by [Example_pullByTag](https://github.com/oras-project/oras-go/blob/main/registry/remote/example_test.go#L490) and [Example_pullByDigest](https://github.com/oras-project/oras-go/blob/main/registry/remote/example_test.go#L523)

wangxiaoxuan273:
> @wangxiaoxuan273 Could you help sweeping all the scenarios and add examples for them?

Sure. Please assign this to me.

Wwwsylvia:
I just noticed that there is no item related to `referrers` in the example list. 
In the following items, does `artifact reference` mean `subject` here? Or do we actually mean `referrers`?  
-  Upload artifact reference manifest
-  Download artifact reference manifest
-  Download blob of artifact reference manifest
-  Copy reference
-  Copy artifact and references

[...]

PR #86, opened 2021-12-24
OCI Target

Wwwsylvia:
Resolves https://github.com/oras-project/oras-go/issues/55


internal/cas/oci.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	defer f.Close()
+
+	value, err := ioutil.ReadAll(content, expected)
+	if err != nil {
+		return "", fmt.Errorf("failed to read from content: %v", err)
+	}
+	if _, err := f.Write(value); err != nil {
+		return "", fmt.Errorf("failed to write ingest file: %v", err)
+	}
```
shizhMSFT, line 159: You may take a look at [io.Copy](https://pkg.go.dev/io#Copy), or its optimized version [io.CopyBuffer](https://pkg.go.dev/io#CopyBuffer). If you choose to use `io.CopyBuffer`, [sync.Pool](https://pkg.go.dev/sync#Pool) is your friend.

internal/cas/oci.go
```diff
@@ -0,0 +1,176 @@
...
+	// simultaneously will not choose the same file.
+	f, err := os.CreateTemp(o.ingestRoot, expected.Digest.Encoded()+"_*")
+	if err != nil {
+		return "", fmt.Errorf("failed to create ingest file: %v", err)
+	}
+	defer f.Close()
+
+	value, err := ioutil.ReadAll(content, expected)
+	if err != nil {
+		return "", fmt.Errorf("failed to read from content: %v", err)
```
shizhMSFT, line 155: If there is any error in this function, will the temp file be removed?

internal/cas/oci.go
```diff
@@ -0,0 +1,176 @@
...
+	if err := os.Chmod(f.Name(), 0444); err != nil {
+		return "", fmt.Errorf("failed to make readonly: %v", err)
+	}
+
+	return f.Name(), nil
+}
+
+// ensureDir ensures the directories of the path exists.
+func ensureDir(path string) error {
+	return os.MkdirAll(path, 0755)
```
shizhMSFT, line 175: It is OK to have `0777` as it will be masked by `umask`.

content/oci/oci.go
```diff
@@ -0,0 +1,265 @@
...
+	root           string
+	referencesPath string
+}
+
+// Reference represents a reference relationship
+// between a descriptor and its referrers.
+type Reference struct {
+	descriptor.Descriptor
+	Referrers []ocispec.Descriptor `json:"referrers"`
+}
```
shizhMSFT, line 53: Do we need to export this structure?

internal/resolver/oci.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	err := oci.loadIndex()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := oci.saveIndex(); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
shizhMSFT: The question behind it is that `loadIndex` is a heavy operation that it need to read the `index.json` file and build the reference. It might not be efficient if we `loadIndex` each time we do read operations.

content/oci/oci.go
```diff
@@ -0,0 +1,265 @@
...
+	resolver, err := resolver.NewOCI(root)
+	if err != nil {
+		return nil, err
+	}
+
+	store := &Store{
+		storage:        storage,
+		resolver:       resolver,
+		root:           root,
+		referencesPath: filepath.Join(root, "references.json"),
```
shizhMSFT, line 71: Hint: The `references.json` file is not a part of the OCI layout.

shizhMSFT (changes requested):

internal/ioutil/io.go
```diff
@@ -17,10 +17,18 @@ package ioutil
 import (
 	"errors"
 	"io"
+	"sync"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+var bufPool = sync.Pool{
```
northtyphoon, line 25: Can you add a comment to call out the buffer size is 1M?

internal/ioutil/io.go
```diff
@@ -44,12 +52,38 @@ func ReadAll(r io.Reader, desc ocispec.Descriptor) ([]byte, error) {
...
+// The copied content is verified against the size and the digest.
+func Copy(dst io.Writer, src io.Reader, desc ocispec.Descriptor) error {
+	// verify while copying
+	verifier := desc.Digest.Verifier()
+	mw := io.MultiWriter(verifier, dst)
+	buf := bufPool.Get().(*[]byte)
+	if _, err := io.CopyBuffer(mw, src, *buf); err != nil {
+		return errors.New("copy failed")
+	}
+	bufPool.Put(buf)
```
northtyphoon, line 72: It doesn't really matter. I usually add "defer bufPool.Put(buf)" right after `bufPool.Get()`.

content/oci/storage.go
```diff
@@ -0,0 +1,166 @@
...
+
+func NewStorage(root string) *Storage {
+	return &Storage{
+		blobRoot:   filepath.Join(root, "blobs"),
+		ingestRoot: filepath.Join(root, "ingest"),
+	}
+}
+
+// Fetch fetches the content identified by the descriptor.
+func (s *Storage) Fetch(_ context.Context, target ocispec.Descriptor) (io.ReadCloser, error) {
```
northtyphoon, line 47: Just curious why context parameter is ignored in Storage implementation? What is it used for?

content/oci/storage.go
```diff
@@ -0,0 +1,166 @@
...
+	if err := dgst.Validate(); err != nil {
+		return "", fmt.Errorf("cannot calculate blob path from invalid digest %s: %v", dgst.String(), err)
+	}
+
+	return filepath.Join(s.blobRoot, dgst.Algorithm().String(), dgst.Encoded()), nil
+}
+
+// ensureDir ensures the directories of the path exists.
+func ensureDir(path string) error {
+	return os.MkdirAll(path, 0777)
```
northtyphoon, line 152: Does it require read/write/execute for everyone?

content/oci/oci.go
```diff
@@ -0,0 +1,227 @@
...
+	switch desc.MediaType {
+	case ocispec.MediaTypeImageManifest, ocispec.MediaTypeImageIndex:
+		if !isDigest(reference) {
+			if desc.Annotations == nil {
+				desc.Annotations = map[string]string{}
+			}
+			desc.Annotations[ocispec.AnnotationRefName] = reference
+		}
+
+		s.resolver.Tag(ctx, desc, reference)
```
northtyphoon, line 116: If `reference` is digest, why does the code continue to run `resolver.Tag`?

content/oci/oci.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	defer indexFile.Close()
+
+	if err := json.NewDecoder(indexFile).Decode(&s.index); err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	for _, desc := range s.index.Manifests {
+		tag := desc.Annotations[ocispec.AnnotationRefName]
```
northtyphoon, line 191: Can `Annotations` be nil?

content/oci/oci.go
```diff
@@ -0,0 +1,227 @@
...
+		return err
+	}
+	if layout.Version != ocispec.ImageLayoutVersion {
+		return errdef.ErrUnsupportedVersion
+	}
+
+	return nil
+}
+
+// load reads the index.json from the file system.
```
northtyphoon, line 169: load => loadIndex

content/oci/oci.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+
+		// traverse the whole DAG and index up edges for all the nodes.
+		s.finder.IndexAll(ctx, s.storage, desc)
+	}
+
+	return nil
+}
+
+// save writes the index.json to the file system.
```
northtyphoon, line 205: save => saveIndex

content/oci/oci.go
```diff
@@ -0,0 +1,227 @@
...
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/internal/graph"
+	reslv "oras.land/oras-go/v2/internal/resolver"
+)
+
+// ociImageIndexFile is the file name of the index from the OCI Image Layout Specification.
+// Reference: https://github.com/opencontainers/image-spec/blob/master/image-layout.md#indexjson-file
+const ociImageIndexFile = "index.json"
```
northtyphoon, line 37: It's interesting that the spec go implementation doesn't include file name.

content/oci/oci.go
```diff
@@ -0,0 +1,227 @@
...
+		indexPath: filepath.Join(root, ociImageIndexFile),
+	}
+
+	if err := store.validateOCILayoutFile(); err != nil {
+		return nil, err
+	}
+
+	if err := store.loadIndex(); err != nil {
+		return nil, err
+	}
```
northtyphoon, line 67: What component is supposed to do the initialization to create the layerout and index file?

content/oci/oci.go
```diff
@@ -0,0 +1,227 @@
...
+	if layout.Version != ocispec.ImageLayoutVersion {
+		return errdef.ErrUnsupportedVersion
+	}
+
+	return nil
+}
+
+// load reads the index.json from the file system.
+func (s *Store) loadIndex() error {
```
akashsinghal, line 170: Just to curious as to why `loadIndex` and 'saveIndex` are not being exported anymore? If I remember correctly, the other OCI target had it exported and it could be used outside.

Wwwsylvia: Do we need to manually call `loadIndex` and `saveIndex` somewhere outside? The OCI does it automatically when tagging and resolving.

content/oci/storage.go
Wwwsylvia: No, but this will be mased by [umask](https://man7.org/linux/man-pages/man2/umask.2.html)

content/oci/oci.go
Wwwsylvia: Nice catch!

content/oci/storage.go
Wwwsylvia: This parameter is defined in the methods of the [Storage interfaces](https://github.com/oras-project/oras-go/blob/main/content/storage.go#L25-L38). The OCI storage does not use context, but other targets may use it.

content/oci/oci.go
Wwwsylvia: One may want to resolve by a digest if they have tagged by digest. Besides, we need the resolver map to construct index.json in `saveIndex()` https://github.com/Wwwsylvia/oras-go/blob/oci_target/content/oci/oci.go#L219-L222

content/oci/oci.go
Wwwsylvia: The layout file is initialized in [validateOCILayoutFile()](https://github.com/Wwwsylvia/oras-go/blob/oci_target/content/oci/oci.go#L155) when loading. The index file will be initialized on first tag in [saveIndex](https://github.com/Wwwsylvia/oras-go/blob/oci_target/content/oci/oci.go#L230)

northtyphoon: I see. It is not clear from the name that `validateOCILayeroutFile` creates the layout file if it doesn't exist. Maybe you can rename it something like validateOrCreateOCILayoutFile

northtyphoon: Do you know where I can find a code snippet example of copy an oci artifact from registry to local?

content/oci/oci.go
```diff
@@ -0,0 +1,228 @@
...
+		// tag by digest if tag is not present.
+		if desc.Annotations == nil || desc.Annotations[ocispec.AnnotationRefName] == "" {
+			s.resolver.Tag(ctx, desc, desc.Digest.String())
+		} else {
+			tag := desc.Annotations[ocispec.AnnotationRefName]
+			s.resolver.Tag(ctx, desc, tag)
+		}
+
+		// traverse the whole DAG and index up edges for all the nodes.
+		s.finder.IndexAll(ctx, s.storage, desc)
```
shizhMSFT, line 200: What if a node has already been indexed? For example, manifest A references manifest B and both A and B exists in the index file. Will manifest B be indexed twice?

content/oci/oci.go
```diff
@@ -0,0 +1,228 @@
...
+			return fmt.Errorf("digest reference mismatch: %s: expect %s: %w", reference, desc.Digest.String(), errdef.ErrInvalidReference)
+		}
+	} else {
+		if desc.Annotations == nil {
+			desc.Annotations = map[string]string{}
+		}
+		desc.Annotations[ocispec.AnnotationRefName] = reference
+	}
+
+	s.resolver.Tag(ctx, desc, reference)
```
shizhMSFT, line 118: Although it is a memory resolver and will never return an error, we still need to write the code to check the error.

content/oci/oci.go
shizhMSFT, line 200: It seems that the return values (error) of `Tag()` and `IndexAll()` are not checked.

shizhMSFT (changes requested):

content/memory/memory.go
```diff
@@ -18,28 +18,28 @@ import (
...
-	upEdges  sync.Map // map[descriptor.Descriptor]map[descriptor.Descriptor]ocispec.Descriptor
+	finder   *graph.Memory
 }
 
 // New creates a new memory based store.
 func New() *Store {
 	return &Store{
 		storage:  cas.NewMemory(),
 		resolver: resolver.NewMemory(),
+		finder:   graph.NewMemory(),
```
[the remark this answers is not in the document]
Wwwsylvia: Make sense! (naming is hard...)

content/oci/storage.go
```diff
@@ -0,0 +1,174 @@
...
+func (s *Storage) Push(_ context.Context, expected ocispec.Descriptor, content io.Reader) error {
+	target, err := s.blobPath(expected.Digest)
+	if err != nil {
+		return fmt.Errorf("%s: %s: %w", expected.Digest, expected.MediaType, errdef.ErrInvalidDigest)
+	}
+
+	// check if the target content exists in the blob directory.
+	if _, err := os.Stat(target); err == nil {
+		return fmt.Errorf("%s: %s: %w", expected.Digest, expected.MediaType, errdef.ErrAlreadyExists)
+	}
```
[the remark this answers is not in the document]
Wwwsylvia: Thanks for catching this!

content/oci/oci.go
```diff
@@ -0,0 +1,228 @@
...
+// based on file system with the OCI-Image layout.
+// Reference: https://github.com/opencontainers/image-spec/blob/master/image-layout.md
+type Store struct {
+	storage  content.Storage
+	resolver *reslv.Memory
+	finder   *graph.Memory
+	index    *ocispec.Index
+
+	root      string
+	indexPath string
```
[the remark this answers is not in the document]
Wwwsylvia: We need the `indexPath` when loading and saving index file.

internal/ioutil/io.go
```diff
@@ -44,12 +52,39 @@ func ReadAll(r io.Reader, desc ocispec.Descriptor) ([]byte, error) {
...
+// Copy copies from src to dst until either EOF is reached on src or an error occurs.
+// The copied content is verified against the size and the digest.
+func Copy(dst io.Writer, src io.Reader, desc ocispec.Descriptor) error {
+	// verify while copying
+	verifier := desc.Digest.Verifier()
+	mw := io.MultiWriter(verifier, dst)
+	buf := bufPool.Get().(*[]byte)
+	defer bufPool.Put(buf)
+
+	if _, err := io.CopyBuffer(mw, src, *buf); err != nil {
```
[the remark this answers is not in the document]
Wwwsylvia: good idea!

content/oci/oci.go
Wwwsylvia: We shall update the [examples](https://github.com/Wwwsylvia/oras-go/tree/oci_target/examples) once the implementation is done... Before that, @akashsinghal has written some code to copy some artifacts from OCI target to remote registry, which is helpful - https://github.com/akashsinghal/accelerated-container-image/blob/47ef7a728bed74020ef48814e8efb461f3db0236/cmd/ctr/overlaybd_conv.go#L740-L781

content/oci/oci.go
Wwwsylvia: @shizhMSFT Yeah it makes sense that we provide both ways and users can decide to auto save the index or not. I just exported the `SaveIndex()` method, and added a `AutoSaveIndex` option in the OCI store, and make it default to true. The users can set it to false if they want to manually save the index like:
```go
s := oci.New(root)
s.AutoSaveIndex = false
// do something

s.Tag(foo)
s.Tag(bar)

// now save the index
s.SaveIndex()
```

content/oci/oci.go
shizhMSFT: Linking to #60

content/oci/oci.go
```diff
@@ -0,0 +1,228 @@
...
+	if isDigest(reference) {
+		// a digest reference must match the digest of the descriptor
+		if reference != desc.Digest.String() {
+			return fmt.Errorf("digest reference mismatch: %s: expect %s: %w", reference, desc.Digest.String(), errdef.ErrInvalidReference)
+		}
+	} else {
+		if desc.Annotations == nil {
+			desc.Annotations = map[string]string{}
+		}
+		desc.Annotations[ocispec.AnnotationRefName] = reference
```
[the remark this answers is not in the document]
Wwwsylvia: This is a bit tricky. Actually I don't see a need to store the FQDN, we may open an issue to discuss it if needed.
`containerd` stores the FQDN in the `io.containerd.image.name` annotation, but it would be strange if we use the name containing "containerd". We may need to come up with a new name.
Besides, we store the passed reference string in `org.opencontainers.image.ref.name` annotation, which is supposed to be a tag (or digest), how do we know the FQDN of the descriptor?

content/oci/oci.go
```diff
@@ -0,0 +1,228 @@
...
+
+	return s.finder.Index(ctx, s.storage, expected)
+}
+
+// Exists returns true if the described content exists.
+func (s *Store) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+	return s.storage.Exists(ctx, target)
+}
+
+// Tag tags a descriptor with a reference string.
```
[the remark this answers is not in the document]
Wwwsylvia: I noticed that for digest-referenced manifest, `containerd`  stores the digest in the `org.opencontainers.image.ref.name` annotation as `"org.opencontainers.image.ref.name":  "@sha256:123abc". I'm going for this way.

content/oci/oci.go
Wwwsylvia: Yeah there was duplicate work... Fixed in `graph.memory` .

content/oci/oci.go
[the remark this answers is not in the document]
shizhMSFT: What if the user invokes `s.Tag(ctx, desc, "localhost:5000/myrepo:mytag")` since it is a common scenario in `Copy()`. Will you parse it and only store `mytag` in `org.opencontainers.image.ref.name`?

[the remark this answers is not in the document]
shizhMSFT: or consider `localhost:5000/myrepo:mytag` as an invalid reference?

content/oci/oci.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return nil
+}
+
+// loadIndex reads the index.json from the file system.
+func (s *Store) loadIndex(ctx context.Context) error {
+	indexFile, err := os.Open(s.indexPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to open index file: %v", err)
```
shizhMSFT, line 191: nit: it will be better to use `%w` instead of `%v` for wrapping errors. Same comment will be applied to all similar occourances.

shizhMSFT (changes requested):

content/oci/oci.go
shizhMSFT: Please see [fmt.Errorf()](https://pkg.go.dev/fmt#Errorf) for the documentation of `%w`.

content/oci/storage.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return false, err
+}
+
+// ingest write the content into a temporary ingest file.
+func (s *Storage) ingest(expected ocispec.Descriptor, content io.Reader) (path string, ingestErr error) {
+	if err := ensureDir(s.ingestRoot); err != nil {
+		ingestErr = fmt.Errorf("failed to ensure ingest dir: %v", err)
+		return
```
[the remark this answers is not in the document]
Wwwsylvia: ah good to learn this!

content/oci/oci.go
Wwwsylvia: > What if the user invokes s.Tag(ctx, desc, "localhost:5000/myrepo:mytag") since it is a common scenario in Copy(). Will you parse it and only store mytag in org.opencontainers.image.ref.name? or consider localhost:5000/myrepo:mytag as an invalid reference? 

We need to store the original user-provided reference in the `org.opencontainers.image.ref.name` as we use it to construct ref map when loading index. As per the [OCI image spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md), a reference like "localhost:5000/myrepo:mytag" is valid in grammar.
For the common `Copy` scenarios, isn't it like
```go
// do something
ref := "mytag"
err := oci.Push(ctx, desc, content)
err := oci.Tag(ctx, desc, ref)

repo := "localhost:5000/myrepo"
repository, err := remote.NewRepository(repo)
// do something
root, err := oras.Copy(ctx, oci, ref, repository, ref)
```
? The user would call `oci.Tag(ctx, desc, "mytag") ` instead of `oci.Tag(ctx, desc, "localhost:5000/myrepo:mytag") `

internal/graph/memory.go
```diff
@@ -0,0 +1,146 @@
...
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/internal/descriptor"
+	"oras.land/oras-go/v2/internal/status"
+)
+
+// Memory is a memory based UpEdgeFinder.
+type Memory struct {
+	upEdges sync.Map // map[descriptor.Descriptor]map[descriptor.Descriptor]ocispec.Descriptor
+	indexed sync.Map // map[descriptor.Descriptor]bool
```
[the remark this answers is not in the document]
Wwwsylvia: `upEdges` represents a node and its upEdges, while `indexed` represents that if a node has been recorded in `upEdges` as an upEdge.
The `indexed` is only checked in the `IndexAll()` so I didn't update it in `Index()`... But what if `Index()` is called first before `IndexAll()`? Maybe it's safer to update it in `Index()` as well.

internal/graph/memory.go
```diff
@@ -0,0 +1,146 @@
...
+		// skip the node if it has been indexed
+		key := descriptor.FromOCI(desc)
+		_, exists := m.indexed.Load(key)
+		if exists {
+			// mark the content as done
+			close(done)
+			return nil, ErrSkipDesc
+		}
+
+		return content.DownEdges(ctx, fetcher, desc)
```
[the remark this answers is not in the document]
Wwwsylvia: Right!

content/oci/oci.go
shizhMSFT: How about this scenario?
```golang
ref := "localhost:5000/hello:v1"
src, _ := remote.NewRepository(ref)
dst, _ := oci.New("hello")
oras.Copy(context.Background(), src, ref, dst, "")
```

internal/graph/memory.go
```diff
@@ -0,0 +1,137 @@
...
+		if err := m.index(ctx, desc, downEdges); err != nil {
+			return nil, err
+		}
+
+		return
+	})
+
+	postHandler := HandlerFunc(func(ctx context.Context, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+		return nil, nil
+	})
```
shizhMSFT, line 96: nit: curious if we can do
```golang
postHandler := Handlers()
```

internal/graph/memory.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// prepare pre-handler
+	preHandler := HandlerFunc(func(ctx context.Context, desc ocispec.Descriptor) (downEdges []ocispec.Descriptor, err error) {
+		defer func() {
+			if err == nil {
+				// mark the node as done on success
+				done, _ := tracker.TryCommit(desc)
+				close(done)
+			}
+		}()
```
shizhMSFT, line 65: nit: it seems can be simplified by removing this `defer` block and add `close(done)` at line 90.

shizhMSFT (approved):
LGTM with suggestions. Please also convert the reference discussion into an issue so that we can address later.

content/oci/oci.go
Wwwsylvia: I think it's ok. In this case we will have `"org.opencontainers.image.ref.name"`: `"localhost:5000/hello:v1"`.
The [OCI image spec](https://github.com/opencontainers/image-spec/blob/main/image-layout.md#indexjson-file) mentions that
> No semantic restriction is given for the "org.opencontainers.image.ref.name" annotation of descriptors.
So it's make sense to store FQDN in `org.opencontainers.image.ref.name`.

In terms of API usage, users can still do 
```go
desc, _ := dst.Resolve(ctx, "localhost:5000/hello:v1")
```

shizhMSFT (approved):
LGTM

sajayantony:
@shizhMSFT please squash merge. Let’s discuss the API usability to consume these and clean up before release.

jdolitsky (approved):
Lot to review, but nothing jumps out at me to prevent merge

PR #87, opened 2021-12-29
Fixed registry client token

sabre1041:
Corrected an issue where token returned from registry was not being managed properly


sajayantony:
@shizhMSFT @Wwwsylvia @deitch

shizhMSFT (approved):
Good catch. LGTM.

PR #92, opened 2022-01-11
Enable fallback on Docker config using custom auth file

jdolitsky:
related to https://github.com/helm/helm/pull/10536
cc @scottrigby

### DIFF
--- a/pkg/auth/docker/client.go
+++ b/pkg/auth/docker/client.go
@@ -64,6 +64,39 @@ func NewClient(configPaths ...string) (auth.Client, error) {
 	}, nil
 }
 
+// NewClientWithDockerFallback creates a new auth client
+// which falls back on Docker's default config path.
+// This allows support for ~/.docker/config.json as a fallback,
+// as well as support for the DOCKER_CONFIG environment variable.
+func NewClientWithDockerFallback(configPaths ...string) (auth.Client, error) {
+	if len(configPaths) == 0 {
+		return NewClient()
+	}
+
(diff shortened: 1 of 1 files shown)


sajayantony (approved):

sajayantony:
@jdolitsky @Wwwsylvia should this be in backported to main or has main removed need for docker-cli and it way more complicated now?

pkg/auth/docker/client.go
```diff
@@ -64,6 +64,39 @@ func NewClient(configPaths ...string) (auth.Client, error) {
 	}, nil
 }
 
+// NewClientWithDockerFallback creates a new auth client
+// which falls back on Docker's default config path.
+// This allows support for ~/.docker/config.json as a fallback,
+// as well as support for the DOCKER_CONFIG environment variable.
+func NewClientWithDockerFallback(configPaths ...string) (auth.Client, error) {
+	if len(configPaths) == 0 {
+		return nil, errors.New("must provide at least one config path")
```
sabre1041, line 73: Would we want to call `NewClient` when no config paths have been provided? Or is this a way to enforce the end user call `NewClient` if they do not provide any `configPaths`

jdolitsky: Yes, done

shizhMSFT:
> should this be in backported to main or has main removed need for docker-cli and it way more complicated now?

This should not be backported to `main` as `main` will remove `pkg/` eventually.

PR #101, opened 2022-01-20
Fix pushing to private registries

### LINKED ISSUE #97: Fail to push images to private registries
`PUT` requests are attempted twice. The body is read for auth and gets empty for the retried requests.

shizhMSFT:
Fix #97

### DIFF
--- a/registry/remote/auth/client.go
+++ b/registry/remote/auth/client.go
@@ -199,6 +199,9 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
 			if token, err := cache.GetToken(ctx, registry, SchemeBearer, key); err == nil {
 				req = originalReq.Clone(ctx)
 				req.Header.Set("Authorization", "Bearer "+token)
+				if err := rewindRequestBody(req); err != nil {
+					return nil, err
+				}
 
 				resp, err := c.send(req)
 				if err != nil {
@@ -225,6 +228,9 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
 		req.Header.Set("Authorization", "Bearer "+token)
 	default:
 		return resp, nil
+	}
+	if err := rewindRequestBody(req); err != nil {
+		return nil, err
 	}
 
 	return c.send(req)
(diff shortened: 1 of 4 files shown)


registry/remote/auth/client.go
```diff
@@ -365,3 +371,19 @@ func (c *Client) fetchOAuth2Token(ctx context.Context, realm, service string, sc
...
+
+// rewindRequestBody tries to rewind the request body if exists.
+func rewindRequestBody(req *http.Request) error {
+	if req.Body == nil || req.Body == http.NoBody {
+		return nil
+	}
+	if req.GetBody == nil {
+		return fmt.Errorf("%s %q: request body is not rewindable", req.Method, req.URL)
+	}
+	body, err := req.GetBody()
```
sajayantony, line 383: I would have expected this to actually call a seek. Is that possible without invoking getbody again or we ensuring that this is used only in the authentication flow?

registry/remote/repository.go
```diff
@@ -195,7 +203,26 @@ func (r *Repository) push(ctx context.Context, expected ocispec.Descriptor, cont
...
+	// and serve from the memory.
+	client := r.client()
+	if _, ok := client.(*auth.Client); ok && req.GetBody == nil {
+		store := cas.NewMemory()
+		err := store.Push(ctx, expected, content)
+		if err != nil {
+			return err
+		}
+		req.GetBody = func() (io.ReadCloser, error) {
+			return store.Fetch(ctx, expected)
```
sajayantony, line 218: Is there any call path that would expose this to storing large chuncks in memory?

shizhMSFT: No, this memory usage is scoped to this function and will be released after function return. Since it is only for manifests, the memory usage will be small.

registry/remote/auth/client.go
shizhMSFT: This is the common method to "seek" to be begining. The [transport.go](https://github.com/golang/go/blob/go1.17.6/src/net/http/transport.go#L659-L675) in the built-in golang package `net/http` does the same way.

sajayantony (approved):

PR #104, opened 2022-01-21
Update Makefile and scripts

### LINKED ISSUE #61: Update Makefile and scripts
We need to update Makefile and scripts once we revamp the `oras-go` library.

shizhMSFT:
Resolves #61.

- Update and clean up Makefile.
- Fix test script.
- Detect examples instead of hard coding.
- Update go version to `1.17.6` for testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,13 +16,13 @@ clean:
 
 .PHONY: check-encoding
 check-encoding:
-	! find . -name "*.go" -type f -exec file "{}" ";" | grep CRLF
+	! find . -not -path "./vendor/*" -name "*.go" -type f -exec file "{}" ";" | grep CRLF
 	! find scripts -name "*.sh" -type f -exec file "{}" ";" | grep CRLF
 
 .PHONY: fix-encoding
 fix-encoding:
-	find . -type f -name "*.go" -exec sed -i -e "s/\r//g" {} +
-	find scripts -type f -name "*.sh" -exec sed -i -e "s/\r//g" {} +
+	find . -not -path "./vendor/*" -name "*.go" -type f -exec sed -i -e "s/\r//g" {} +
+	find scripts -name "*.sh" -type f -exec sed -i -e "s/\r//g" {} +
 
 .PHONY: vendor
 vendor:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -19,41 +16,14 @@ clean:
 
 .PHONY: check-encoding
 check-encoding:
-	! find pkg examples -name "*.go" -type f -exec file "{}" ";" | grep CRLF
+	! find . -name "*.go" -type f -exec file "{}" ";" | grep CRLF
```
deitch, line 19: Given that later we call `go mod vendor`, this could search all of the vendored files as well. If you can exclude `vendor/` from it, that would work.

Makefile
```diff
@@ -19,41 +16,14 @@ clean:
...
 	! find scripts -name "*.sh" -type f -exec file "{}" ";" | grep CRLF
 
 .PHONY: fix-encoding
 fix-encoding:
-	find pkg examples -type f -name "*.go" -exec sed -i -e "s/\r//g" {} +
+	find . -type f -name "*.go" -exec sed -i -e "s/\r//g" {} +
```
deitch, line 24: Same as previous

scripts/acceptance.sh
```diff
@@ -27,8 +27,10 @@ docker rm -f oras-acceptance-registry || true
 # Build the examples into binaries
 (
   cd examples
-  CGO_ENABLED=0 go build -v -o ../bin/oras-acceptance-simple ./simple
-  CGO_ENABLED=0 go build -v -o ../bin/oras-acceptance-advanced ./advanced
+  for example in */; do
+    example=${example%/}
+    CGO_ENABLED=0 go build -v -o ../bin/oras-acceptance-$example ./$example
```
deitch, line 32: I really like this. We probably should properly protect the var as `./${example}` (even if both are ugly to read).

shizhMSFT: Protected `$example` by changing to `${example}`.

Makefile
shizhMSFT: Good catch. I've excluded `vendor` in `find` in the updated commit be07485.

Makefile
shizhMSFT: Resolved in be07485.

deitch (approved):
Whenever you are ready, merge it in.

PR #112, opened 2022-03-11
Add registry target example for oras-go V2

qweeah:
This PR is for #60, which added example code for below remote registry actions:

1. Upload blobs
2. Upload a manifest 
3. Download blobs
4. Download a manifest
6. Tag a manifest
7. Upload manifest with reference
8. Resolve reference
9. Copy a manifest
10. List registry catalog and tags


deitch:
@qweeah thanks for this.

1. You need to sign the DCO
2. Why did you delete the existing examples, as opposed to adding a new one?

qweeah:
@deitch Thanks for reminding, signed and repushed.

> Why did you delete the existing examples, as opposed to adding a new one?

We are working on the V2 library and the old code in `pkg` folder will deprecated, so do the old examples

deitch:
No issues with replacing the examples, but the new ones need to be at least as understandable.

shizhMSFT:
@deitch Thanks for reviewing this PR.

Billy Zha (@qweeah) is ramping up and handing on writing examples for oras-go v2. This draft PR is still **in an early stage**. Once Billy is comfortable with this PR, he will convert it into a formal PR ready to be reviewed.

sajayantony:
I think I agree with @deitch - Let's not delete the files in this PR and drop them as a part of removing the pkg folder.

examples/registry/push_pull_copy.go
```diff
@@ -0,0 +1,192 @@
+// This example shows how to use oras go library to do some
```
sajayantony, line 1: Consider splitting these up into smaller example files which target only one scenarios. 
For e.g. a download manifest should be a few lines of code in the file.

qweeah: Since the scenarios have dependencies on each other (e.g. pulling a blob requires pushing it first, resolving a tag requires pushing an tagging first), we still need to put all the registry-related scenarios in a single file, otherwise there will be to much redundant code.

For each scenario, I have added ad-hoc function in the `function.go` file. If a user need to copy some sample snippet to implement their own go app, they can copy the function directly.

If you have any comment or concern, please let me know.

qweeah:
Will move the examples to go docs. Will do that in a new PR.

PR #115, opened 2022-03-15
fix memory graph

### LINKED ISSUE #110: internal/graph/memory.go improvement
In the `IndexAll()` function, the `done` channel is not actually used. We should improve the code.
See https://github.com/oras-project/oras-go/blob/main/internal/graph/memory.go#L60

Wwwsylvia:
Fixes #110 


shizhMSFT (approved):
LGTM

PR #119, opened 2022-03-18
Add examples for repository push, pull, resolve and tags

qweeah:
This PR works for item #60 and adds examples for:
 Push & pull blobs
 Tag a image
 Resolve the tag
 List tags in a repository


registry/remote/example_test.go
```diff
@@ -0,0 +1,312 @@
...
+		panic(err) // Handle error
+	}
+
+	mediaType, content := ocispec.MediaTypeImageLayer, []byte("Example blob content") // Setup input: 1) media type and 2)[]byte content
+	desc := ocispec.Descriptor{                                                       // Assemble a descriptor
+		MediaType: mediaType,                 // Set mediatype
+		Digest:    digest.FromBytes(content), // Calculate digest
+		Size:      int64(len(content)),       // Include content size
+	}
+	repo.Push(ctx, desc, bytes.NewReader(content)) // Push the blob
```
sajayantony, line 137: Can we use the output from push?

registry/remote/example_test.go
```diff
@@ -0,0 +1,312 @@
...
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
+	if err != nil {
+		panic(err) // Handle error
+	}
+	digest := "sha256:aafc6b9fa2094cbfb97eca0355105b9e8f5dfa1a4b3dbe9375a30b836f6db5ec"
+	descriptor, err := repo.Resolve(ctx, digest) // Resolve digest to the descriptor
+	if err != nil {
+		panic(err) // Handle error
+	}
+	fmt.Println(descriptor.Size)
```
sajayantony, line 185: Would recommend adding the mediaType since it’s typically used more than size. But showing both would be nice.

registry/remote/example_test.go
qweeah: The push only returns error. I can add handlers for returned error and only output `succeed` when there is no error.

registry/remote/example_test.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/registry/remote"
+)
+
+const exampleRepositoryName = "example"
+const exampleDigest = "sha256:aafc6b9fa2094cbfb97eca0355105b9e8f5dfa1a4b3dbe9375a30b836f6db5ec"
+const exampleTag = "latest"
+const exampleBlob = "Example blob content"
+const exampleUploadUUid = "0bc84d80-837c-41d9-824e-1907463c53b3"
```
shizhMSFT, line 39: nit: make it a const block like
```go
const (
  exampleRepositoryName = "example"
  exampleDigest         = "sha256:aafc6b9fa2094cbfb97eca0355105b9e8f5dfa1a4b3dbe9375a30b836f6db5ec"
  // ...
)
```

registry/remote/example_test.go
```diff
@@ -0,0 +1,326 @@
...
+		panic(err) // Handle error
+	}
+
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	repo.Tags(ctx, func(tags []string) error {
```
shizhMSFT, line 112: error check required.

registry/remote/example_test.go
```diff
@@ -0,0 +1,326 @@
...
+	descriptor, err := repo.Resolve(ctx, tag) // First resolve the tag to the descriptor
+	if err != nil {
+		panic(err) // Handle error
+	}
+	r, err := repo.Fetch(ctx, descriptor) // Fetch the blob from the repository
+	if err != nil {
+		panic(err) // Handle error
+	}
+	defer r.Close() // don't forget to close
+	pulledBlob, err := io.ReadAll(r)
```
shizhMSFT, line 227: Use `content.FetchAll()` for validation

registry/remote/example_test.go
```diff
@@ -0,0 +1,326 @@
...
+	descriptor, err := repo.Resolve(ctx, digest) // We still need to resolve first, don't create a new descriptor with the digest, blob size is unknown
+	if err != nil {
+		panic(err) // Handle error
+	}
+	r, err := repo.Fetch(ctx, descriptor) // Fetch the blob from the repository
+	if err != nil {
+		panic(err) // Handle error
+	}
+	defer r.Close() // don't forget to close
+	pulledBlob, err := io.ReadAll(r)
```
shizhMSFT, line 259: same here

shizhMSFT (changes requested):

registry/remote/example_test.go
qweeah: Good catch. Will the function return non-nil error even when the callback only returns nil?

registry/remote/example_test.go
qweeah: Should we use it as validation here?
I can add a new example for `FetchAll()` if that is what you want

Wwwsylvia:
We will need examples for `FetchReference` once https://github.com/oras-project/oras-go/pull/124 gets merged.

registry/remote/example_test.go
qweeah: Resolved. Added digest validation here and new examples for content.FetchAll()

content/example_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package content_test gives examples code of functions in the content package.
+package content_test
```
shizhMSFT, line 17: All examples in `content/example_test.go` should be merged to `registry/remote/example_test.go`. Ideally, no `registry` should be referenced in the `content` package.

registry/remote/example_test.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		panic(err) // Handle error
+	}
+	fmt.Println(string(pulledBlob))
+	// **You still need to validate the digest**
+	if descriptor.Digest != ocidigest.FromBytes(pulledBlob) {
```
shizhMSFT, line 254: and also the size.

registry/remote/example_test.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	defer r.Close() // don't forget to close
+
+	// You can 1) fetch a whole blob
+	pulledBlob, err := io.ReadAll(r)
+	if err != nil {
+		panic(err) // Handle error
+	}
+	fmt.Println(string(pulledBlob))
+	// Fetch() doesn't provide validation, you need to do it **
```
shizhMSFT, line 291: Try using terms defined in [BCP 14](https://tools.ietf.org/search/bcp14). In this case, callers **SHOULD** verify what they have fetched.

registry/remote/example_test.go
```diff
@@ -0,0 +1,371 @@
...
+	// Succeed
+}
+
+// ExampleRegistry_Repositories gives example snippets for listing respositories in a HTTPS registry with pagination.
+func ExampleRegistry_Repositories() {
+	reg, err := remote.NewRegistry(host)
+	if err != nil {
+		panic(err) // Handle error
+	}
+	// If you want to play with your local registry, try to override
```
shizhMSFT, line 353: nit: Try using objective wording and try to avoid words like _you_, _he_, _she_, _I_, _we_.

content/example_test.go
qweeah: But this is only content_test, can't it reference `registry` either? If I move everything to the `registry` package then godoc won't render example for FetchAll.

registry/remote/example_test.go
qweeah: This is really helpful

content/example_test.go
qweeah: Moved to sceonario-based examples in remote package

registry/remote/example_test.go
```diff
@@ -0,0 +1,424 @@
...
+		panic(err) // Handle error
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	tag := "latest"
+	descriptor, err := repo.Resolve(ctx, tag) // Resolve digest to the descriptor
```
Wwwsylvia, line 187: > // Resolve digest to the descriptor

digest? Do you mean tag?

registry/remote/example_test.go
```diff
@@ -0,0 +1,424 @@
...
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	// suppose we are going to tag a blob with below digest
+	digest := "sha256:aafc6b9fa2094cbfb97eca0355105b9e8f5dfa1a4b3dbe9375a30b836f6db5ec"
+
+	// 1. Resolve the target desc
```
Wwwsylvia, line 331: nit: The variable name is `descriptor`, but the comment calls it `desc`.
It's `descriptor` in other comments.

registry/remote/example_test.go
```diff
@@ -0,0 +1,424 @@
...
+	if err != nil {
+		panic(err) // Handle error
+	}
+	// Output:
+	// public/repo1
+	// public/repo2
+	// internal/repo3
+}
+
+func Example_pullByTag() {
```
Wwwsylvia, line 374: What is the different between this and the `ExampleRepository_Fetch_byTag()`?  Just the use of `Fetch()` and `FetchAll()`?
The name makes me feels like this function is to "pull" the whole image, or do something different than "fetch"...

qweeah: Since the `FetchAll` belongs to the package content, we can only use scenario suffix `pullByXXX` here

qweeah:
> The idea is that, when communicating with remote registries, users may think of "how do I get a manifest" and "how do I get a layer". It could be confusing to generally call everything a blob.
Thanks for the advice. In the example, I pulled a tagged layer from the registry, 
1) Is it the right thing to show to oras-go user?
2) Comparing to call it a blob, will it be less confusing if I call it a layer?

> Also, functions like `Repository.Fetch()` , `Repository.Push()`, etc. can fetch/push both manifest and blobs, while `Repository.Resolve()` only resolves manifests, should we clarify a bit on this? Users may end up resolving a blob using `Repository.Resolve()`.

I think manifest is also a blob with specific media type, which should also be the common sense of oras-go user (correct me if I am wrong). Thus we don't need to clarify that, at least not in the example.

User can also resolve the descriptor of a (layer or config) blob by `Repository.Resolve()`

Wwwsylvia:
> In the example, I pulled a tagged layer from the registry

You actually pulled a manifest.
Ah I see what's not right, you used the same digest for mocking `/v2/<repository>/manifests/<reference>` API and `/v2/<repository>/blobs/<reference>` API, so you didn't notice which API you were accessing. You may refer to [TestRepository_Fetch](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository_test.go#L72-L88) for the mocks.

> I think manifest is also a blob with specific media type, which should also be the common sense of oras-go user (correct me if I am wrong). Thus we don't need to clarify that, at least not in the example.

Yeah but when querying remote registries, manifests and blobs go through different API endpoints. Just that functions like [Repository.Fetch()](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L122-L125) automatically choose endpoints based on the media type.
For users familiar with [Docker V2 API](https://docs.docker.com/registry/spec/api/#detail), they may want to know how the oras-go functions correspond to the V2 APIs? (Just a thought)

> User can also resolve the descriptor of a (layer or config) blob by `Repository.Resolve()`

No they can't.
[Repository.Resolve()](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L153-L157) calls [Manifests().Resolve()](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L752-L754) behind the scenes, which queries the `/v2/<repository>/manifests/<reference>` API (See https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L760). So it only resolves manifests.
```go
// Resolve resolves a reference to a manifest descriptor.
// See also `ManifestMediaTypes`.
func (r *Repository) Resolve(ctx context.Context, reference string) (ocispec.Descriptor, error) {
	return r.Manifests().Resolve(ctx, reference)
}
```

qweeah:
> > In the example, I pulled a tagged layer from the registry
> 
> You actually pulled a manifest. Ah I see what's not right, you used the same digest for mocking `/v2/<repository>/manifests/<reference>` API and `/v2/<repository>/blobs/<reference>` API (https://github.com/qweeah/oras-go/blob/repository-example/registry/remote/example_test.go#L77-L82), so you didn't notice which API you were accessing. You may refer to [TestRepository_Fetch](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository_test.go#L72-L88) for the mocks.
> 
> > I think manifest is also a blob with specific media type, which should also be the common sense of oras-go user (correct me if I am wrong). Thus we don't need to clarify that, at least not in the example.
> 
> Yeah but when querying remote registries, manifests and blobs go through different API endpoints. Just that functions like [Repository.Fetch()](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L122-L125) automatically choose endpoints based on the media type. For users familiar with [Docker V2 API](https://docs.docker.com/registry/spec/api/#detail), they may want to know how the oras-go functions correspond to the V2 APIs? (Just a thought)
> 
> > User can also resolve the descriptor of a (layer or config) blob by `Repository.Resolve()`
> 
> No they can't. [Repository.Resolve()](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L153-L157) calls [Manifests().Resolve()](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L752-L754) behind the scenes, which queries the `/v2/<repository>/manifests/<reference>` API (See https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L760). So it only resolves manifests.
> 
> ```go
> // Resolve resolves a reference to a manifest descriptor.
> // See also `ManifestMediaTypes`.
> func (r *Repository) Resolve(ctx context.Context, reference string) (ocispec.Descriptor, error) {
> 	return r.Manifests().Resolve(ctx, reference)
> }
> ```

Yes you are right. Looks like I mixed Repository.Resolve with BlobStore.Resolve. I will separate blob pull and manifest pull.

qweeah:
@Wwwsylvia I separated examples for manifest and layer fetch and added new examples for manifes FetchReference. Please take a review when you got time, thanks

registry/remote/example_test.go
```diff
@@ -0,0 +1,539 @@
...
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	descriptor := ocispec.Descriptor{
+		MediaType: ocispec.MediaTypeImageLayer,
+		Digest:    digest.FromBytes([]byte(exampleLayer)),
+		Size:      int64(len(exampleLayer)),
+	}
+	r, err := repo.Fetch(ctx, descriptor) // Fetch the blob from the repository
```
Wwwsylvia, line 385: You may want to use `Repository.Blobs().Resolve()` to get the descriptor of the layer? See [Test_BlobStore_Resolve](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository_test.go#L1421-L1424) and [Test_BlobStore_FetchReference](https://github.com/oras-project/oras-go/blob/main/registry/remote/repository_test.go#L1495-L1499).

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+	// Output:
+	// tag1
+	// tag2
+}
+
+// ExampleRepository_Push gives example snippets for pushing a layer.
+func ExampleRepository_Push() {
+	reg, err := remote.NewRegistry(host)
+	if err != nil {
+		panic(err) // Handle error
```
sajayantony, line 154:
```suggestion
                panic(err)
```

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+}
+
+// ExampleRepository_Fetch_byTag gives example snippets for downloading a manifest by tag.
+func ExampleRepository_Fetch_manifestByTag() {
+	reg, err := remote.NewRegistry(host)
+	if err != nil {
+		panic(err) // Handle error
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
```
sajayantony, line 238: Does this actually do any request. My understanding was that this is just parsing and creating a new repo object. Recommend removing the comment since the `.Repositry` initialization might be self explanatory. 
```suggestion
        repo, err := reg.Repository(ctx, exampleRepositoryName)
```


Similarly recommend reducing the comments at the end of the line unless really needed. 
Also would recommend moving comments to above the statements since the rendering is really wide in some cases. 

image

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	digest := "sha256:00e5ffa7d914b4e6aa3f1a324f37df0625ccc400be333deea5ecaa199f9eff5b"
+	descriptor, err := repo.Resolve(ctx, digest) // Still need to resolve first, don't create a new descriptor with the digest, blob size is unknown
```
sajayantony, line 280:
```suggestion
        //  Resolve the blob descriptor to obtain the size of the blob
	digest := "sha256:00e5ffa7d914b4e6aa3f1a324f37df0625ccc400be333deea5ecaa199f9eff5b"
	descriptor, err := repo.Resolve(ctx, digest)
```

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+		panic(err) // Handle error
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	digest := "sha256:00e5ffa7d914b4e6aa3f1a324f37df0625ccc400be333deea5ecaa199f9eff5b"
+	descriptor, r, err := repo.FetchReference(ctx, digest) // Fetch the blob from the repository
```
sajayantony, line 349:
```suggestion
        descriptor, rc, err := repo.FetchReference(ctx, digest)
```

This is not against the blob API btw. Also the comment is not required since you are clearly calling FetchReference by digest. I would recommend just dropping the comment. NIT: rename reader closer to `rc`.

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+	if err != nil {
+		panic(err) // Handle error
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	// Tag a manifest with below digest
```
sajayantony, line 438:
```suggestion
        // Tag a manifest referenced by the digest below
```

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+
+	tag := "latest"
+	descriptor, err := repo.Resolve(ctx, tag) // First resolve the descriptor
+	if err != nil {
+		panic(err) // Handle error
+	}
+	fmt.Println(descriptor.Digest)
+	fmt.Println(descriptor.Size)
+
+	pulledBlob, err := content.FetchAll(ctx, repo, descriptor) // Fetch the blob from the repository
```
sajayantony, line 501: Move all the comments above the call.

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+	// internal/repo3
+}
+
+func Example_pullByTag() {
+	reg, err := remote.NewRegistry(host)
+	if err != nil {
+		panic(err) // Handle error
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName) // Get the repository from registry
```
sajayantony, line 488: Same as above probably can remove the comment and there is no `GET` operation is my understanding. 
Try to see if you find all occurences of this pattern and remove the comment.

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+
+	digest := "sha256:00e5ffa7d914b4e6aa3f1a324f37df0625ccc400be333deea5ecaa199f9eff5b"
+	descriptor, err := repo.Resolve(ctx, digest) // Resolve the descriptor
+	if err != nil {
+		panic(err) // Handle error
+	}
+	fmt.Println(descriptor.Digest)
+	fmt.Println(descriptor.Size)
+
+	pulledBlob, err := content.FetchAll(ctx, repo, descriptor) // Fetch from the repository
```
sajayantony, line 532:
```suggestion
        // Fetch the content byte[] from the repository
        pulledBlob, err := content.FetchAll(ctx, repo, descriptor) 
```

registry/remote/example_test.go
qweeah: Removed unnecessary comments at the end of the line to make the doc looks better in HTML

registry/remote/example_test.go
```diff
@@ -0,0 +1,542 @@
...
+	r, err := repo.Fetch(ctx, descriptor) // Fetch the blob from the repository
+	if err != nil {
+		panic(err) // Handle error
+	}
+	defer r.Close() // don't forget to close
+	pulled, err := io.ReadAll(r)
+	if err != nil {
+		panic(err) // Handle error
+	}
+	// caller SHOULD verify what is fetched
```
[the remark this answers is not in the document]
qweeah: We need to emphrasize that caller should verify the content obtained via `Repository.Fetch`, what do you think @shizhMSFT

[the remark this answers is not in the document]
shizhMSFT: Both sentences describe the same meaning but in different wording.

registry/remote/example_test.go
```diff
@@ -0,0 +1,552 @@
...
+		panic(err)
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName)
+	if err != nil {
+		panic(err)
+	}
+
+	digest := "sha256:00e5ffa7d914b4e6aa3f1a324f37df0625ccc400be333deea5ecaa199f9eff5b"
+	descriptor, rc, err := repo.FetchReference(ctx, digest) // Fetch the blob from the repository
```
Wwwsylvia, line 357: Do we need this comment?

registry/remote/example_test.go
```diff
@@ -0,0 +1,552 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// Output:
+	// tag1
+	// tag2
+}
+
+// ExampleRepository_Push gives example snippets for pushing a layer.
```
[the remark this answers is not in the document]
qweeah: Pushing a manifest will require 60+ extra lines of code, including assemebling a manifest and push all the layer and config. I can add a new scenario-based example, but for this function-based example, let's keep it simple so the key snippet can be understood and reused by any reader.

Wwwsylvia (commented):
LGTM

sajayantony (approved):

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+		panic(err)
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName)
+	if err != nil {
+		panic(err)
+	}
+
+	// 1. assemble a descriptor
+	mediaType, content := ocispec.MediaTypeImageLayer, []byte("Example layer content")
```
shizhMSFT, line 164: Could we put `ocispec.MediaTypeImageLayer` directly in line 166?

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+	if err != nil {
+		panic(err)
+	}
+	defer rc.Close() // don't forget to close
+	pulledBlob, err := io.ReadAll(rc)
+	if err != nil {
+		panic(err)
+	}
+	// verify the fetched content
+	if descriptor.Digest != ocidigest.FromBytes(pulledBlob) || descriptor.Size != int64(len(pulledBlob)) {
```
shizhMSFT, line 263: nit: check size first then the digest.

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName)
+	if err != nil {
+		panic(err)
+	}
+
+	tag := "latest"
+	// specify the expected manifest media type for the blob
+	reg.ManifestMediaTypes = append(reg.ManifestMediaTypes, ocispec.MediaTypeImageManifest)
```
shizhMSFT, line 324: This is optional

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+	}
+	defer rc.Close() // don't forget to close
+	pulledBlob, err := io.ReadAll(rc)
+	if err != nil {
+		panic(err)
+	}
+	// verify the fetched content
+	if descriptor.Digest != ocidigest.FromBytes(pulledBlob) || descriptor.Size != int64(len(pulledBlob)) {
+		panic(err)
+	}
```
shizhMSFT, line 337: Looks like we have lots of duplicated code like this.

@Wwwsylvia What do you think if we promote [internal/ioutil.ReadAll()](https://github.com/oras-project/oras-go/blob/main/internal/ioutil/io.go#L37) to the `content` package?

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+	if err != nil {
+		panic(err)
+	}
+	ctx := context.Background()
+	repo, err := reg.Repository(ctx, exampleRepositoryName)
+	if err != nil {
+		panic(err)
+	}
+
+	descriptor, err := repo.Blobs().Resolve(ctx, digest.FromBytes([]byte(exampleLayer)).String())
```
shizhMSFT, line 390: `digest.FromBytes([]byte(exampleLayer)).String()` is confusing. Can we replace it with a pre-computed variable or constant?

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+	if err != nil {
+		panic(err)
+	}
+	rc, err := repo.Fetch(ctx, descriptor)
+	if err != nil {
+		panic(err)
+	}
+	defer rc.Close() // don't forget to close
+
+	// option 1: fetch a WHOLE blob
```
shizhMSFT, line 400:
```suggestion
	// option 1: sequential access
```

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+		panic(err)
+	}
+	fmt.Println(string(pulledBlob))
+
+	// verify the fetched content
+	if descriptor.Digest != ocidigest.FromBytes(pulledBlob) || descriptor.Size != int64(len(pulledBlob)) {
+		panic(err)
+	}
+
+	// option 2: partially read, if the remote registry supports
```
shizhMSFT, line 412:
```suggestion
	// option 2: random access, if the remote registry supports
```

registry/remote/example_test.go
```diff
@@ -0,0 +1,624 @@
...
+	digest := "sha256:00e5ffa7d914b4e6aa3f1a324f37df0625ccc400be333deea5ecaa199f9eff5b"
+	descriptor, err := repo.Resolve(ctx, digest)
+	if err != nil {
+		panic(err)
+	}
+	tag := "latest"
+	err = repo.Tag(ctx, descriptor, tag)
+	if err != nil {
+		panic(err)
+	}
```
shizhMSFT, line 456: @Wwwsylvia `Resolve` and `Tag` will have 3 requests in total. We probably need a more efficient one by having a generic `Tag()` function by leveraging `FetchReference()` and `PushReference()`.

registry/remote/example_test.go
Wwwsylvia: That would be useful.

registry/remote/example_test.go
Wwwsylvia: Make sense. 
Like `func (r *Repository) ReTag(ctx context.Context, srcRef, destRef string) error`? Need to figure out a new function name.

shizhMSFT: How about `func Tag(ctx context.Context, repo Repository, srcRef, dstRef string) error` in package `registry`?

shizhMSFT (approved):
LGTM

registry/remote/example_test.go
Wwwsylvia: If it's in package `registry`, it feels like it should instead support `func Tag(ctx context.Context, srcRep, dstRepo Repository, srcRef, dstRef string) error`

Wwwsylvia: OK

Wwwsylvia: @qweeah has created an issue for this: https://github.com/oras-project/oras-go/issues/120

registry/remote/example_test.go
Wwwsylvia: Issue created: #128

PR #121, opened 2022-03-22
Add example for copying

qweeah:
Signed-off-by: Billy Zha [jinzha1@microsoft.com](mailto:jinzha1@microsoft.com)

This PR works for item https://github.com/oras-project/oras-go/issues/60 and adds an example for copying


copy_test.go
```diff
@@ -352,3 +360,97 @@ func TestCopyGraph_PartialCopy(t *testing.T) {
...
+
+func ExampleCopy() {
+	ctx := context.Background()
+	ts1, sourceRepo, err1 := getTestRepository(ctx, "source")
+	ts2, targetRepo, err2 := getTestRepository(ctx, "target")
+	if err1 != nil || err2 != nil {
+		// handle it
+		return
+	}
+	defer ts1.Close()
```
sajayantony, line 422: 1. Remove ts1 and ts1 defer. Does not pertain to using this function - Is there any issue if we dont' defer it. I think the test framework will take it down, but you can confirm. 
2. use `reg.Respository(ctx. "source")` instead hidden function - See the generated docs and it will be hard to follow and users will wonder what is getRepository doing. 

I think if you use Testing Main then you can just have a var that has the `reg` testregistry.

qweeah:
Convert to draft since refactoring is still WIP

copy_test.go
qweeah: 1. If it's not deferred, then the test http server will be closed before receving mocked request, I have confirmed.
2. I will refactor the implementation like registry examples, remove the getTestRepository function and other stuff. Making this PR as draft before the refactoring is pushed.

shizhMSFT:
It is better to have Target to Target copy examples.

shizhMSFT:
At least, we need 4 types of examples:
1. Remote to Remote
2. Remote to Local
3. Local to Remote
4. Local to Local

You have done the first one. It is better to have the rest of three.

sajayantony (approved):

shizhMSFT (approved):
LGTM

PR #124, opened 2022-03-28
Add FetchReference() for Repository Target

### LINKED ISSUE #122: Add Function to Fetch Manifest Using Reference
- Currently, v2 requires a manifest descriptor to be resolved first before calling `Fetch` to get the artifact manifest
- Propose to add a function that will take in a reference and return the artifact manifest

Wwwsylvia:
Closes #122

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -24,6 +24,7 @@ import (
 	"mime"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -249,6 +250,11 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
 func (r *Repository) parseReference(reference string) (registry.Reference, error) {
 	ref, err := registry.ParseReference(reference)
 	if err != nil {
+		// reference is not a FQDN
+		if index := strings.IndexByte(reference, '@'); index != -1 {
+			// drop tag since the digest is present
+			reference = reference[index+1:]
+		}
 		ref = registry.Reference{
 			Registry:   r.Reference.Registry,
 			Repository: r.Reference.Repository,
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -241,48 +241,7 @@ func (r *Repository) push(ctx context.Context, expected ocispec.Descriptor, cont
 // It is equivalent to call `Resolve()` and then `Fetch()`
 // but more efficient or equal.
 func (r *Repository) FetchTag(ctx context.Context, reference string) (io.ReadCloser, error) {
-	// TODO: move to manifeststore
-	ref, err := r.parseReference(reference)
-	if err != nil {
-		return nil, err
-	}
-
-	ctx = withScopeHint(ctx, ref, auth.ActionPull)
-	url := buildRepositoryManifestURL(r.PlainHTTP, ref)
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return nil, err
-	}
-	req.Header.Set("Accept", manifestAcceptHeader(r.ManifestMediaTypes))
-
-	resp, err := r.client().Do(req)
-	if err != nil {
-		return nil, err
-	}
-	defer func() {
-		if err != nil {
-			resp.Body.Close()
-		}
-	}()
-
-	switch resp.StatusCode {
-	case http.StatusOK:
-		// no-op
-	case http.StatusNotFound:
-		return nil, fmt.Errorf("%s: %w", ref.Reference, errdef.ErrNotFound)
-	default:
-		return nil, errutil.ParseErrorResponse(resp)
-	}
-
-	// validate digest if the provided reference is a digest
-	if refDigest, err := ref.Digest(); err == nil {
-		if err = verifyContentDigest(resp, refDigest); err != nil {
-			return nil, err
-		}
-		return resp.Body, nil
-	}
-
-	return resp.Body, nil
+	return r.Manifests().FetchTag(ctx, reference)
 }
 
 // parseReference validates the reference.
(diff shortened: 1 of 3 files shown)


registry/repository.go
```diff
@@ -65,6 +65,7 @@ type BlobStore interface {
 	content.Storage
 	content.Deleter
 	content.Resolver
+	TagFetcher
```
shizhMSFT, line 68: Can we also make `Repository` implement `TagFetcher`, which defaults for manifests?

registry/remote/repository.go
```diff
@@ -601,6 +608,56 @@ func (s *blobStore) Resolve(ctx context.Context, reference string) (ocispec.Desc
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.repo.client().Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
```
shizhMSFT, line 636: which `err` are you referring to?

registry/repository.go
```diff
@@ -75,6 +76,14 @@ type TagPusher interface {
 	PushTag(ctx context.Context, expected ocispec.Descriptor, content io.Reader, reference string) error
 }
 
+// TagFetcher provides advanced fetch with the tag service.
+type TagFetcher interface {
+	// FetchTag fetches the content identified by a reference.
+	// It is equivalent to call `Resolve()` and then `Fetch()`
+	// but more efficient or equal.
+	FetchTag(ctx context.Context, reference string) (io.ReadCloser, error)
```
shizhMSFT, line 84: It is better to return
```go
(io.ReadCloser, ocispec.Descriptor, error)
```

registry/remote/repository.go
```diff
@@ -739,6 +796,58 @@ func (s *manifestStore) Resolve(ctx context.Context, reference string) (ocispec.
...
+		return nil, err
+	}
+	req.Header.Set("Accept", manifestAcceptHeader(s.repo.ManifestMediaTypes))
+
+	resp, err := s.repo.client().Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
```
shizhMSFT, line 821: same here

shizhMSFT (changes requested):

registry/remote/repository.go
Wwwsylvia: Oops!

registry/repository.go
Wwwsylvia: Good idea!

registry/remote/repository.go
```diff
@@ -739,6 +796,58 @@ func (s *manifestStore) Resolve(ctx context.Context, reference string) (ocispec.
 	}, nil
 }
 
+// FetchTag fetches the content identified by a reference.
+// It is equivalent to call `Resolve()` and then `Fetch()`
```
sajayantony, line 800: This should not be equivalent of resolve and fetch. They are 2 API requests where are this should be only one GET request.

sajayantony (changes requested):
Does this have both casss of GET by digest and by tag name?

registry/remote/repository.go
shizhMSFT: It is equivalent in terms of functionality. We need to mention that it is more efficient as it has only 1 GET request in the documentation.

sajayantony: I would agree but would like to keep the docs for the function without ambiguity. It is clearly 2 APIs and they have different behaviors depending on intermediaries.  

Maybe edit to keep it concise and have more details in the example. 
`//FetchTag fetches the content identified by the reference. The reference may be a tag or digest.`

sajayantony: On now thinking more - its sounds incorrect to call this method `FetchTag` if we are passing a digest. 
We should consider some other name like `FetchByReference(string reference....)`. 
Or am I missing something here?

Wwwsylvia: Agreed. I've rename it to `FetchReference` and made the doc concise.

sajayantony:
The renames look really good. The fact that we are standardizing on ref seems much more consistent.  Do you have a test case for this format 

`docker image pull k8s.gcr.io/pause-amd64:some-made-up-tag@sha256:59eec8837a4d942cc19a52b8c09ea75121acc38114a2c68b98983ce9356b8610`

Wwwsylvia:
> The renames look really good. The fact that we are standardizing on ref seems much more consistent. Do you have a test case for this format
> 
> `docker image pull k8s.gcr.io/pause-amd64:some-made-up-tag@sha256:59eec8837a4d942cc19a52b8c09ea75121acc38114a2c68b98983ce9356b8610`

Test cases added.

Wwwsylvia:
@shizhMSFT @sajayantony Could you please review the updates?

registry/remote/repository.go
```diff
@@ -583,17 +595,72 @@ func (s *blobStore) Resolve(ctx context.Context, reference string) (ocispec.Desc
 	default:
 		return ocispec.Descriptor{}, errutil.ParseErrorResponse(resp)
 	}
+
+	return s.generateDescriptor(resp, refDigest)
```
shizhMSFT, line 599: nit: let put this line to line 592

registry/remote/repository.go
```diff
@@ -583,17 +595,72 @@ func (s *blobStore) Resolve(ctx context.Context, reference string) (ocispec.Desc
...
+	default:
+		return ocispec.Descriptor{}, nil, errutil.ParseErrorResponse(resp)
+	}
+
+	desc, err = s.generateDescriptor(resp, refDigest)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, err
+	}
+
+	return desc, resp.Body, nil
```
shizhMSFT, line 645: After checking RFC 7233, it is possible to have `httputil.ReadSeekCloser` for this method, where the request should have a range header `Range: bytes=0-`.

Wwwsylvia: Nice

shizhMSFT (approved):
LGTM

sajayantony (approved):

sajayantony:
@qweeah will you add an example for this one?

PR #130, opened 2022-04-18
Extended Copy

### LINKED ISSUE #53: Extended Copy API
Besides #52, we also need to implemented the [Extended Copy](https://oras.land/client_libraries/#extended-copy) API.

Wwwsylvia:
Closes #53 

### DIFF
--- a/content/traceablestorage.go
+++ b/content/traceablestorage.go
@@ -13,18 +13,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package oras
+package content
 
-import "oras.land/oras-go/v2/content"
-
-// Target is a CAS with generic tags.
-type Target interface {
-	content.Storage
-	content.TagResolver
-}
-
-// TraceableTarget is a traceable-CAS with generic tags.
-type TraceableTarget interface {
-	content.TraceableStorage
-	content.TagResolver
+// TraceableStorage represents a traceable-CAS that supports parent node finding.
+type TraceableStorage interface {
+	Storage
+	UpEdgeFinder
 }
--- a/extendedcopy.go
+++ b/extendedcopy.go
@@ -1,0 +1,123 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package oras
+
+import (
+	"context"
+	"errors"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/internal/descriptor"
+)
+
+// ExtendedCopy copies the directed acyclic graph (DAG) that are reachable from
+// the given tagged node from the source TraceableTarget to the destination Target.
+// The destination reference will be the same as the source reference if the
+// destination reference is left blank.
+// Returns the descriptor of the tagged node on successful copy.
+func ExtendedCopy(ctx context.Context, src TraceableTarget, srcRef string, dst Target, dstRef string) (ocispec.Descriptor, error) {
+	if src == nil {
+		return ocispec.Descriptor{}, errors.New("nil source traceable target")
+	}
+	if dst == nil {
+		return ocispec.Descriptor{}, errors.New("nil destination target")
+	}
+	if dstRef == "" {
+		dstRef = srcRef
+	}
+
+	node, err := src.Resolve(ctx, srcRef)
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+
+	if err := ExtendedCopyGraph(ctx, src, dst, node); err != nil {
+		return ocispec.Descriptor{}, err
+	}
+
+	if err := dst.Tag(ctx, node, dstRef); err != nil {
+		return ocispec.Descriptor{}, err
+	}
+
+	return node, nil
+}
+
+// ExtendedCopyGraph copies the directed acyclic graph (DAG) that are reachable
+// from the given node from the source traceable-CAS to the destination CAS.
+func ExtendedCopyGraph(ctx context.Context, src content.TraceableStorage, dst content.Storage, node ocispec.Descriptor) error {
+	exists, err := src.Exists(ctx, node)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		return errdef.ErrNotFound
+	}
+
+	// find the root nodes traceable from the current node
+	roots := make(map[descriptor.Descriptor]ocispec.Descriptor)
+	if err := traceRoots(ctx, src, roots, node); err != nil {
+		return err
+	}
+
+	// copy the sub-DAGs rooted by the root nodes
+	for _, root := range roots {
+		if err := CopyGraph(ctx, src, dst, root); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// traceRoots traces the root nodes from the given node,
+// and records them in the given map.
+func traceRoots(ctx context.Context, storage content.TraceableStorage, roots map[descriptor.Descriptor]ocispec.Descriptor, node ocispec.Descriptor) error {
+	exists, err := storage.Exists(ctx, node)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


Wwwsylvia:
Hi @shizhMSFT can you help to review this?

extendedcopy.go
```diff
@@ -0,0 +1,123 @@
...
+// and records them in the given map.
+func traceRoots(ctx context.Context, storage content.TraceableStorage, roots map[descriptor.Descriptor]ocispec.Descriptor, node ocispec.Descriptor) error {
+	exists, err := storage.Exists(ctx, node)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		return errdef.ErrNotFound
+	}
```
shizhMSFT, line 98: Do we need to check the existence of the node? Shouldn't the `UpEdges()` do the check?

Could you also double check the oras artifact spec? I remember that we can have an artifact reference manifest where the subject does not exist.

extendedcopy.go
```diff
@@ -0,0 +1,123 @@
...
+		key := descriptor.FromOCI(node)
+		if _, exists := roots[key]; !exists {
+			roots[key] = node
+		}
+		return nil
+	}
+
+	// The current node has parents nodes. Keep tracing from the parent nodes.
+	for _, upEdge := range upEdges {
+		if err := traceRoots(ctx, storage, roots, upEdge); err != nil {
```
shizhMSFT, line 117: It is a recursive implementation. Could we make it iterative? Meanwhile, can we optimize the algorithm to address something like

```
  A   H
 / \ /
v   v
B   C
 \ /
  v
  D
 / \
v   v
E   F
 \ /
  v
  G
```
Given `G`, the algorithm should output `A` and `H` while other nodes should be only traversed once?

content/traceablestorage.go
```diff
@@ -0,0 +1,22 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package content
+
+// TraceableStorage represents a traceable-CAS that supports parent node finding.
+type TraceableStorage interface {
```
shizhMSFT, line 19: We probably need a better name other than `TraceableXXX`. Since we have `ExtendedCopy`, how about `ExtendedStorage`?

extendedcopy.go
```diff
@@ -0,0 +1,123 @@
...
+	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/internal/descriptor"
+)
+
+// ExtendedCopy copies the directed acyclic graph (DAG) that are reachable from
+// the given tagged node from the source TraceableTarget to the destination Target.
+// The destination reference will be the same as the source reference if the
+// destination reference is left blank.
+// Returns the descriptor of the tagged node on successful copy.
+func ExtendedCopy(ctx context.Context, src TraceableTarget, srcRef string, dst Target, dstRef string) (ocispec.Descriptor, error) {
```
shizhMSFT, line 33: Just curious. What if we pass a generic `Target` as `src`? It will do normal `Copy()` if target is not a `UpEdgeFinder`, and do `ExtendedCopy()` if it is, isn't it?

shizhMSFT: If so, we don't even need to worry about name like `TraceableTarget` and `TraceableStorage`.

content/traceablestorage.go
Wwwsylvia: It's an option. But users might be confused when looking at the name `ExtendedStorage` -- extend what? 🤔
How about `SourceableStorage`?

extendedcopy.go
Wwwsylvia: If we pass a `Target` which is not a `UpEdgeFinder` to `ExtendedCopy`, should we throw an error saying extended copy is not supported for this `Target`?

content/traceablestorage.go
shizhMSFT: Here the `Storage` stands for CAS. It only cares about raw content regardless of what kind of data structure it is storing. In terms of the definition, it is better to be named as `GraphStorage`, isn't it?

extendedcopy.go
shizhMSFT: In that case, we can fall back to regular `Copy()`.

Wwwsylvia: I think falling back to regular `Copy()` is confusing, as users expect it to do extended copy. If it can't do, it should let users know.

extendedcopy.go
Wwwsylvia: The `subject` is an optional field, but when specified, the specified manifest must exist in the repository.
image

extendedcopy.go
shizhMSFT: I am thinking about a scenario that a user is copying an artifact from one target to another. That user wants the artifact together with the associated signatures. The user has to do reflection on the Target first to see if it supports ExtendedCopy or not. Precisely,

```go
var desc ocispec.Descriptor
var err error
if target, ok := target.(TraceableTarget); ok {
    desc, err = oras.ExtendedCopy(...)
} else {
    desc, err = oras.Copy(...)
}
if err != nil {
    // handle error
}
```

From UX perspective, it would be better if `ExtendCopy()` is interpreted as _copying the artifact as well as its associated items if present_. A regular `Target`, which does not support `UpEdgeFinder`, can be viewed as a `TraceableTarget` that always return no up edges on any nodes.

extendedcopy.go
shizhMSFT: Have you tried to upload an artifact manifest to a registry, and then delete the subject manifest, and then try accessing the referral API?

shizhMSFT: The [Manifest Referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#api-discovery) specifies that `UpEdges()` returns it parents even if the node is deleted.

extendedcopy.go
Wwwsylvia: If the user wants to copy the artifact together with the associated signatures, they should use `ExtendedCopy()`.
If they does not care about the signature, they just uses `Copy()`.
The user can decide which one to use by themself.

content/traceablestorage.go
Wwwsylvia: `GraphStorage` sounds good!

extendedcopy.go
Wwwsylvia: Interesting behavior....

extendedcopy.go
Wwwsylvia: Fixed.

extendedcopy.go
```diff
@@ -0,0 +1,123 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// traceRoots traces the root nodes from the given node,
+// and records them in the given map.
+func traceRoots(ctx context.Context, storage content.TraceableStorage, roots map[descriptor.Descriptor]ocispec.Descriptor, node ocispec.Descriptor) error {
```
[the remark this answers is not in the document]
Wwwsylvia: Let's go for `findRoots()`.

extendedcopy.go
Wwwsylvia: OK let's not check the node existence to reduce the requests.

extendedcopy.go
```diff
@@ -0,0 +1,123 @@
...
+	for _, root := range roots {
+		if err := CopyGraph(ctx, src, dst, root); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// findRoots finds the root nodes reachable from the given node.
```
shizhMSFT, line 77: The current implementation is done by depth-first search. It might be good to call it out in the documentation somewhere for readability.

extendedcopy.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// ExtendedCopy copies the directed acyclic graph (DAG) that are reachable from
+// the given tagged node from the source GraphTarget to the destination Target.
+// The destination reference will be the same as the source reference if the
+// destination reference is left blank.
+// Returns the descriptor of the tagged node on successful copy.
+func ExtendedCopy(ctx context.Context, src GraphTarget, srcRef string, dst Target, dstRef string) (ocispec.Descriptor, error) {
+	if src == nil {
+		return ocispec.Descriptor{}, errors.New("nil source traceable target")
```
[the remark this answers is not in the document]
Wwwsylvia: Good catch!

extendedcopy.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return err
+	}
+
+	// copy the sub-DAGs rooted by the root nodes
+	for _, root := range roots {
+		if err := CopyGraph(ctx, src, dst, root); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
Wwwsylvia: Will link an issue once this PR get merged.

Issue #141, opened 2022-05-10
ProviderWrapper can't read intact desc when the size is large


raygecao:
When I use ProviderWrapper to get descriptor，an error is shown as follows:
```
children: invalid desc sha256:47ac59fec96c3e65abc21dc579dfbc0eee7dd2b4ae0cc687fe273c78e9ee81d5: invalid character '\x00' in string literal
```
The error is thrown during `json.Unmarshal`，because the  manifest is partial read. 

After some debug, the bug is direct read  shown below causing partial read:
https://github.com/oras-project/oras-go/blob/52eccf5764eb9994d88d1aa69025da04d6963187/pkg/oras/provider.go#L74

An example for partial read:
```
func TestGet(t *testing.T) {
	req, err := http.NewRequest("GET", "http://myregistry:5000/v2/fluent-bit/manifests/sha256:47ac59fec96c3e65abc21dc579dfbc0eee7dd2b4ae0cc687fe273c78e9ee81d5", nil)
	if err != nil {
		panic(err)
	}
	c := &http.Client{}
	resp, _ := c.Do(req)
	assert.NoError(t, err)
	defer resp.Body.Close()
	b := make([]byte, 3000)
	n, err := resp.Body.Read(b)
	fmt.Printf("read %d bytes, err is: %v\n", n, err)
	content, _ := ioutil.ReadAll(resp.Body)
	fmt.Printf("leave %d byte unread after first read", len(content))
}
```

The output is:
```
read 3000 bytes, err is: <nil>
leave 4665 byte unread after first read
```

shizhMSFT:
Closing as resolved by #142

PR #142, opened 2022-05-10
use io.ReadFull for ProviderWrapper in case of partial read

raygecao:
fixes https://github.com/oras-project/oras-go/issues/141


shizhMSFT:
Thanks @raygecao for fixing the issue. The issue is for v1 implementation of `oras-go`. Could you change your target branch to `v1` and fix DCO?

shizhMSFT (approved):
LGTM

Issue #129, opened 2022-04-18
Failed to copy manifest with 0 size config


northtyphoon:
I tried to copy a manifest from registry `binducentralus.azurecr.io/samples/artifact:1.0`. This manifest has a config whose size is 0. 

```
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.unknown.config.v1+json",
    "digest": "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "size": 0
  },
  "layers": [
    {
      "mediaType": "application/vnd.unknown.layer.v1+txt",
      "digest": "sha256:6291bbe45f529c52451285872d5fce81296aafac2190634d6f2e17c08c2e3dcb",
      "size": 44,
      "annotations": {
        "org.opencontainers.image.title": "artifact.txt"
      }
    }
  ]
}
```

The following code failed as 400 bad request. 

```
src, err := remote.NewRepository("binducentralus.azurecr.io/samples/artifact:1.0")
dest := file.New("temp")
desc, err := oras.Copy(ctx, src, src.Reference.Reference, dest, "")
```

The problem was the invalid range header value (`bytes=0--1`) in the download request due to https://github.com/oras-project/oras-go/blob/c34895e747dcd988adb3cc959302b5e94117195b/registry/remote/repository.go#L447.

wangxiaoxuan273:
Please assign this issue to me, thanks

shizhMSFT:
Resolved by #145

PR #145, opened 2022-05-13
added basic range header handling

wangxiaoxuan273:
#129 Added an if check before setting a range header limit value. Added one test that performs range header check when a range header is present.

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -444,7 +444,9 @@ func (s *blobStore) Fetch(ctx context.Context, target ocispec.Descriptor) (rc io
 	// Docker spec allows range header form of "Range: bytes=<start>-<end>".
 	// However, the remote server may still not RFC 7233 compliant.
 	// Reference: https://docs.docker.com/registry/spec/api/#blob
-	req.Header.Set("Range", fmt.Sprintf("bytes=0-%d", target.Size-1))
+	if target.Size > 0 {
+		req.Header.Set("Range", fmt.Sprintf("bytes=0-%d", target.Size-1))
+	}
 
 	resp, err := s.repo.client().Do(req)
 	if err != nil {
--- a/registry/remote/repository_test.go
+++ b/registry/remote/repository_test.go
@@ -1207,6 +1207,63 @@ func Test_BlobStore_Fetch_Seek(t *testing.T) {
 	}
 }
 
+func Test_BlobStore_Fetch_ZeroSizedBlob(t *testing.T) {
+	blob := []byte("")
+	blobDesc := ocispec.Descriptor{
+		MediaType: "test",
+		Digest:    digest.FromBytes(blob),
+		Size:      int64(len(blob)),
+	}
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			t.Errorf("unexpected access: %s %s", r.Method, r.URL)
+			w.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
+
+		switch r.URL.Path {
+		case "/v2/test/blobs/" + blobDesc.Digest.String():
+			if rangeHeader := r.Header.Get("Range"); rangeHeader != "" {
+				t.Errorf("unexpected range header")
+				w.WriteHeader(http.StatusBadRequest)
+				return
+			}
+			w.Header().Set("Content-Type", "application/octet-stream")
+			w.Header().Set("Docker-Content-Digest", blobDesc.Digest.String())
+		default:
+			w.WriteHeader(http.StatusNotFound)
+		}
+	}))
+	defer ts.Close()
+	uri, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatalf("invalid test http server: %v", err)
+	}
+
+	repo, err := NewRepository(uri.Host + "/test")
+	if err != nil {
+		t.Fatalf("NewRepository() error = %v", err)
+	}
+	repo.PlainHTTP = true
+	store := repo.Blobs()
+	ctx := context.Background()
+
+	rc, err := store.Fetch(ctx, blobDesc)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


registry/remote/repository_test.go
```diff
@@ -2146,3 +2146,63 @@ func Test_ManifestStore_FetchReference(t *testing.T) {
 		t.Errorf("Manifests.FetchReference() = %v, want %v", got, manifest)
 	}
 }
+
+func Test_ZeroSizedBlob_Fetch(t *testing.T) {
```
shizhMSFT, line 2150: Could we have consistent test names in a pattern of
```
Test_<TypeName>_<FunctionName>_<Scenario>
```
?

For instance, `Test_BlobStore_Fetch_ZeroSizedBlob`?

shizhMSFT: You might also want to put the test asides other `blobStore` tests.

registry/remote/repository_test.go
```diff
@@ -2146,3 +2146,63 @@ func Test_ManifestStore_FetchReference(t *testing.T) {
...
+		}
+		rangeHeader := r.Header.Get("Range")
+		if rangeHeader != "" {
+			rangeLimit, err := strconv.Atoi(rangeHeader[8:])
+			if err == nil && rangeLimit < 0 {
+				t.Errorf("invalid range limit: 0-%d", rangeLimit)
+				w.WriteHeader(http.StatusBadRequest)
+				return
+			}
+		}
```
shizhMSFT, line 2171: Since we fail as long as the range header presents, we can simplify the mock as

```go
if rangeHeader != "" {
    t.Errorf("unexpected range header: %s", rangeHeader )
    w.WriteHeader(http.StatusBadRequest)
    return
}
```

wangxiaoxuan273: It's not the case that we fail as long as the range header presents. We only fail when the range header presents, and it's of the form of "bytes=0-(a negative number)". When the header is of the form "bytes=0-" (acceptable according to line 626) and "bytes=0-(a positive number)", we don't fail.

registry/remote/repository.go
```diff
@@ -444,7 +444,9 @@ func (s *blobStore) Fetch(ctx context.Context, target ocispec.Descriptor) (rc io
 	// Docker spec allows range header form of "Range: bytes=<start>-<end>".
 	// However, the remote server may still not RFC 7233 compliant.
 	// Reference: https://docs.docker.com/registry/spec/api/#blob
-	req.Header.Set("Range", fmt.Sprintf("bytes=0-%d", target.Size-1))
+	if target.Size != 0 {
```
Wwwsylvia, line 447: What if `target.Size` is -1?

registry/remote/repository_test.go
shizhMSFT: First of all, the range header check should happen only if the blob exists. In other words, this block of code should in the switch case between lines 2173 and 2174.
Next, as long as the blob to be fetched has zero length, the server returns an error if the range header presents. The error is either 400 or 416. Nonetheless, the unit test fails anyway for 400 or 416.

registry/remote/repository_test.go
wangxiaoxuan273: Changed the test name to "Test_BlobStore_Fetch_ZeroSizedBlob". Only the blobStore Fetch function has this issue, so only blobStore test is needed. The manifest Fetch function doesn't set Range headers.

registry/remote/repository_test.go
wangxiaoxuan273: Moved the range header check inside the switch. Differentiated the cases of 400 and 416. The server now returns 400 when the blob has 0 length and a range header is present.

registry/remote/repository.go
wangxiaoxuan273: Changed this line to if target.Size > 0. Thanks for the feedback.

registry/remote/repository_test.go
```diff
@@ -1207,6 +1207,71 @@ func Test_BlobStore_Fetch_Seek(t *testing.T) {
...
+		switch r.URL.Path {
+		case "/v2/test/blobs/" + blobDesc.Digest.String():
+			if rangeHeader := r.Header.Get("Range"); rangeHeader != "" {
+				if blobDesc.Size <= 0 {
+					t.Errorf("unexpected range header")
+					w.WriteHeader(http.StatusBadRequest)
+					return
+				}
+				rangeLimit, err := strconv.ParseInt((rangeHeader[8:]), 10, 64)
+				if err == nil && rangeLimit != blobDesc.Size-1 {
```
shizhMSFT, line 1233: What if `err != nil`? Should we return `400` in this case?

shizhMSFT: Also, the `blobDesc.Size` is always `0` since `blob := []byte("")`. Thus, `rangeLimit != blobDesc.Size-1` translates to `rangeLimit != -1` and is always `true`.

wangxiaoxuan273: I kept `err == nil` for the (unlikely) case of "Range: bytes=0-", which is acceptable according to line 623: "The form of "Range: bytes=<start>-" is also acceptable."

I don't know if I need to further differentiate the cases where `err != nil`.

wangxiaoxuan273: When blobDesc.Size = 0, the first if statement returns. For whatever reaches the rangeLimit part, the size is >0.

shizhMSFT: > When blobDesc.Size = 0, the first if statement returns. For whatever reaches the rangeLimit part, the size is >0.

Nice call out. As I've mentioned above `blobDesc.Size` is always `0`, thus `blobDesc.Size <= 0` is always true. Therefore, the lines 1232 - 1236 are never executed, and thus can be removed.

wangxiaoxuan273: Removed these lines.

shizhMSFT (approved):
LGTM

PR #150, opened 2022-05-23
Copy Options

Wwwsylvia:
Support advanced options for Copy and ExtendedCopy.


copyoptions.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"context"
+	"math"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/semaphore"
+)
+
+const defaultConcurrencyLimit = 3
```
Wwwsylvia, line 26: Leart this from [containerd](https://sourcegraph.com/github.com/containerd/containerd@eb39aac/-/blob/pkg/cri/config/config_unix.go?L101) and [moby](https://sourcegraph.com/github.com/moby/moby@8abb84c/-/blob/daemon/config/config.go?L26)

copyoptions.go
```diff
@@ -0,0 +1,86 @@
...
+
+func NewExtendedCopyOptions() *extendedCopyOptions {
+	return &extendedCopyOptions{
+		copyOptions: NewCopyOptions(),
+		depth:       math.MaxUint32,
+	}
+}
+
+func (eco *extendedCopyOptions) WithConcurrencyLimit(concurrency int64) *extendedCopyOptions {
+	eco.limiter = semaphore.NewWeighted(concurrency)
```
qweeah, line 64: how about `eco.copyOptions.WithConcurrencyLimit(concurrency)`

copyoptions.go
```diff
@@ -0,0 +1,86 @@
...
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/semaphore"
+)
+
+const defaultConcurrencyLimit = 3
+
+type copyOptions struct {
+	limiter             *semaphore.Weighted
+	copyCallbackHandler func(ctx context.Context, desc ocispec.Descriptor) error
```
qweeah, line 30: abount naming: this is more like a `copyDoneHandler`.

copyoptions.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (co *copyOptions) WithCopyCallbackHandler(fn func(ctx context.Context, desc ocispec.Descriptor) error) *copyOptions {
+	co.copyCallbackHandler = fn
+	return co
+}
+
+type extendedCopyOptions struct {
+	*copyOptions
+	depth               uint32
+	upEdgeFilterHandler func(ctx context.Context, desc ocispec.Descriptor) (bool, error)
```
qweeah, line 52: naming: this is more like a `upEdgeFilter`

copyoptions.go
```diff
@@ -0,0 +1,86 @@
...
+	customUpEdgesFinder func(ctx context.Context, desc ocispec.Descriptor) ([]ocispec.Descriptor, error)
```
qweeah, line 53: naming: this is more like a `upEdgeFinder`

copyoptions.go
qweeah: Also can we add another handler called we a node being found existed during copying (see [here](https://github.com/oras-project/oras-go/blob/main/copy.go#L86-L88))?

It will be a compliment to this copy call back function, client can handle 
- copied desc through `copyDoneHandler`
- skipped desc through `copySkippedHandler`


Feel free to resolve directly if it's out of this PR's scope.

copyoptions.go
Wwwsylvia: A default `upEdgesFinder` will be used if this is not provided. So this is more like a "custom" one.

copyoptions.go
Wwwsylvia: What would be the E2E UX for oras CLI? Do we want something like below:
```
$ oras cp $src $dst 
sha256:abc123: Completed
sha256:def456: Already Exists
```

qweeah: Yes, and the `copySkippedHandler` can be used for the `Already Exists` output

Wwwsylvia: How about other scenarios like blob not found error?

copyoptions.go
Wwwsylvia: `copyOptions` is embeded in `extendedCopyOptions`. So `eco.copyOptions.WithConcurrencyLimit(concurrency)` is equivalent to `eco.WithConcurrencyLimit(concurrency)`.

copyoptions.go
shizhMSFT: We can have a design meeting for this scenario.

copyoptions.go
shizhMSFT: I think @qweeah is referring

```go
func (eco *extendedCopyOptions) WithConcurrencyLimit(concurrency int64) *extendedCopyOptions {
    eco.copyOptions.WithConcurrencyLimit(concurrency)
    return eco
}
```

copy.go
```diff
@@ -33,7 +33,7 @@ import (
 // The destination reference will be the same as the source reference if the
 // destination reference is left blank.
 // Returns the descriptor of the root node on successful copy.
-func Copy(ctx context.Context, src Target, srcRef string, dst Target, dstRef string) (ocispec.Descriptor, error) {
+func Copy(ctx context.Context, src Target, srcRef string, dst Target, dstRef string, copyOpts *copyOptions) (ocispec.Descriptor, error) {
```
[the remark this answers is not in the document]
shizhMSFT: Or make it an interface.

copyoptions.go
shizhMSFT: I also noticed the drawback of this pattern. `eco` is not able to reuse the `WithConcurrencyLimit` function of `copyOptions`, and thus we have redundant codes. 

Thus, I suggest dropping the fluent-style interface. Instead, we can have

```go
func (co *copyOptions) LimitConcurrency(concurrency int64)
```

Another drawback for the fluent-style interface is the error handling. What if you have

```go
func (eco *extendedCopyOptions) WithMagic(magic any) (*extendedCopyOptions, error)
```

Then we are not able to chain the options.

copyoptions.go
```diff
@@ -0,0 +1,86 @@
...
+	eco.limiter = semaphore.NewWeighted(concurrency)
+	return eco
+}
+
+func (eco *extendedCopyOptions) WithCopyCallbackHandler(fn func(ctx context.Context, desc ocispec.Descriptor) error) *extendedCopyOptions {
+	eco.copyCallbackHandler = fn
+	return eco
+}
+
+func (eco *extendedCopyOptions) WithDepth(depth uint32) *extendedCopyOptions {
```
shizhMSFT, line 73: Curious why `depth` is just `int`? We don't expect we have a huge DAG.

shizhMSFT: where negative depth means infinity? or why not `uint64`?

extendedcopy.go
```diff
@@ -76,49 +80,110 @@ func ExtendedCopyGraph(ctx context.Context, src content.GraphStorage, dst conten
...
 }
+
+func getUpEdges(ctx context.Context, finder content.GraphStorage, node ocispec.Descriptor, opts *extendedCopyOptions) ([]ocispec.Descriptor, error) {
+	var upEdges []ocispec.Descriptor
+	var err error
+	if opts.customUpEdgesFinder != nil {
+		upEdges, err = opts.customUpEdgesFinder(ctx, node)
+	} else {
+		upEdges, err = finder.UpEdges(ctx, node)
+	}
```
shizhMSFT, line 144: What if we set `opts.customUpEdgesFinder = finder.UpEdges` in the first place?

extendedcopy.go
```diff
@@ -76,49 +80,110 @@ func ExtendedCopyGraph(ctx context.Context, src content.GraphStorage, dst conten
...
+func (s *stack) push(i item) {
+	s.items = append(s.items, i)
+}
+
+func (s *stack) pop() item {
+	last := len(s.items) - 1
+	curr := s.items[last]
+	s.items = s.items[:last]
+	return curr
+}
```
shizhMSFT, line 189: Is it possible to refactor this?

copy.go
```diff
@@ -62,10 +62,13 @@ func Copy(ctx context.Context, src Target, srcRef string, dst Target, dstRef str
 
 // CopyGraph copies a rooted directed acyclic graph (DAG) from the source CAS to
 // the destination CAS.
-func CopyGraph(ctx context.Context, src, dst content.Storage, root ocispec.Descriptor) error {
+func CopyGraph(ctx context.Context, src, dst content.Storage, root ocispec.Descriptor, opts *copyOptions) error {
```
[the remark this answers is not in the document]
Wwwsylvia: I'm wondering what options apply only to `Copy` and not to `CopyGraph`?

copyoptions.go
qweeah: Adding a E2E UX for the design discussion when using oras pull command to download a manifest config:

Suppose the target config don't have a title name stored in the annotation or the stored title doesn't match the expected download file name, the oras cli client need to 1) rewrite the title annonation 2) call oras-go for copying 3) print out the config name and digest

`copyDoneHandler` won't satify the need since it only covers step 1), we should also provide another handler and `beforeCopyHandler` and maybe as a step further, pack those two in one copy option property.

copy.go
```diff
@@ -128,11 +157,34 @@ func CopyGraph(ctx context.Context, src, dst content.Storage, root ocispec.Descr
...
+	}
+
+	if err := copyNode(ctx, src, dst, desc); err != nil {
+		return err
+	}
+
+	if opts.PostCopyHandler != nil {
+		if err := opts.PostCopyHandler(ctx, desc); err != nil {
+			return err
+		}
```
qweeah, line 185:
```suggestion
		return opts.PostCopyHandler(ctx, desc)
```

copy.go
```diff
@@ -49,7 +64,16 @@ func Copy(ctx context.Context, src Target, srcRef string, dst Target, dstRef str
 		return ocispec.Descriptor{}, err
 	}
 
-	if err := CopyGraph(ctx, src, dst, root); err != nil {
+	// TODO: manifest filter? 1. media type filter 2. platform filter
```
qweeah, line 67: The platform type filtering depends on the client, oras-go only need to provide the manifestFilter options.

For the mediatype, manifestFilter can do the manifest filtering but we still need downEdgeFilter support.

extendedcopy.go
```diff
@@ -45,7 +58,16 @@ func ExtendedCopy(ctx context.Context, src GraphTarget, srcRef string, dst Targe
 		return ocispec.Descriptor{}, err
 	}
 
-	if err := ExtendedCopyGraph(ctx, src, dst, node); err != nil {
+	// TODO: manifest filter? 1. media type filter 2. platform filter
```
qweeah, line 61: The platform type filtering depends on the client, oras-go only need to provide the manifestFilter options.

For the mediatype, manifestFilter can do the manifest filtering but we still need downEdgeFilter support.

copy.go
Wwwsylvia: good catch!

extendedcopy.go
```diff
@@ -76,49 +80,110 @@ func ExtendedCopyGraph(ctx context.Context, src content.GraphStorage, dst conten
...
+}
+
+type item struct {
+	node  ocispec.Descriptor
+	depth uint32
+}
+
+type stack struct {
+	items []item
+}
```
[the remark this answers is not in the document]
Wwwsylvia: `*stack` It's easier to use

extendedcopy.go
```diff
@@ -76,49 +80,110 @@ func ExtendedCopyGraph(ctx context.Context, src content.GraphStorage, dst conten
...
+	}
 
-	// push the initial node to the stack
-	stack = append(stack, node)
-	for len(stack) > 0 {
-		// pop the current node from the stack
-		top := len(stack) - 1
-		current := stack[top]
-		stack = stack[:top]
+	stack := &stack{}
```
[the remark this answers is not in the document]
Wwwsylvia: `*stack` is easier to use

copy.go
Wwwsylvia: Yeah this was a temporary note to remind myself.

copy.go
```diff
@@ -21,19 +21,55 @@ import (
...
+// The value is consistent with dockerd and containerd.
+const defaultConcurrency = int64(3)
+
+// CopyOptions contains parameters for oras.Copy.
+type CopyOptions struct {
+	CopyGraphOptions
+	// ManifestFilter filters the manifest that is to be used as the root node
+	// for copy. Returns true if the manifest matches the filter, otherwise
+	// returns false.
+	ManifestFilter func(ctx context.Context, manifest ocispec.Descriptor) (bool, error)
```
Wwwsylvia, line 44: I'm re-considering this. We may still need two filter, one for the resolved manifest itself, another one for its down edges.

copy_test.go
```diff
@@ -352,3 +354,306 @@ func TestCopyGraph_PartialCopy(t *testing.T) {
...
+	root := descs[6]
+	ref := "foobar"
+	err := src.Tag(ctx, root, ref)
+	if err != nil {
+		t.Fatal("fail to tag root node", err)
+	}
+
+	// test copy with media type filter
+	opts := oras.CopyOptions{
+		RootFilter: func(ctx context.Context, root ocispec.Descriptor) (ocispec.Descriptor, bool, error) {
```
qweeah, line 432: Can you help explain the scenarios we need to use this root filter?

copy.go
```diff
@@ -21,19 +21,50 @@ import (
...
+// defaultConcurrency is the default concurrency limit.
+// The value is consistent with dockerd and containerd.
+const defaultConcurrency = int64(3)
+
+// CopyOptions contains parameters for oras.Copy.
+type CopyOptions struct {
+	CopyGraphOptions
+	// RootFilter filters the descriptor that is to be used as the root node for
+	// copy.
+	RootFilter func(ctx context.Context, root ocispec.Descriptor) (desc ocispec.Descriptor, found bool, err error)
```
qweeah, line 41: Can you help explain the scenarios we need to use this root filter?

Wwwsylvia: Some scenarios I can think of are:
1. Filter manifests of allowed media types for copy. e.g. Clients may want to copy only when the resolved manifest is a OCI manifest instead of a Docker manifest, like https://github.com/Wwwsylvia/oras-go/blob/copy_options/copy_test.go#L430-L439.
2. Filter manifests of allowed platforms for copy. e.g. Clients may want to pick a manifest of a specific platform from a resolved manifest list, like https://github.com/Wwwsylvia/oras-go/blob/copy_options/copy_test.go#L468-L489.

When there are multiple matching manifests, clients can decide which manifest to return. Or they can generate a new manifest list and return the descriptor of the manifest list. Maybe we can add some utility functions in the future to support filtering and generating manifest list.

Does it make sense?

qweeah: Sorted out offline with Sylvia. The second use case is the most common one(like `--platform` flag for crane and ctr image pull) and the current RootFilter implementation does cover it.

extendedcopy.go
```diff
@@ -21,15 +21,37 @@ import (
...
+}
+
+// ExtendedCopyGraphOptions contains parameters for oras.ExtendedCopyGraph.
+type ExtendedCopyGraphOptions struct {
+	CopyGraphOptions
+	// Depth limits the maximum depth of the directed acyclic graph (DAG) that
+	// will be extended-copied.
+	// If Depth is no specified, or the specified value is less or equal than 0,
+	// the depth limit will be considered as infinity.
+	Depth int
```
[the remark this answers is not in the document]
Wwwsylvia: Using the example in [artifact-spec](https://github.com/oras-project/artifacts-spec#overview), specifing `Depth = 1` allows us to copy `net-monitor:v1` along with `wabbit-networks signature` but without the `wabbit-networks SBOM signature`.
image

copy.go
```diff
@@ -21,19 +21,50 @@ import (
...
 	"oras.land/oras-go/v2/internal/status"
 )
 
+// defaultConcurrency is the default concurrency limit.
+// The value is consistent with dockerd and containerd.
+const defaultConcurrency = int64(3)
+
+// CopyOptions contains parameters for oras.Copy.
+type CopyOptions struct {
+	CopyGraphOptions
```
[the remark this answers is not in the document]
Wwwsylvia: `ReferenceFetcher` and `ReferencePusher` can only be applied to the root node. Maybe we can have some options for clients to use them in `Copy()` and do something like:
1. Fetch the root by reference
2. Call `CopyGraph()` on the down edges of the root
3. Push the root by reference

So that we can save 1 `Resolve()` call and 1 `Tag()` call.

registry/repository.go
```diff
@@ -59,6 +58,11 @@ type Repository interface {
 	// - https://docs.docker.com/registry/spec/api/#tags
 	// See also `Tags()` in this package.
 	Tags(ctx context.Context, fn func(tags []string) error) error
+
+	// Referrers returns the manifest descriptors directly referencing the given
+	// manifest descriptor.
+	// Reference: https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md
+	Referrers(ctx context.Context, desc ocispec.Descriptor, fn func(referrers []artifactspec.Descriptor) error) error
```
shizhMSFT, line 65: We probably don't want to add `Referrers` here because not all registries support `Referrers` API.

Wwwsylvia: Moved this out.

copy.go
```diff
@@ -19,21 +19,62 @@ import (
...
+	OnCopySkipped func(ctx context.Context, desc ocispec.Descriptor) error
+}
+
 // Copy copies a rooted directed acyclic graph (DAG) with the tagged root node
 // in the source Target to the destination Target.
 // The destination reference will be the same as the source reference if the
 // destination reference is left blank.
+// When MapRoot option is provided, the descriptor resolved from the source
+// reference will be passed to the MapRoot, and the mapped descriptor will
+// be used as the root node for copy.
```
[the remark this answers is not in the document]
Wwwsylvia: make sense

copy.go
```diff
@@ -148,3 +214,126 @@ func copyNode(ctx context.Context, src, dst content.Storage, desc ocispec.Descri
...
+		//       and dst is a ReferencePusher.
+	} else {
+		root, err = src.Resolve(ctx, srcRef)
+	}
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+
+	return root, nil
+}
```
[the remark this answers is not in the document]
Wwwsylvia: nice catch!

shizhMSFT (approved):
LGTM

shizhMSFT:
We can merge this once the unit tests are added.

internal/registryutil/proxy.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	exists, err := p.Cache.Exists(ctx, target)
+	if err != nil {
+		return ocispec.Descriptor{}, nil, err
+	}
+	if exists {
+		// no need to cache
+		return target, rc, nil
+	}
```
shizhMSFT, line 69: The logic of this block of code is problematic. If should cache be used, `rc` should come from `p.Cache`.

Wwwsylvia: It's not using cache, just to avoid pushing to cache again.

shizhMSFT (approved):
LGTM

PR #151, opened 2022-05-24
Bump containerd to v1.6.4

lucacome:
Bumps containerd. 

I had to use `go mod tidy -compat=1.17` as go `1.16` would select different dependencies, so I think this makes it incompatible with go `1.16`. I think this is fine since `1.16` is not supported anymore.


shizhMSFT (approved):
LGTM

PR #153, opened 2022-05-24
Update to Go 1.18

lucacome:
Bumps go to 1.18

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module oras.land/oras-go/v2
 
-go 1.17
+go 1.18
 
 require (
 	github.com/opencontainers/go-digest v1.0.0


shizhMSFT:
Since this is library code, adopters may still use older golang versions. We need to be careful on bumping up go versions.

@jdolitsky @deitch @sajayantony @SteveLasker @feynmanzhou As a library, we should support a list of golang versions. For example, the last 2 - 3 versions.

shizhMSFT:
Here's the reference for [go.mod](https://go.dev/doc/modules/gomod-ref).

The go version in the `go.mod` is _the minimum version of Go required by the current module_. I'd like to also mention that the go module with minimum requirement of 1.17 can be compiled against 1.18 and the resulted binary is in 1.18.

shizhMSFT:
Linking #155

shizhMSFT:
@lucacome Could you open an issue to justify the reason to bump up the minimum golang version requirement of `oras-go`? This is required by our [reviewing guide](https://github.com/oras-project/oras-go/blob/main/REVIEWING.md).

lucacome:
> The go version in the `go.mod` is _the minimum version of Go required by the current module_. I'd like to also mention that the go module with minimum requirement of 1.17 can be compiled against 1.18 and the resulted binary is in 1.18.

It can also be the other way around, with the minimum of 1.18 it can be compiled against 1.17 if the library is not using anything specific to 1.18 (for example generics)

lucacome:
I guess this was done in https://github.com/oras-project/oras-go/pull/270 🤷‍♂️

PR #162, opened 2022-06-07
Make the proxy struct and related methods public 

### LINKED ISSUE #163: Can we make the proxy struct and related methods public?
When working on upgrading pull command for ORAS CLI, I need to create a OCI store to avoid fetching duplicated blobs again and again from the remote source. I found the internal [Proxy ](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go) struct in ORAS-go's internal cas package is just what I need. Can we make the proxy struct public so clients can reuse it to build light-weight read-through caches?

Pros:
ORAS-go users can reuse the code to build caches in their software.

Cons:
The current implementation is only a read-through cache with no consistency guaranteed. If any blob is deleted on the origin storage side, the change won't be reflected in the proxy.

qweeah:
This PR adds a new public package for read-through cache proxy.

Resolves #163

### DIFF
--- a/content/proxy/proxy.go
+++ b/content/proxy/proxy.go
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package cas
+package proxy
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


shizhMSFT:
Closing since the original issue is closed.

PR #167, opened 2022-06-07
Bump cobra to v1.4.0

lucacome:
Bumps cobra to v1.4.0 

Cobra removed Viper from their codebase, removing a lot of indirect dependencies https://github.com/spf13/cobra/releases/tag/v1.4.0


shizhMSFT (approved):
LGTM. This one is interesting as it removes lots of dependencies.

Issue #166, opened 2022-06-07
Release v1.1.2


lucacome:
Is it possible to release `v1.1.2`? It will contain a bug fix and dependencies updates. Thanks!

shizhMSFT:
@jdolitsky Could you help with this since it is v1?

PR #168, opened 2022-06-08
Bump dependencies

lucacome:
Grouping a bunch of updates together since they're all very small, let me know if that's okay. 

Most notable update is for `golang.org/x/crypto` that removes https://pkg.go.dev/vuln/GO-2021-0356

Hopefully this is in preparation for
-  #166  😉


shizhMSFT (approved):
LGTM

PR #174, opened 2022-06-09
Implement Delete in the oci store content implementation

viovanov:
Signed-off-by: viovanov <vlad@aserto.com>


viovanov:
Thanks @shizhMSFT!
Do you know when this will make it into a v1 release?

shizhMSFT:
> Thanks @shizhMSFT! Do you know when this will make it into a v1 release?

I think it will happen very soon. We just need a ceremony :D

viovanov:
anything we can do to help for the ceremony @shizhMSFT?
maybe send some cake over? :)

PR #179, opened 2022-06-16
Remove v1 implementation

### LINKED ISSUE #63: Remove old implementation

Wwwsylvia:
Resolves #63 


Wwwsylvia:
Also removing the `examples` directory and the acceptance check since they are using the v1 implementation. We may want to have a new acceptance test for v2.

shizhMSFT:
The acceptance test is there for `v1` since `v1` does not have sufficient unit tests to catch regressions.

With tons of unit tests and [testable examples](https://go.dev/blog/examples), we don't need to maintain acceptance tests any more.

shizhMSFT (approved):
LGTM

deitch:
A few issues or questions.

1. Should we be removing this without acceptance tests and examples? Maybe we should replace them in `main` before removing entirely? I saw we added a lot more testable examples. Are there enough to cover the existing and simple use cases? If so, then maybe we should link to it from the main page / README. Let's make it easy for people to get started.
2. We link using the go package semver structure, i.e. `oras.land/oras-go/v2`; does that "do the right thing" when including the root dir of this package?
3. Do we have a migrating log? Showing the differences between v1 and v2?

shizhMSFT:
> Are there enough to cover the existing and simple use cases?

Yes, I'll create an issue for improving the doc to direct the users to the examples for push and pull.

> does that "do the right thing" when including the root dir of this package?

That's actually how go module works. The module name for `v2.x.x` must ends with `/v2` (and similarly `/v3` for `v3.0.0`). Although the module name is `oras.land/oras-go/v2`, all source code SHOULD be placed at `oras.land/oras-go`.

BTW, users can import `oras.land/oras-go/v2` and `oras.land/oras-go` at the same time. Hence, users can migrate from `v1` to `v2` iteratively.

> Do we have a migrating log? Showing the differences between v1 and v2?

That's a good point. I will create an issue to track it.

shizhMSFT:
Created the following issues for tracking
- https://github.com/oras-project/oras-go/issues/180
- https://github.com/oras-project/oras-go/issues/181

deitch (approved):
Thanks for addressing those two.

PR #187, opened 2022-06-20
Implement Info in the oci store content implementation

oanatmaria:
Signed-off-by: oanatmaria oana@aserto.com

### DIFF
--- a/pkg/content/oci.go
+++ b/pkg/content/oci.go
@@ -251,11 +251,10 @@ func (s *OCI) validateOCILayoutFile() error {
 	return nil
 }
 
-// TODO: implement (needed to create a content.Store)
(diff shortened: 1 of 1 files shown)


viovanov:
Hi @shizhMSFT !
One more please? :)

FeynmanZhou:
> 

Hi @viovanov ,

I think we can prepare a new release for v1 to include this change. Considering [ORAS-go v2](https://github.com/oras-project/oras-go#v2-experimental) is still experimental, ORAS-go v1 will be maintained by the ORAS community until ORAS v2 GA. Currently, ORAS-go v1 is not under active development, thus the ORAS community will only publish new releases to include necessary fixes/updates at a low rate.

Issue #206, opened 2022-06-27
Ensure optional subject


shizhMSFT:

Wwwsylvia:
Closing as confirmed that subject can be null. (See https://github.com/oras-project/oras-go/blob/main/pack.go#L174 and https://github.com/oras-project/oras-go/blob/main/content/graph.go#L82-L84)

Issue #205, opened 2022-06-27
Validate with client side sorting


shizhMSFT:

shizhMSFT:
@binbin-li We need to update the `notation-go` implementation once this issue is addressed.

PR #223, opened 2022-07-01
Add PackArtifact

### LINKED ISSUE #175: creation of a new ORAS Artifact manifest should set `io.cncf.oras.artifact.created`
ORAS Artifact manifests added `io.cncf.oras.artifact.created` to account for artifact lifecycle management as an artifact is promoted across registries. See [ORAS Artifact Manifest Properties](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#oras-artifact-manifest-properties) and the [Referrers Sorting Results API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#sorting-results)

To assure all new artifacts have a `io.cncf.oras.artifact.created` value, `oras-go` should set it on manifest creation. If a user provides the value, we should accept their value, be we should have a default.

Setting the `created` annotation enables scenarios for pulling a collection of reference types, that may overlay the same information.
- I may pull a collection of scan reports for the `net-monitor:v1` image. How do I sort the collection to know the last, or even the first?
- I pull a collection of annotations, that update the `eol` date for the `net-monitor:v1` image. I need the newest value to account for the latest information.

"io.cncf.oras.artifact.created": "2022-06-13T22:11:10.08Z"

### LINKED ISSUE #202: Upgrade Pack to Support ORAS Artifact Manifest

Wwwsylvia:
Add `PackArtifact` to support packing ORAS Artifact Manifest, and set a default `io.cncf.oras.artifact.created` annotation.

Resolves #175
Resolves #202
Related: #205, #206

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -21,32 +21,61 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/opencontainers/go-digest"
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
+
 	specs "github.com/opencontainers/image-spec/specs-go"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
 )
 
-// MediaTypeUnknownConfig is the default mediaType used when no
-// config media type is specified.
-const MediaTypeUnknownConfig = "application/vnd.unknown.config.v1+json"
+const (
+	// MediaTypeUnknownConfig is the default mediaType used when no
+	// config media type is specified.
+	MediaTypeUnknownConfig = "application/vnd.unknown.config.v1+json"
+
+	// AnnotationArtifactCreated is the annotation key representing the UTC date
+	// and time on which the artifact was created, in RFC3339 format.
+	// References:
+	// - https://datatracker.ietf.org/doc/html/rfc3339#section-5.6
+	// - https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#oras-artifact-manifest-properties
+	AnnotationArtifactCreated = "io.cncf.oras.artifact.created"
+)
+
+// ErrMissingArtifactType is returned by PackArtifact() when no artifact type is
+// specified.
+var ErrMissingArtifactType = errors.New("missing artifact type")
 
 // PackOptions contains parameters for oras.Pack.
 type PackOptions struct {
-	ConfigDescriptor    *ocispec.Descriptor
-	ConfigMediaType     string
-	ConfigAnnotations   map[string]string
+	// ConfigDescriptor is a pointer to the descriptor of the config blob.
+	ConfigDescriptor *ocispec.Descriptor
+	// ConfigMediaType is the media type of the config blob.
+	// If not specified, MediaTypeUnknownConfig will be used.
+	ConfigMediaType string
+	// ConfigAnnotations is the annotation map of the config descriptor.
+	ConfigAnnotations map[string]string
+	// ManifestAnnotations is the annotation map of the manifest.
+	ManifestAnnotations map[string]string
+}
+
+// PackArtifactOptions contains parameters for oras.PackArtifact.
+type PackArtifactOptions struct {
+	// Subject is the subject of the ORAS Artifact Manifest.
+	Subject *artifactspec.Descriptor
+	// ManifestAnnotations is the annotation map of the manifest.
 	ManifestAnnotations map[string]string
 }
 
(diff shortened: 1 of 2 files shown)


pack.go
```diff
@@ -21,32 +21,61 @@ import (
...
+	Subject *artifactspec.Descriptor
+	// ManifestAnnotations is the annotation map of the manifest.
 	ManifestAnnotations map[string]string
 }
 
 // Pack packs the given layers, generates a manifest for the pack,
-// and pushes it to a content storage. If succeeded, returns a descriptor of the manifest.
+// and pushes it to a content storage.
+// If succeeded, returns a descriptor of the manifest.
 func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descriptor, opts PackOptions) (ocispec.Descriptor, error) {
```
Wwwsylvia, line 76: Should we push the layers for users? Otherwise we may want to call out "layers must be pushed before Pack() is called.".

pack.go
```diff
@@ -21,32 +21,61 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/opencontainers/go-digest"
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
+
```
shizhMSFT, line 28: nit: remove empty line

pack.go
```diff
@@ -90,7 +119,60 @@ func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descripto
...
+	return manifestDesc, nil
+}
+
+// PackArtifact packs the given blobs, generates an ORAS Artifact Manifest for
+// the pack, and pushes it to a content storage.
+// If succeeded, returns a descriptor of the manifest.
+// Returns ErrMissingArtifactType if artifactType is empty.
+// Reference: https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md
+func PackArtifact(ctx context.Context, pusher content.Pusher, blobs []artifactspec.Descriptor, artifactType string, opts PackArtifactOptions) (ocispec.Descriptor, error) {
```
shizhMSFT, line 135:
```suggestion
func PackArtifact(ctx context.Context, pusher content.Pusher, artifactType string, blobs []artifactspec.Descriptor, opts PackArtifactOptions) (ocispec.Descriptor, error) {
```

pack.go
```diff
@@ -90,7 +119,60 @@ func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descripto
...
+	if opts.ManifestAnnotations == nil {
+		opts.ManifestAnnotations = make(map[string]string)
+	}
+
+	_, ok := opts.ManifestAnnotations[AnnotationArtifactCreated]
+	if !ok {
+		// set creation time in RFC3339 format, if not provided
+		// reference: https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#oras-artifact-manifest-properties
+		now := time.Now().UTC()
+		opts.ManifestAnnotations[AnnotationArtifactCreated] = now.Format(time.RFC3339)
```
shizhMSFT, line 150: Map is a reference type. Setting `opts.ManifestAnnotations[AnnotationArtifactCreated] = now.Format(time.RFC3339)` may modify the original map, rendering `PackArtifact()` not concurrent safe.

pack.go
akashsinghal, line 135: Just want to confirm that the OCI store target could be used as the `pusher` here? I believe the interface is implemented but want to double check.

shizhMSFT: Yes, OCI store implements `Pusher` (`Pusher` is a part of `Target`)

Wwwsylvia: It's better.

pack.go
```diff
@@ -90,7 +118,61 @@ func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descripto
...
+	if !ok {
+		// set creation time in RFC3339 format, if not provided
+		// reference: https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#oras-artifact-manifest-properties
+		now := time.Now().UTC()
+		annotations[AnnotationArtifactCreated] = now.Format(time.RFC3339)
+	}
+
+	if blobs == nil {
+		blobs = []artifactspec.Descriptor{} // make it an empty array to prevent potential server-side bugs
+	}
```
Wwwsylvia, line 155: Wondering if this is really needed. 
As per [the artifact spec](https://github.com/oras-project/artifacts-spec/blob/main/artifact-manifest.md#oras-artifact-manifest-properties), `blobs` is an OPTIONAL collection of 0 or more blobs.
Since it is optional, `null` should be acceptable. Just that an empty array might be safer.

shizhMSFT (approved):
LGTM

PR #227, opened 2022-07-05
Check ORAS-Api-Version for referrers API

m5i-work:
Resoves #220 

Signed-off-by: Wei Meng <wemeng@microsoft.com>

PR #229, opened 2022-07-07
Skip version checking for pre-RC1 endpoint

m5i-work:


qweeah:
Is it possible to add a new unit case that "ORAS-Api-Version" header should not be set for legacy API？

shizhMSFT (approved):
LGTM

qweeah (approved):

shizhMSFT:
This PR fixed a fallback bug introduced in #227

PR #230, opened 2022-07-07
Validate sorting order for Referrer API

### LINKED ISSUE #205: Validate with client side sorting

m5i-work:
Resolves #205 

Client-side sorting makes pagination useless. So client only verifies order but will not sort unordered results.

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -523,8 +523,7 @@ func validateReferrersSorting(refs []artifactspec.Descriptor, last *time.Time) (
 
 // parseCreatedTime parses the created time from artifact annotation.
 func parseCreatedTime(desc artifactspec.Descriptor) (time.Time, error) {
-	// TODO: use constant from artifact-spec
-	created, ok := desc.Annotations["io.cncf.oras.artifact.created"]
+	created, ok := desc.Annotations[artifactspec.AnnotationArtifactCreated]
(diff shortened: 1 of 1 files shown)


northtyphoon:
Why does a client side sdk need to validate the sorting of the response? Will it give an opportunity to let the application layer skip the validation or ignore the error?

northtyphoon:
> MUST is used in [specification](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#sorting-results). So I think the error should not be ignored?

My understanding is the spec requirement should be honored by the producer who produce the data. I just try to understand the use case to enforce it on the consumer side, especially on a client side sdk.

Wwwsylvia:
> > MUST is used in [specification](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#sorting-results). So I think the error should not be ignored?
> 
> My understanding is the spec requirement should be honored by the producer who produces the data. I just try to understand the use case to enforce it on the consumer side, especially on a client side sdk.
> 
> I don't notice other sdk (like storage, cosmosdb) to check the order of the response items so I'm a little curious about the practice.

Agreed. It looks like the spec does not require the client to sort or to verify the sorting results.

shizhMSFT:
Make sense. The client-side sorting is not there in the spec.

PR #243, opened 2022-07-19
Bug: oras.Copy() may skip tagging if manifest blob already exists in the destination

### LINKED ISSUE #239: Bug: oras.Copy() may skip tagging if manifest blob already exists in the destination
# Reproduce Step
oras CLI version: 0.13.0
1. Package an OCI artifact and push
`echo "test" > hi.txt && oras push $REG/$REPO hi.txt`
2. Repush the artifact
`oras push $REG/$REPO:tag hi.txt`
3. Push will be successful on client side but `tag` won't be created in the registry.

# Root Cause
If a manifest blob is found existed (https://github.com/oras-project/oras-go/blob/main/copy.go#L154-L163), the copy process will be skipped as well as the retagging process.

_Originally posted by @qweeah in https://github.com/oras-project/oras/issues/447#issuecomment-1185199358_

m5i-work:
Resolves #239 

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -352,12 +352,11 @@ func prepareCopy(ctx context.Context, dst Target, dstRef string, proxy *cas.Prox
 		if !descriptor.EqualOCI(desc, root) {
 			return nil
 		}
-		// copy reference for skipped root
+		// enforce tagging when root manifest is skipped
 		if refPusher, ok := dst.(registry.ReferencePusher); ok {
 			return copyCachedNodeWithReference(ctx, proxy, refPusher, desc, dstRef)
-		} else {
-			return dst.Tag(ctx, root, dstRef)
 		}
+		return dst.Tag(ctx, root, dstRef)
 	}
 
 	return nil
--- a/copy.go
+++ b/copy.go
@@ -352,7 +352,7 @@ func prepareCopy(ctx context.Context, dst Target, dstRef string, proxy *cas.Prox
 		if !descriptor.EqualOCI(desc, root) {
 			return nil
 		}
-		// enforce tagging when root manifest is skipped
+		// enforce tagging when root is skipped
 		if refPusher, ok := dst.(registry.ReferencePusher); ok {
 			return copyCachedNodeWithReference(ctx, proxy, refPusher, desc, dstRef)
 		}
--- a/copy.go
+++ b/copy.go
@@ -353,13 +353,11 @@ func prepareCopy(ctx context.Context, dst Target, dstRef string, proxy *cas.Prox
 			return nil
 		}
 		// copy reference for skipped root
-		var err error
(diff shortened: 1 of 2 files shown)


copy_test.go
```diff
@@ -135,6 +135,100 @@ func TestCopy_FullCopy(t *testing.T) {
 	}
 }
 
+func TestCopy_ExistedRoot(t *testing.T) {
```
Wwwsylvia, line 138: Since the implementation overrides `OnCopySkipped`, we may want to verify that the original `OnCopySkipped` is called.

copy.go
```diff
@@ -342,5 +342,23 @@ func prepareCopy(ctx context.Context, dst Target, dstRef string, proxy *cas.Prox
...
+		}
+		if !descriptor.EqualOCI(desc, root) {
+			return nil
+		}
+		// copy reference for skipped root
+		if refPusher, ok := dst.(registry.ReferencePusher); ok {
+			return copyCachedNodeWithReference(ctx, proxy, refPusher, desc, dstRef)
+		} else {
+			return dst.Tag(ctx, root, dstRef)
+		}
```
qweeah, line 360:
```suggestion
		}
		return dst.Tag(ctx, root, dstRef)

```

We can save one more line.

copy.go
```diff
@@ -342,5 +342,23 @@ func prepareCopy(ctx context.Context, dst Target, dstRef string, proxy *cas.Prox
...
+	opts.OnCopySkipped = func(ctx context.Context, desc ocispec.Descriptor) error {
+		if onCopySkipped != nil {
+			if err := onCopySkipped(ctx, desc); err != nil {
+				return err
+			}
+		}
+		if !descriptor.EqualOCI(desc, root) {
+			return nil
+		}
+		// copy reference for skipped root
```
qweeah, line 355:
```suggestion
		// enforce tagging when root manifest is skipped
```

Wwwsylvia: I think we should not use the word `manifest` here. It's just a root node.

copy.go
```diff
@@ -342,5 +342,22 @@ func prepareCopy(ctx context.Context, dst Target, dstRef string, proxy *cas.Prox
...
+	opts.OnCopySkipped = func(ctx context.Context, desc ocispec.Descriptor) error {
+		if onCopySkipped != nil {
+			if err := onCopySkipped(ctx, desc); err != nil {
+				return err
+			}
+		}
+		if !descriptor.EqualOCI(desc, root) {
+			return nil
+		}
+		// enforce tagging when root manifest is skipped
```
Wwwsylvia, line 355: nit: I think we should avoid using the word `manifest` here. It's just a root node.

Wwwsylvia (approved):
LGTM with a minor suggestion

PR #244, opened 2022-07-20
feat: Support platform selection on Copy

### LINKED ISSUE #210: Support platform selection on Copy
When using [Copy](https://github.com/oras-project/oras-go/blob/main/copy.go#L84), clients may want to select a manifest of a specific platform from a Manifest List/Manifest Index that is resolved from a reference.

To support this, we can provide some utility methods that configure the [MapRoot](https://github.com/oras-project/oras-go/blob/main/copy.go#L54) option for clients.

lizMSFT:
Support platform selection on Copy by adding `AddPlatformFilter` func that configure the `MapRoot` option for clients.

Platform property ref: [OCI Image Index Specification](https://github.com/opencontainers/image-spec/blob/main/image-index.md#oci-image-index-specification)

The given platform matches the target platform if all of the below conditions are met:
- Architecture and OS exactly match.
- Variant and OSVersion exactly match if target platform provided.
- OSFeatures of the target platform are the subsets of the OSFeatures array of the given platform.
- Features property is reserved for future versions of the specification, will skip this for now. ref: [descriptor.go](https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/descriptor.go#L50)

Note: Variant, OSVersion, OSFeatures and Features are optional fields, will skip the comparison if the target platform does not provide specfic value.

If multiple manifests match a client or runtime's requirements, the first matching entry will be returned.

Resolves #210 

Signed-off-by: REDMOND\zoeyli <zoeyli@microsoft.com>

### DIFF
--- a/content.go
+++ b/content.go
@@ -44,22 +44,29 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
 	return target.Tag(ctx, desc, dst)
 }
 
+var (
+	// DefaultResolveOptions provides the default ResolveOptions.
+	DefaultResolveOptions = ResolveOptions{
+		TargetPlatform: nil,
+	}
+)
+
 // ResolveOptions contains parameters for oras.Resolve.
 type ResolveOptions struct {
-	// MatchPlaform is the target platform.
+	// TargetPlatform is the target platform.
 	// Will do the platform selection if specified.
-	MatchPlatform *ocispec.Platform
+	TargetPlatform *ocispec.Platform
 }
 
-// Resolve returns the resolved descriptor.
+// Resolve resolves a descriptor with provided reference from the target.
 func Resolve(ctx context.Context, target Target, ref string, opts ResolveOptions) (ocispec.Descriptor, error) {
 	desc, err := target.Resolve(ctx, ref)
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
 
-	if opts.MatchPlatform != nil {
-		desc, err = selectPlatform(ctx, target, desc, opts.MatchPlatform)
+	if opts.TargetPlatform != nil {
+		desc, err = selectPlatform(ctx, target, desc, opts.TargetPlatform)
 		if err != nil {
 			return ocispec.Descriptor{}, err
 		}
--- a/content.go
+++ b/content.go
@@ -18,6 +18,7 @@ package oras
 import (
 	"context"
 
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"oras.land/oras-go/v2/registry"
 )
 
@@ -41,4 +42,28 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
 		return err
 	}
 	return target.Tag(ctx, desc, dst)
+}
+
+// ResolveOptions contains parameters for oras.Resolve.
+type ResolveOptions struct {
+	// MatchPlaform is the target platform.
+	// Will do the platform selection if specified.
+	MatchPlatform *ocispec.Platform
+}
+
+// Resolve returns the resolved descriptor.
+func Resolve(ctx context.Context, target Target, ref string, opts ResolveOptions) (ocispec.Descriptor, error) {
+	desc, err := target.Resolve(ctx, ref)
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+
+	if opts.MatchPlatform != nil {
+		desc, err = selectPlatform(ctx, target, desc, opts.MatchPlatform)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+	}
+
+	return desc, nil
 }
--- a/content.go
+++ b/content.go
@@ -17,8 +17,12 @@ package oras
 
 import (
 	"context"
+	"fmt"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/internal/cas"
+	"oras.land/oras-go/v2/internal/docker"
 	"oras.land/oras-go/v2/registry"
 )
 
@@ -44,33 +48,53 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
 	return target.Tag(ctx, desc, dst)
 }
 
-var (
-	// DefaultResolveOptions provides the default ResolveOptions.
-	DefaultResolveOptions = ResolveOptions{
-		TargetPlatform: nil,
-	}
-)
+// DefaultResolveOptions provides the default ResolveOptions.
+var DefaultResolveOptions ResolveOptions
 
 // ResolveOptions contains parameters for oras.Resolve.
 type ResolveOptions struct {
-	// TargetPlatform is the target platform.
-	// Will do the platform selection if specified.
+	// TargetPlatform ensures the resolved content matches the target platform
+	// if the node is a manifest, or selects the first resolved content that
+	// matches the target platform if the node is a manifest list.
 	TargetPlatform *ocispec.Platform
 }
 
 // Resolve resolves a descriptor with provided reference from the target.
 func Resolve(ctx context.Context, target Target, ref string, opts ResolveOptions) (ocispec.Descriptor, error) {
+	if opts.TargetPlatform == nil {
+		return target.Resolve(ctx, ref)
+	}
+
+	if refFetcher, ok := target.(registry.ReferenceFetcher); ok {
+		// optimize performance for ReferenceFetcher targets
+		desc, rc, err := refFetcher.FetchReference(ctx, ref)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+		defer rc.Close()
+
+		switch desc.MediaType {
+		case docker.MediaTypeManifestList, ocispec.MediaTypeImageIndex,
+			docker.MediaTypeManifest, ocispec.MediaTypeImageManifest:
+			// create a proxy to cache the fetched descriptor
+			store := cas.NewMemory()
+			err = store.Push(ctx, desc, rc)
+			if err != nil {
+				return ocispec.Descriptor{}, err
+			}
+
+			proxy := cas.NewProxy(target, store)
+			proxy.StopCaching = true
+			return selectPlatform(ctx, proxy, desc, opts.TargetPlatform)
+		default:
+			return ocispec.Descriptor{}, fmt.Errorf("%s: %s: %w", desc.Digest, desc.MediaType, errdef.ErrUnsupported)
+		}
+	}
+
 	desc, err := target.Resolve(ctx, ref)
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
 
-	if opts.TargetPlatform != nil {
-		desc, err = selectPlatform(ctx, target, desc, opts.TargetPlatform)
-		if err != nil {
-			return ocispec.Descriptor{}, err
-		}
-	}
-
-	return desc, nil
+	return selectPlatform(ctx, target, desc, opts.TargetPlatform)
 }
--- a/copy.go
+++ b/copy.go
@@ -27,7 +27,9 @@ import (
 	"oras.land/oras-go/v2/errdef"
 	"oras.land/oras-go/v2/internal/cas"
 	"oras.land/oras-go/v2/internal/descriptor"
+	"oras.land/oras-go/v2/internal/docker"
 	"oras.land/oras-go/v2/internal/graph"
+	"oras.land/oras-go/v2/internal/platform"
 	"oras.land/oras-go/v2/internal/registryutil"
 	"oras.land/oras-go/v2/internal/status"
 	"oras.land/oras-go/v2/registry"
@@ -52,6 +54,41 @@ type CopyOptions struct {
 	// reference will be passed to MapRoot, and the mapped descriptor will be
 	// used as the root node for copy.
 	MapRoot func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error)
+}
+
+// selectPlatform implements platform filter and returns the descriptor of
+// the first matched manifest from the manifest list / image index
+func (o *CopyOptions) selectPlatform(ctx context.Context, src content.Storage, root ocispec.Descriptor, p *ocispec.Platform) (ocispec.Descriptor, error) {
+	if root.MediaType == docker.MediaTypeManifestList || root.MediaType == ocispec.MediaTypeImageIndex {
+		manifests, err := content.Successors(ctx, src, root)
+		if err != nil {
+			return ocispec.Descriptor{}, errdef.ErrNotFound
+		}
+
+		// platform filter
+		for _, m := range manifests {
+			matched := platform.MatchPlatform(m.Platform, p)
+			if matched {
+				return m, nil
+			}
+		}
+		return ocispec.Descriptor{}, errdef.ErrNotFound
+	}
+	return ocispec.Descriptor{}, errdef.ErrNotFound
+}
+
+// AddPlatformFilter adds the selectPlatform func into the MapRoot func
+func (o *CopyOptions) AddPlatformFilter(p *ocispec.Platform) {
+	mapRoot := o.MapRoot
+	o.MapRoot = func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error) {
+		var err error
+		if mapRoot != nil {
+			if root, err = mapRoot(ctx, src, root); err != nil {
+				return ocispec.Descriptor{}, err
+			}
+		}
+		return o.selectPlatform(ctx, src, root, p)
+	}
 }
 
 // CopyGraphOptions contains parameters for oras.CopyGraph.
--- a/copy.go
+++ b/copy.go
@@ -70,12 +70,12 @@ func getPlatformFromConfig(ctx context.Context, src content.Storage, desc ocispe
 	}
 	defer rc.Close()
 
-	var platform *ocispec.Platform
-	if err = json.NewDecoder(rc).Decode(&platform); err != nil {
+	var platform ocispec.Platform
+	if err = json.NewDecoder(rc).Decode(&platform); err != nil && err != io.EOF {
 		return nil, err
 	}
 
-	return platform, nil
+	return &platform, nil
 }
 
 // selectPlatform implements platform filter and returns the descriptor of the
@@ -110,9 +110,6 @@ func selectPlatform(ctx context.Context, src content.Storage, root ocispec.Descr
 		cfgPlatform, err := getPlatformFromConfig(ctx, src, descs[0], configMediaType)
 		if err != nil {
 			return ocispec.Descriptor{}, err
-		}
-		if cfgPlatform == nil {
-			return ocispec.Descriptor{}, fmt.Errorf("%s: missing platform info", descs[0].Digest)
 		}
 
 		if platform.Match(cfgPlatform, p) {
--- a/copy.go
+++ b/copy.go
@@ -57,10 +57,30 @@ type CopyOptions struct {
 	MapRoot func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error)
 }
 
-// selectPlatform implements platform filter and returns the descriptor of
-// the first matched manifest if the root is a manifest list / image index.
-// If the root is a manifest, then return the root descriptor if platform
-// matches.
+// getPlatformFromConfig returns a platform object which is made up from the
+// fields in config blob.
+func getPlatformFromConfig(ctx context.Context, src content.Storage, desc ocispec.Descriptor, targetConfigMediaType string) (*ocispec.Platform, error) {
+	if desc.MediaType != targetConfigMediaType {
+		return nil, fmt.Errorf("mismatch MediaType %s: expect %s", desc.MediaType, targetConfigMediaType)
+	}
+
+	rc, err := src.Fetch(ctx, desc)
+	if err != nil {
+		return nil, err
+	}
+	defer rc.Close()
+
+	var platform *ocispec.Platform
+	if err = json.NewDecoder(rc).Decode(&platform); err != nil {
+		return nil, err
+	}
+
+	return platform, nil
+}
+
+// selectPlatform implements platform filter and returns the descriptor of the
+// first matched manifest if the root is a manifest list. If the root is a
+// manifest, then return the root descriptor if platform matches.
 func selectPlatform(ctx context.Context, src content.Storage, root ocispec.Descriptor, p *ocispec.Platform) (ocispec.Descriptor, error) {
 	switch root.MediaType {
 	case docker.MediaTypeManifestList, ocispec.MediaTypeImageIndex:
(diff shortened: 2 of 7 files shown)


copy.go
```diff
@@ -54,6 +56,41 @@ type CopyOptions struct {
...
+	return ocispec.Descriptor{}, errdef.ErrNotFound
+}
+
+// AddPlatformFilter adds the selectPlatform func into the MapRoot func
+func (o *CopyOptions) AddPlatformFilter(p *ocispec.Platform) {
+	mapRoot := o.MapRoot
+	o.MapRoot = func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error) {
+		var err error
+		if mapRoot != nil {
+			if root, err = mapRoot(ctx, src, root); err != nil {
```
binbin-li, line 86: one question: is it possible that `root` gets changed once returned?

internal/platform/platform.go
```diff
@@ -0,0 +1,67 @@
...
+*/
+
+package platform
+
+import (
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// MatchPlatform checks whether the current platform matches the target platform.
+// MatchPlatform will return true if:
```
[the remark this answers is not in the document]
lizMSFT: Should be AND. I will change the sentence to` MatchPlatform will return true if all of the following conditions are met.`

copy.go
lizMSFT: According to line 141 in copy.go https://github.com/oras-project/oras-go/blob/bc29e42ccc18897b95d1162f6aeb9ff4e3635da0/copy.go#L141
The root will eventually be changed to the returned descriptor if there is the matched manifest

copy.go
```diff
@@ -54,6 +56,41 @@ type CopyOptions struct {
 	MapRoot func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error)
 }
 
+// selectPlatform implements platform filter and returns the descriptor of
+// the first matched manifest from the manifest list / image index
+func (o *CopyOptions) selectPlatform(ctx context.Context, src content.Storage, root ocispec.Descriptor, p *ocispec.Platform) (ocispec.Descriptor, error) {
+	if root.MediaType == docker.MediaTypeManifestList || root.MediaType == ocispec.MediaTypeImageIndex {
+		manifests, err := content.Successors(ctx, src, root)
+		if err != nil {
+			return ocispec.Descriptor{}, errdef.ErrNotFound
```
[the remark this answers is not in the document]
lizMSFT: That's a good question. In line 62, we checked that whether the MediaType of current node is manifest list or image index. If it is, then the root should have the child nodes, and function `Successors` will return the nodes directly pointed by the current node. Therefore, if there is an error occurred during the fetch processing, we will not be able to get the returned manifests, then we will not be able to filter these manifests based on the platform and find the target manifest. 

Do you have any suggestion for the error?

internal/platform/platform.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if len(target.OSFeatures) != 0 && !isSubset(curr.OSFeatures, target.OSFeatures) {
+		return false
+	}
+
+	return true
+}
+
+// isSubset returns true if all items in target slice are present in current slice
+func isSubset(curr, target []string) bool {
```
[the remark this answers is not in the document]
lizMSFT: `curr` is the string slice in the manifests, and `target` is the string slice that the target platform requires. This function is used to check whether all of the elements of the OSFeatures of the target platform are present in the current manifest. Perhaps the name `curr` and `target` are a little confusing, or I could simply change the input name to a and b, like


```
// isSubset returns true if all items in b slice are present in a slice
func isSubset(a, b []string) bool {
	if len(a) < len(b) {
		return false
	}

	set := make(map[string]bool)
	for _, v := range b {
		set[v] = true
	}
	for _, v := range a {
		if _, ok := set[v]; !ok {
			return false
		}
	}

	return true
}
```

internal/platform/platform.go
```diff
@@ -0,0 +1,67 @@
...
+	if len(curr) < len(target) {
```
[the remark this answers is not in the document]
lizMSFT: right, I will remove the length check then. Thanks for pointing this out

internal/platform/platform.go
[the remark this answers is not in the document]
lizMSFT: Made requested change.

copy.go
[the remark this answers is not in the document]
lizMSFT: Made requested change.

copy.go
qweeah: @binbin-li The root will also being used for tagging before return

copy.go
```diff
@@ -54,6 +56,41 @@ type CopyOptions struct {
...
+		// platform filter
+		for _, m := range manifests {
+			matched := platform.MatchPlatform(m.Platform, p)
+			if matched {
+				return m, nil
+			}
+		}
+		return ocispec.Descriptor{}, errdef.ErrNotFound
+	}
+	return ocispec.Descriptor{}, errdef.ErrNotFound
```
qweeah, line 77: Suggest to change the returning to indent less lines


```suggestion
	if root.MediaType != docker.MediaTypeManifestList && root.MediaType != ocispec.MediaTypeImageIndex {
	    return ocispec.Descriptor{}, errdef.ErrNotFound
	}
	
        manifests, err := content.Successors(ctx, src, root)
        if err != nil {
	        return ocispec.Descriptor{}, err
        }
        
        // platform filter
        for _, m := range manifests {
	        matched := platform.MatchPlatform(m.Platform, p)
	        if matched {
		        return m, nil
	        }
        }
        return ocispec.Descriptor{}, errdef.ErrNotFound
```

qweeah:
NIT: function comments should alway ends with a dot. https://tip.golang.org/doc/comment#Funcs

internal/platform/platform.go
```diff
@@ -0,0 +1,63 @@
...
+
+// MatchPlatform checks whether the current platform matches the target platform.
+// MatchPlatform will return true if all of the following conditions are met.
+// - Architecture and OS exactly match.
+// - Variant and OSVersion exactly match if target platform provided.
+// - OSFeatures of the target platform are the subsets of the OSFeatures array
+//   of the current platform.
+// Note: Variant, OSVersion and OSFeatures are optional fields, will skip the
+// comparison if the target platform does not provide specfic value.
+func MatchPlatform(curr *ocispec.Platform, target *ocispec.Platform) bool {
```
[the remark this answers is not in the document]
lizMSFT: good idea, I will change it.

copy.go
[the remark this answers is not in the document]
lizMSFT: Made requested change.

internal/platform/platform.go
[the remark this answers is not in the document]
lizMSFT: Yes, so I named this function as `MatchPlatform`

copy.go
```diff
@@ -54,6 +56,40 @@ type CopyOptions struct {
 	MapRoot func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error)
 }
 
+// selectPlatform implements platform filter and returns the descriptor of
+// the first matched manifest from the manifest list / image index.
+func (o *CopyOptions) selectPlatform(ctx context.Context, src content.Storage, root ocispec.Descriptor, p *ocispec.Platform) (ocispec.Descriptor, error) {
```
Wwwsylvia, line 61: Why is this a method of `CopyOptions`? Is `o` used anywhere?

copy.go
```diff
@@ -54,6 +56,40 @@ type CopyOptions struct {
...
+	for _, m := range manifests {
+		if platform.MatchPlatform(m.Platform, p) {
+			return m, nil
+		}
+	}
+	return ocispec.Descriptor{}, errdef.ErrNotFound
+}
+
+// AddPlatformFilter adds the selectPlatform func into the MapRoot func.
+func (o *CopyOptions) AddPlatformFilter(p *ocispec.Platform) {
```
Wwwsylvia, line 81: Maybe we can call this `WithPlatformFilter`?

copy.go
```diff
@@ -54,6 +56,40 @@ type CopyOptions struct {
...
+	// platform filter
+	for _, m := range manifests {
+		if platform.MatchPlatform(m.Platform, p) {
+			return m, nil
+		}
+	}
+	return ocispec.Descriptor{}, errdef.ErrNotFound
+}
+
+// AddPlatformFilter adds the selectPlatform func into the MapRoot func.
```
Wwwsylvia, line 80: This is user-facing documentation; hence we should not mention any internal function names. We should explain what's the effect of this function.

copy_test.go
```diff
@@ -494,6 +483,11 @@ func TestCopy_WithOptions(t *testing.T) {
 			},
 		},
 	}
+	targetPlatform := ocispec.Platform{
+		Architecture: "test-arc-2",
+		OS:           "test-os-2",
+	}
+	opts.AddPlatformFilter(&targetPlatform)
```
Wwwsylvia, line 490: The new option function worth a standalone test function.

copy_test.go
```diff
@@ -531,6 +525,66 @@ func TestCopy_WithOptions(t *testing.T) {
...
+	// verify tag
+	gotDesc, err = dst.Resolve(ctx, ref)
+	if err != nil {
+		t.Fatal("dst.Resolve() error =", err)
+	}
+	if !reflect.DeepEqual(gotDesc, wantDesc) {
+		t.Errorf("dst.Resolve() = %v, want %v", gotDesc, wantDesc)
+	}
+
+	// test copy with platform filter and exisiting MapRoot func, but no matching node can be found
```
Wwwsylvia, line 566: nit: typo: existing

Wwwsylvia (changes requested):

copy.go
```diff
@@ -54,6 +56,40 @@ type CopyOptions struct {
 	MapRoot func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error)
 }
 
+// selectPlatform implements platform filter and returns the descriptor of
+// the first matched manifest from the manifest list / image index.
+func (o *CopyOptions) selectPlatform(ctx context.Context, src content.Storage, root ocispec.Descriptor, p *ocispec.Platform) (ocispec.Descriptor, error) {
+	if root.MediaType != docker.MediaTypeManifestList && root.MediaType != ocispec.MediaTypeImageIndex {
```
Wwwsylvia, line 62: I'm wondering, what if the root is not a multi-platform manifest?

copy_test.go
Wwwsylvia: I think we can have a `TestCopy_WithPlatformFilterOptions` and keep the original test case as it is.

internal/platform/platform.go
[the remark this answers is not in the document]
qweeah: If this function is placed under a `manifest` package so we might have other match functions like `MatchMedia` or `MatchAnnotation`, then we need to name this func as `MatchPlatform`. But it's under `platform` package so we don't need to repeat what's already implied in the package name.

It's you choice. Feel free to resolve this comment if `MatchPlatform` makes more sense to you.

copy.go
lizMSFT: For now, the code checks the MediaType of the node first. If it is a manifest list or image index, then will call func `Successors` to fetch the nodes directly pointed by the current node, and then compare the returned manifests with the target platform. Otherwise we will return `ErrUnsupported` error.

For other media types [OCI Image Media Types](https://github.com/opencontainers/image-spec/blob/main/media-types.md) (such as Image Manifest), since they do not contain the platform property, I think it is reasonable to return the ErrUnsupported error. Besides, [crane](https://github.com/google/go-containerregistry/blob/5749ee68eb46376d695c495d15f251943fd0759f/pkg/v1/remote/descriptor.go#L152) only supports platform filter for image index/manifest list as well.

Do you have any suggestions for the expected behavior for a non multi-platform manifest? 
cc @FeynmanZhou @yizha1

internal/platform/platform.go
```diff
@@ -0,0 +1,63 @@
+/*
```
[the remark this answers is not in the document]
qweeah: Added a issue: https://github.com/oras-project/oras-go/issues/246 @lizMSFT please resolve this comment.

copy.go
Wwwsylvia: Makes sense.

copy.go
```diff
@@ -54,6 +56,40 @@ type CopyOptions struct {
 	MapRoot func(ctx context.Context, src content.Storage, root ocispec.Descriptor) (ocispec.Descriptor, error)
 }
 
+// selectPlatform implements platform filter and returns the descriptor of
+// the first matched manifest from the manifest list / image index.
+func selectPlatform(ctx context.Context, src content.Storage, root ocispec.Descriptor, p *ocispec.Platform) (ocispec.Descriptor, error) {
+	if root.MediaType != docker.MediaTypeManifestList && root.MediaType != ocispec.MediaTypeImageIndex {
+		return ocispec.Descriptor{}, errdef.ErrUnsupported
```
Wwwsylvia, line 63: nit: We may want to elaborate the error message a bit, so that callers can know which node and which media type is not supported here.

copy_test.go
```diff
@@ -658,6 +658,178 @@ func TestCopy_WithOptions(t *testing.T) {
 	}
 }
 
+func TestCopy_WithPlatformFilterOptions(t *testing.T) {
```
Wwwsylvia, line 661: Do we have a test case where `ErrUnsupported` is returned?

copy.go
qweeah: FYI: use [%w](https://golangbyexample.com/wrapping-and-unwrapping-error-golang/) to do the wrapping

copy_test.go
lizMSFT: Added👌

copy.go
lizMSFT: Made requested change

copy.go
shizhMSFT: For Docker/OCI image manifests, the platform information is stored in the config blob. You may open it and try to inspect the platform info to match the platform.

lizMSFT: I'm wondering whether it is necessary to support platform selection for a manifest.

copy.go
```diff
@@ -54,6 +57,68 @@ type CopyOptions struct {
...
+
+		for _, desc := range descs {
+			if desc.MediaType == docker.MediaTypeImage || desc.MediaType == ocispec.MediaTypeImageConfig {
+				rc, err := src.Fetch(ctx, desc)
+				if err != nil {
+					return ocispec.Descriptor{}, err
+				}
+				var currPlatform ocispec.Platform
+				err = json.NewDecoder(rc).Decode(&currPlatform)
+				if err != nil {
```
qweeah, line 93:
```suggestion
				if err = json.NewDecoder(rc).Decode(&currPlatform); err != nil {
```

copy.go
```diff
@@ -54,6 +57,68 @@ type CopyOptions struct {
...
+					return ocispec.Descriptor{}, err
+				}
+				defer rc.Close()
```
qweeah, line 96: This should be put just after the err checking of `Fetch`

copy_test.go
```diff
@@ -658,6 +658,267 @@ func TestCopy_WithOptions(t *testing.T) {
...
+	src := memory.New()
```
qweeah, line 662: Need make `test-arc-1`, `test-os-1` constants.

internal/platform/platform_test.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+func TestMatch(t *testing.T) {
+	tests := []struct {
+		curr   ocispec.Platform
```
qweeah, line 27: curr -> got
target -> want
want -> isMatched

content.go
```diff
@@ -42,3 +43,27 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
 	}
 	return target.Tag(ctx, desc, dst)
 }
+
+// ResolveOptions contains parameters for oras.Resolve.
+type ResolveOptions struct {
+	// MatchPlaform is the target platform.
+	// Will do the platform selection if specified.
+	MatchPlatform *ocispec.Platform
```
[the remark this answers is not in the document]
lizMSFT: Nice catch! thanks for pointing out

content.go
```diff
@@ -42,3 +43,27 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
 	}
 	return target.Tag(ctx, desc, dst)
 }
+
```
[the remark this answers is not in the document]
lizMSFT: Added

content.go
```diff
@@ -42,3 +43,34 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
...
+// ResolveOptions contains parameters for oras.Resolve.
+type ResolveOptions struct {
+	// TargetPlatform is the target platform.
+	// Will do the platform selection if specified.
+	TargetPlatform *ocispec.Platform
+}
+
+// Resolve resolves a descriptor with provided reference from the target.
+func Resolve(ctx context.Context, target Target, ref string, opts ResolveOptions) (ocispec.Descriptor, error) {
+	desc, err := target.Resolve(ctx, ref)
```
qweeah, line 63: @Wwwsylvia @shizhMSFT If 1) The target is an ReferenceFetcher and 2) the input ref is a tag, do we need to use FetchReference to save 1 http request?

copy.go
```diff
@@ -54,6 +57,67 @@ type CopyOptions struct {
...
+		}
+		return ocispec.Descriptor{}, errdef.ErrNotFound
+	case docker.MediaTypeManifest, ocispec.MediaTypeImageManifest:
+		descs, err := content.Successors(ctx, src, root)
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+
+		for _, desc := range descs {
+			if desc.MediaType == docker.MediaTypeImage || desc.MediaType == ocispec.MediaTypeImageConfig {
```
shizhMSFT, line 86: Let's don't mix config media types.

copy.go
```diff
@@ -54,6 +57,67 @@ type CopyOptions struct {
...
+				rc, err := src.Fetch(ctx, desc)
+				if err != nil {
+					return ocispec.Descriptor{}, err
+				}
+				defer rc.Close()
```
shizhMSFT, line 91: `defer` in a for loop may not behave as expected.

copy.go
```diff
@@ -54,6 +57,83 @@ type CopyOptions struct {
...
+	}
+
+	rc, err := src.Fetch(ctx, desc)
+	if err != nil {
+		return nil, err
+	}
+	defer rc.Close()
+
+	var platform *ocispec.Platform
+	if err = json.NewDecoder(rc).Decode(&platform); err != nil {
```
shizhMSFT, line 74: There is an edge case on golang JSON parsing. If `rc` contains literally `null`, the decode will succeed and `platform` will be set to `nil`.

content.go
```diff
@@ -42,3 +47,53 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
...
+}
+
+// Resolve resolves a descriptor with provided reference from the target.
+func Resolve(ctx context.Context, target Target, ref string, opts ResolveOptions) (ocispec.Descriptor, error) {
+	if opts.TargetPlatform == nil {
+		return target.Resolve(ctx, ref)
+	}
+
+	refFetcher, ok := target.(registry.ReferenceFetcher)
+	if ok {
```
shizhMSFT, line 69: nit:
```suggestion
	if refFetcher, ok := target.(registry.ReferenceFetcher); ok {
```

shizhMSFT (changes requested):

copy.go
shizhMSFT: Try this small example:
```go
package main

import (
	"encoding/json"
	"fmt"
	"strings"
)

func main() {
	var a *int

	err := json.NewDecoder(strings.NewReader("null")).Decode(&a)
	if err != nil {
		panic(err)
	}

	fmt.Println(a)
}
```
It outputs
```
<nil>
```

lizMSFT: ummm, that's interesting. So if rc contains literally null (e.g., image configuration JSON document contains "author": "null"), then the platform will be set to nil no matter the rc has other OS, architecture values or not?

content.go
```diff
@@ -42,3 +47,53 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
...
+		case docker.MediaTypeManifestList, ocispec.MediaTypeImageIndex,
+			docker.MediaTypeManifest, ocispec.MediaTypeImageManifest:
+			store := cas.NewMemory()
+			err = store.Push(ctx, desc, rc)
+			if err != nil {
+				return ocispec.Descriptor{}, err
+			}
+
+			proxy := cas.NewProxy(target, store)
+			proxy.StopCaching = true
```
qweeah, line 86: This is good!  Better add some comment for this block like what is optimized with the proxy.

copy.go
```diff
@@ -54,6 +57,83 @@ type CopyOptions struct {
...
+		if err != nil {
+			return ocispec.Descriptor{}, err
+		}
+
+		configMediaType := docker.MediaTypeConfig
+		if root.MediaType == ocispec.MediaTypeImageManifest {
+			configMediaType = ocispec.MediaTypeImageConfig
+		}
+
+		currPlatform, err := getPlatformFromConfig(ctx, src, descs[0], configMediaType)
```
qweeah, line 110: Suggest: how about rename `currPlatform` to `cfgPlatform`.

copy.go
lizMSFT: Got it. If the platform was set to nil, then the runtime error will be raised when calling the platform.Match func. I could add a check on the returned platform. If the platform is nil, then throw the missing platform error. What do you think?

content.go
lizMSFT: Added. 👌

content.go
```diff
@@ -42,3 +47,54 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
...
+			docker.MediaTypeManifest, ocispec.MediaTypeImageManifest:
+			store := cas.NewMemory()
+			err = store.Push(ctx, desc, rc)
+			if err != nil {
+				return ocispec.Descriptor{}, err
+			}
+
+			// create a proxy for the target storage, use the store as the cache
+			proxy := cas.NewProxy(target, store)
+			proxy.StopCaching = true
```
qweeah, line 87:
```suggestion
			docker.MediaTypeManifest, ocispec.MediaTypeImageManifest:
			// create a proxy to cache the fetched descriptor
			store := cas.NewMemory()
			err = store.Push(ctx, desc, rc)
			if err != nil {
				return ocispec.Descriptor{}, err
			}
			proxy := cas.NewProxy(target, store)
			proxy.StopCaching = true
```

lizMSFT: Resolved.

qweeah (approved):
LGTM

copy.go
lizMSFT: Changed to use platform struct instead of pointer, and also added a check on the err. If the error is io.eof, will return the empty platform then. Added tests for both null config blob and empty config blob cases.
Thanks for pointing this out.

shizhMSFT (approved):
LGTM

copy.go
```diff
@@ -54,6 +57,83 @@ type CopyOptions struct {
...
+		if platform.Match(cfgPlatform, p) {
+			return root, nil
+		}
+		return ocispec.Descriptor{}, errdef.ErrNotFound
+	default:
+		return ocispec.Descriptor{}, fmt.Errorf("%s: %s: %w", root.Digest, root.MediaType, errdef.ErrUnsupported)
+	}
+}
+
+// WithTargetPlatform adds the check on the platform attributes.
```
Wwwsylvia, line 124: Should we elaborate the doc a bit more? I think the below information can be helpful:
1. What does this function do
2. If `MapRoot` is provided, the platform selection will be applied on the mapped root

Wwwsylvia (approved):
LGTM with suggestions on doc

copy.go
lizMSFT: How about this
```
// WithTargetPlatform adds the check on the platform attributes. If MapRoot is
// provided, the platform selection will be applied on the mapped root.
// - If the root is a manifest, it will remain the same if platform matches,
// otherwise the root will be nil.
// - If the root is a manifest list, it will be reassigned to the first manifest
// with matching platform attributes.
func (o *CopyOptions) WithTargetPlatform(p *ocispec.Platform) {
```

Wwwsylvia: I feel that documenting is hard... 
Some thoughts:
> ... otherwise the root will be nil.

How about:
> ... otherwise ErrNotFound will be returned.

?

> WithTargetPlatform adds the check on the platform attributes.

This looks not very matching the function name😂

> If the root is a manifest list, it will be reassigned to the first manifest

root will be reassigned to ...? It might be a little bit strange, but I don't have a good idea at this moment😂

lizMSFT: yes it is hard😂
How about:
```
// WithTargetPlatform configures opts.MapRoot to select the manifest whose 
// platform matches the given platform. If MapRoot is provided, the platform 
// selection will be applied on the mapped root.
// - If the root is a manifest, it will remain the same if platform matches,
// otherwise ErrNotFound will be returned.
// - If the root is a manifest list, it will be changed to the first matching 
// manifest. If there is no matching manifest, will return ErrNotFound.
func (opts *CopyOptions) WithTargetPlatform(p *ocispec.Platform) {
```

Wwwsylvia (approved):
LGTM

Issue #178, opened 2022-06-16
Update Examples for ORAS Artifact Manifests


shizhMSFT:
Extending #60, we need examples for ORAS Artifact manifests.

- [x] Upload  ORAS Artifact manifest #172
- [x] Download ORAS Artifact manifest #173
- [x] Download blob of ORAS Artifact manifest #234 
- [x] Copy ORAS Artifact manifest #241
- [x] Copy artifact and its referrers (i.e.  ORAS Artifact manifests) #248

wangxiaoxuan273:
ack

wangxiaoxuan273:
Upload ORAS Artifact manifest: pr #172

wangxiaoxuan273:
Download ORAS Artifact manifest: pr #173

wangxiaoxuan273:
Download blob of ORAS Artifact manifest: pr #234

wangxiaoxuan273:
Copy ORAS Artifact manifest #241

wangxiaoxuan273:
Copy artifact and its referrers (i.e. ORAS Artifact manifests) #248

Wwwsylvia:
Closing as all sub-tasks are done.

PR #248, opened 2022-07-25
Added an example of using ExtendedCopy

wangxiaoxuan273:
Last item of issue #178

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -134,11 +134,8 @@ func TestMain(m *testing.M) {
 			w.Write(exampleSignatureManifest)
 		case strings.Contains(p, "/manifests/latest") && m == "PUT":
 			w.WriteHeader(http.StatusCreated)
-		case strings.Contains(p, "/manifests/latest") && m == "HEAD":
-			w.Header().Set("Content-Type", artifactspec.MediaTypeArtifactManifest)
-			w.Header().Set("Docker-Content-Digest", string(exampleManifestDescriptor.Digest))
-			w.Header().Set("Content-Length", strconv.Itoa(len(exampleManifest)))
-		case strings.Contains(p, "/manifests/"+string(exampleManifestDescriptor.Digest)):
+		case strings.Contains(p, "/manifests/"+string(exampleManifestDescriptor.Digest)),
+			strings.Contains(p, "/manifests/latest") && m == "HEAD":
 			w.Header().Set("Content-Type", artifactspec.MediaTypeArtifactManifest)
 			w.Header().Set("Docker-Content-Digest", string(exampleManifestDescriptor.Digest))
 			w.Header().Set("Content-Length", strconv.Itoa(len(exampleManifest)))
--- a/example_test.go
+++ b/example_test.go
@@ -24,6 +24,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"reflect"
 	"strconv"
 	"strings"
 	"testing"
@@ -44,18 +45,26 @@ var (
 	exampleManifest, _ = json.Marshal(artifactspec.Manifest{
 		MediaType:    artifactspec.MediaTypeArtifactManifest,
 		ArtifactType: "example/content"})
-	exampleManifestDescriptor = artifactspec.Descriptor{
+	exampleManifestDescriptor = ocispec.Descriptor{
 		MediaType: artifactspec.MediaTypeArtifactManifest,
 		Digest:    digest.Digest(digest.FromBytes(exampleManifest)),
 		Size:      int64(len(exampleManifest))}
+	exampleManifestDescriptorArtifactspec = artifactspec.Descriptor{
+		MediaType: exampleManifestDescriptor.MediaType,
+		Digest:    exampleManifestDescriptor.Digest,
+		Size:      exampleManifestDescriptor.Size}
 	exampleSignatureManifest, _ = json.Marshal(artifactspec.Manifest{
 		MediaType:    artifactspec.MediaTypeArtifactManifest,
 		ArtifactType: "example/signature",
-		Subject:      &exampleManifestDescriptor})
+		Subject:      &exampleManifestDescriptorArtifactspec})
 	exampleSignatureManifestDescriptor = ocispec.Descriptor{
 		MediaType: artifactspec.MediaTypeArtifactManifest,
 		Digest:    digest.FromBytes(exampleSignatureManifest),
 		Size:      int64(len(exampleSignatureManifest))}
+	exampleSignatureManifestDescriptorArtifactspec = artifactspec.Descriptor{
+		MediaType: exampleSignatureManifestDescriptor.MediaType,
+		Digest:    exampleSignatureManifestDescriptor.Digest,
+		Size:      exampleSignatureManifestDescriptor.Size}
 )
 
 func pushBlob(ctx context.Context, mediaType string, blob []byte, target oras.Target) (desc ocispec.Descriptor, err error) {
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -311,3 +343,34 @@ func Example_copyArtifactManifestRemoteToLocal() {
...
+		panic(err)
+	}
+	// copy this manifest and its referrer(exampleSignatureManifest) using ExtendedCopy
+	copiedDescriptor, err := oras.ExtendedCopy(ctx, src, exampleTag, dst, exampleTag, oras.DefaultExtendedCopyOptions)
+	if err != nil {
+		panic(err)
+	}
+	// verify that exampleManifest and its referrer exist in dst
+	fmt.Println(reflect.DeepEqual(copiedDescriptor, exampleManifestDescriptor))
+	manifestExists, _ := dst.Exists(ctx, exampleManifestDescriptor)
```
Wwwsylvia, line 368: We should not ignore the returned error.

example_test.go
```diff
@@ -311,3 +343,34 @@ func Example_copyArtifactManifestRemoteToLocal() {
...
+	fmt.Println(manifestExists)
+	referrerExists, _ := dst.Exists(ctx, exampleSignatureManifestDescriptor)
```
Wwwsylvia, line 370: We may not need to check this, as it looks like the `exampleSignatureManifestDescriptor` comes from nowhere. @qweeah Any suggestions?

Wwwsylvia (changes requested):

Wwwsylvia:
@qweeah Could you also take a look?

example_test.go
wangxiaoxuan273: Then how can I verify that the referrers are copied? Or I just don't need to check?

example_test.go
```diff
@@ -311,3 +343,34 @@ func Example_copyArtifactManifestRemoteToLocal() {
...
+	}
+	dst := memory.New()
+	ctx := context.Background()
+
+	// tag exampleManifest as latest in the remote repository
+	exampleTag := "latest"
+	err = src.Tag(ctx, exampleManifestDescriptor, exampleTag)
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
wangxiaoxuan273: removed.

example_test.go
```diff
@@ -311,3 +343,34 @@ func Example_copyArtifactManifestRemoteToLocal() {
...
+	// copy this manifest and its referrer(exampleSignatureManifest) using ExtendedCopy
+	copiedDescriptor, err := oras.ExtendedCopy(ctx, src, exampleTag, dst, exampleTag, oras.DefaultExtendedCopyOptions)
```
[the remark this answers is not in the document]
wangxiaoxuan273: Printed.

example_test.go
wangxiaoxuan273: Deleted this entire part of checking.

example_test.go
wangxiaoxuan273: From our discussion we decided that we don't need to check it.

example_test.go
```diff
@@ -311,3 +343,34 @@ func Example_copyArtifactManifestRemoteToLocal() {
 	// Output:
 	// true
 }
+
+func Example_copyArtifactManifestAndReferrersRemoteToLocal() {
```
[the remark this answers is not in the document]
wangxiaoxuan273: I think the current example after resolving other comments is too simple to need a diagram.

example_test.go
```diff
@@ -311,3 +341,27 @@ func Example_copyArtifactManifestRemoteToLocal() {
...
+	src, err := remote.NewRepository(fmt.Sprintf("%s/source", remoteHost))
+	if err != nil {
+		panic(err)
+	}
+	dst := memory.New()
+	ctx := context.Background()
+
+	tagName := "latest"
+	// ExtendedCopy will copy all reachable nodes from src (i.e. its
+	// predecessors and successors).
```
Wwwsylvia, line 358: Let's avoid using the term `predecessors` and `successors` here, as it might be hard for users to understand.
How about: copy the artifact tagged by "latest" along with all of its referrers from src to dst.

example_test.go
```diff
@@ -311,3 +341,27 @@ func Example_copyArtifactManifestRemoteToLocal() {
 	// Output:
 	// true
 }
+
+// Example_extendedCopyArtifactManifestRemoteToLocal gives an example of
+// copying an artifact manifest, its predecessors and successors from a remote
+// repository to local.
+func Example_extendedCopyArtifactManifestRemoteToLocal() {
```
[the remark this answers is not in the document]
wangxiaoxuan273: I did not want to use the term "referrer" in the name, as ExtendedCopy will copy the entire tree, not only the referrers but its reference too.

example_test.go
```diff
@@ -311,3 +341,27 @@ func Example_copyArtifactManifestRemoteToLocal() {
...
+	ctx := context.Background()
+
+	tagName := "latest"
+	// ExtendedCopy will copy all reachable nodes from src (i.e. its
+	// predecessors and successors).
+	desc, err := oras.ExtendedCopy(ctx, src, tagName, dst, tagName, oras.DefaultExtendedCopyOptions)
+	if err != nil {
+		panic(err)
+	}
+	// verify that exampleManifest and its referrer exist in dst
```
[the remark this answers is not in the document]
wangxiaoxuan273: forgot to remove that. Removed now.

example_test.go
```diff
@@ -311,3 +341,27 @@ func Example_copyArtifactManifestRemoteToLocal() {
 	// Output:
 	// true
 }
+
+// Example_extendedCopyArtifactManifestRemoteToLocal gives an example of
+// copying an artifact manifest, its predecessors and successors from a remote
```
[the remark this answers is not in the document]
wangxiaoxuan273: Changed accordingly.

example_test.go
wangxiaoxuan273: Changed accordingly.

example_test.go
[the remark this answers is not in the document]
wangxiaoxuan273: Changed accordingly.

example_test.go
```diff
@@ -311,3 +341,26 @@ func Example_copyArtifactManifestRemoteToLocal() {
 	// Output:
 	// true
 }
+
+// Example_extendedCopyArtifactAndReferrersRemoteToLocal gives an example of
+// copying an artifact along with its referrers from a remote repository to local.
+func Example_Example_extendedCopyArtifactAndReferrersRemoteToLocal() {
```
qweeah, line 347:
```suggestion
func Example_extendedCopyArtifactAndReferrersRemoteToLocal() {
```

example_test.go
```diff
@@ -311,3 +341,26 @@ func Example_copyArtifactManifestRemoteToLocal() {
...
+
+	tagName := "latest"
+	// ExtendedCopy will copy the artifact tagged by "latest" along with all of its
+	// referrers from src to dst.
+	desc, err := oras.ExtendedCopy(ctx, src, tagName, dst, tagName, oras.DefaultExtendedCopyOptions)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(desc.Digest)
```
qweeah, line 363: Shouldn't we also check the existence of `exampleSignatureManifestDescriptorArtifactspec` in the local target?

Wwwsylvia: How to check? We didn't show this descriptor in the example.

qweeah (approved):
LGTM

example_test.go
wangxiaoxuan273: changed.

Wwwsylvia (approved):
LGTM

Issue #128, opened 2022-04-15
Promote method `internal/ioutil.ReadAll` to package `content`


Wwwsylvia:
Since it's common to read and verify the fetched content, we can promote the utility method [internal/ioutil.ReadAll](https://github.com/oras-project/oras-go/blob/main/internal/ioutil/io.go#L37) to the package `content` for better usability.

wangxiaoxuan273:
Please assign this to me.

shizhMSFT:
Related to #102

PR #249, opened 2022-07-25
Replace io.ReadAll() with content.ReadAll()

wangxiaoxuan273:

Following pr #245
Following issue #128


wangxiaoxuan273:
I believe L628 of registry/remote/example_test.go should remain io.ReadAll().

content/file/file_test.go
```diff
@@ -391,7 +391,7 @@ func TestStore_Dir_Push(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to open internal gz")
 	}
-	gz, err := io.ReadAll(zrc)
+	gz, err := content.ReadAll(zrc, desc)
```
wangxiaoxuan273, line 394: Please take a close look at this test case, TestStore_Dir_Push, as I replaced some but not all io.ReadAll so that the test can pass.

shizhMSFT: `content.ReadAll()` should only applies to `io.ReadCloser`s obtained from `Fetch()` or its equivalents.

Wwwsylvia: This line has nothing to do with descriptors, so `content.ReadAll()` is not applicable here.

registry/remote/example_test.go
```diff
@@ -383,7 +383,7 @@ func ExampleRepository_Fetch_manifestByTag() {
 		panic(err)
 	}
 	defer rc.Close() // don't forget to close
-	pulledBlob, err := io.ReadAll(rc)
+	pulledBlob, err := content.ReadAll(rc, descriptor)
```
Wwwsylvia, line 386: Now that we have `content.ReadAll()`, we don't need to manually verify the fetch content anymore. (See L390-L393 and other similar ones).

Wwwsylvia (changes requested):

PR #254, opened 2022-07-27
Allow skipping unnamed blobs when pushing to file store

### LINKED ISSUE #213: Allow skipping unnamed blobs when pushing to file store
Currently, when an unnamed blob is pushed to a file store, it will be stored in a memory fallback storage by default (See https://github.com/oras-project/oras-go/blob/main/content/file/file.go#L213-L215). 
However, in some cases, clients may not want such blobs to be stored at all, even in the memory.
We should provide an option in the [file store](https://github.com/oras-project/oras-go/blob/main/content/file/file.go#L75) to allow skipping unnamed blobs.

m5i-work:
Resolves #213 

### DIFF
--- a/content/file/errors.go
+++ b/content/file/errors.go
@@ -20,6 +20,7 @@ import "errors"
 var (
 	ErrMissingName             = errors.New("missing name")
 	ErrDuplicateName           = errors.New("duplicate name")
+	ErrSkipUnnamed             = errors.New("unnamed descriptor skipped")
 	ErrPathTraversalDisallowed = errors.New("path traversal disallowed")
 	ErrOverwriteDisallowed     = errors.New("overwrite disallowed")
 	ErrStoreClosed             = errors.New("store already closed")
--- a/content/file/errors.go
+++ b/content/file/errors.go
@@ -20,8 +20,9 @@ import "errors"
 var (
 	ErrMissingName             = errors.New("missing name")
 	ErrDuplicateName           = errors.New("duplicate name")
-	errSkipUnnamed             = errors.New("unnamed descriptor skipped")
 	ErrPathTraversalDisallowed = errors.New("path traversal disallowed")
 	ErrOverwriteDisallowed     = errors.New("overwrite disallowed")
 	ErrStoreClosed             = errors.New("store already closed")
 )
+
+var errSkipUnnamed = errors.New("unnamed descriptor skipped")
--- a/content/file/file.go
+++ b/content/file/file.go
@@ -89,6 +89,9 @@ type Store struct {
 	// IgnoreNoName controls if push operations should ignore descriptors
 	// without a name. When specified, corresponding content will be discarded.
 	// Otherwise, content will be saved to a fallback storage.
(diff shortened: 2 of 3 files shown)


content/file/file.go
```diff
@@ -191,26 +196,34 @@ func (s *Store) Fetch(ctx context.Context, target ocispec.Descriptor) (io.ReadCl
...
 		return err
 	}
 
 	return s.graph.Index(ctx, s, expected)
 }
 
 // push pushes the content, matching the expected descriptor.
-// If name is not specified in the descriptor,
-// the content will be pushed to the fallback storage.
+// If name is not specified in the descriptor, the content will be either
```
Wwwsylvia, line 218: I think it'd be better to call out the default behavior first. Something like:
> the content will be pushed to the fallback storage by default, or will be discarded when Store.IgnoreNoName is set.

content/file/file_test.go
```diff
@@ -1486,6 +1486,61 @@ func TestStore_File_Push_DisableOverwrite(t *testing.T) {
...
+		t.Fatal("json.Marshal() error =", err)
+	}
+	manifestDesc := ocispec.Descriptor{
+		MediaType: ocispec.MediaTypeImageManifest,
+		Digest:    digest.FromBytes(manifestJSON),
+		Size:      int64(len(manifestJSON)),
+	}
+	tempDir := t.TempDir()
+	path := filepath.Join(tempDir, "manifest.json")
+	if err := ioutil.WriteFile(path, manifestJSON, 0444); err != nil {
```
Wwwsylvia, line 1512: Is this file used anywhere?

content/file/file_test.go
```diff
@@ -1486,6 +1486,61 @@ func TestStore_File_Push_DisableOverwrite(t *testing.T) {
...
+	// verify the manifest is not saved
+	exists, err := s.Exists(ctx, manifestDesc)
+	if err != nil {
+		t.Fatal("Store.Exists() error =", err)
+	}
+	if exists {
+		t.Errorf("Unnamed manifest is saved in file store")
+	}
+	// verify the manifest is not indexed
+	predecessors, err := s.Predecessors(ctx, configDesc)
```
Wwwsylvia, line 1535: It looks like the config itself is not pushed, so it will never have predecessors.

shizhMSFT (commented):
General question on how manifest is processed since we need the manifest while the manifests usually does not have a name.

content/file/errors.go
```diff
@@ -20,6 +20,7 @@ import "errors"
 var (
 	ErrMissingName             = errors.New("missing name")
 	ErrDuplicateName           = errors.New("duplicate name")
+	ErrSkipUnnamed             = errors.New("unnamed descriptor skipped")
```
shizhMSFT, line 23: nit: I'm curious that this error is never returned to the user, isn't it?

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #256, opened 2022-07-28
Failed to download all the files in an artifact using oras.Copy()

### LINKED ISSUE #236: Failed to download all the files in an artifact using oras.Copy()
# Repro Steps
1. Prepare two files `foo` and `bar` with same content:
```
foo bar
```
2. Pack and push an artifact with two files containing exact same content
```
{
    "schemaVersion": 2,
    "config": {
        "mediaType": "application/vnd.unknown.config.v1+json",
        "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
        "size": 2
    },
    "layers": [
        {
            "mediaType": "application/vnd.oci.image.layer.v1.tar",
            "digest": "sha256:545c0e29a021fc393a82fb4ead2f1e68f23536f876aa34b0716b2c0eb0d3d1d4",
            "size": 20,
            "annotations": {
                "org.opencontainers.image.title": "foo"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.layer.v1.tar",
            "digest": "sha256:545c0e29a021fc393a82fb4ead2f1e68f23536f876aa34b0716b2c0eb0d3d1d4",
            "size": 20,
            "annotations": {
                "org.opencontainers.image.title": "bar"
            }
        }
    ]
}
```
3. Pull the pushed artifact to a local file store, only `foo` or `bar` will be pulled

# Root cause
When copying, oras-go will deduplicate blob uploading with `status.TryCommit`(related [code](https://github.com/oras-project/oras-go/blob/main/internal/status/tracker.go#L35-L43)). Since file `foo` and `bar` are pointing to the same layer blob, only one file download will be committed. The other one will be sliently ignored.

The deduplication works for all the target types except for file. We should have a workaround for file target.

m5i-work:
Resolves #236

### DIFF
--- a/content/file/file.go
+++ b/content/file/file.go
@@ -86,6 +86,13 @@ type Store struct {
 	// When specified, saving files to existing paths will be disabled.
 	// Default value: false.
 	DisableOverwrite bool
+	// RestoreDuplicates controls if files with same content but different
+	// names are restored after push operations. When a DAG is copied, nodes
+	// with same content will be deduplicated, even if they have different
+	// names. If this is not the desired behavior for file store,
+	// RestoreDuplicates can be specified, which restores missing successor
+	// files after a node is pushed. Default value: false.
+	RestoreDuplicates bool
 
 	workingDir   string   // the working directory of the file store
 	closed       int32    // if the store is closed - 0: false, 1: true.
@@ -202,6 +209,12 @@ func (s *Store) Push(ctx context.Context, expected ocispec.Descriptor, content i
 		return err
 	}
 
+	if s.RestoreDuplicates {
+		if err := s.restoreDuplicates(ctx, expected); err != nil {
+			return fmt.Errorf("Failed to restore duplicated file: %w", err)
+		}
+	}
+
 	return s.graph.Index(ctx, s, expected)
 }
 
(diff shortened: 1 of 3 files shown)


Wwwsylvia:
I think you want to reference issue #236?

copy_test.go
```diff
@@ -658,6 +659,99 @@ func TestCopy_WithOptions(t *testing.T) {
 	}
 }
 
+func TestCopy_RestoreDuplicates(t *testing.T) {
```
qweeah, line 662: Need another one, like `TestCopy_SkipDuplicates` and only foo or bar will be downloaded

content/file/file.go
```diff
@@ -86,6 +86,13 @@ type Store struct {
 	// When specified, saving files to existing paths will be disabled.
 	// Default value: false.
 	DisableOverwrite bool
+	// RestoreDuplicates controls if files with same content but different
+	// names are restored after push operations. When a DAG is copied, nodes
+	// with same content will be deduplicated, even if they have different
+	// names. If this is not the desired behavior for file store,
+	// RestoreDuplicates can be specified, which restores missing successor
+	// files after a node is pushed. Default value: false.
+	RestoreDuplicates bool
```
shizhMSFT, line 95: Will `RestoreSuccessors` be better?

content/file/file.go
```diff
@@ -242,6 +255,40 @@ func (s *Store) push(ctx context.Context, expected ocispec.Descriptor, content i
...
+		}
+		if err := func() error {
+			desc := ocispec.Descriptor{
+				MediaType: successor.MediaType,
+				Digest:    successor.Digest,
+				Size:      successor.Size,
+			}
+			rc, err := s.Fetch(ctx, desc)
+			if err != nil {
+				return fmt.Errorf("%q: %s: %w", name, desc.MediaType, err)
```
shizhMSFT, line 278: if error is "not exists", we should continue.

content/file/file.go
```diff
@@ -86,6 +86,13 @@ type Store struct {
 	// When specified, saving files to existing paths will be disabled.
 	// Default value: false.
 	DisableOverwrite bool
+	// RestoreDuplicates controls if files with same content but different
+	// names are restored after push operations. When a DAG is copied, nodes
+	// with same content will be deduplicated, even if they have different
+	// names. If this is not the desired behavior for file store,
+	// RestoreDuplicates can be specified, which restores missing successor
+	// files after a node is pushed. Default value: false.
```
shizhMSFT, line 94: Need to update the doc and all examples related to `File`.

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #218, opened 2022-06-30
Fix runnable examples


Wwwsylvia:
The following runnable examples fail to run on the GoDoc website. We need to fix this as it may confuse users.

- [Registry interaction examples](https://pkg.go.dev/oras.land/oras-go/v2@main/registry#pkg-examples)
- [Repository interaction examples](https://pkg.go.dev/oras.land/oras-go/v2@main/registry/remote#pkg-examples)
- [Authentication examples](https://pkg.go.dev/oras.land/oras-go/v2@main/registry/remote/auth#pkg-examples)

To reproduce, click the `Run` button under the example.
![image](https://user-images.githubusercontent.com/8113721/176602727-708e57cc-25bb-4166-b64b-17aa56070dd8.png)

lizMSFT:
assigned

PR #257, opened 2022-07-28
fix: Fix runnable examples

lizMSFT:
Fix runnable examples by moving the const/variables into another file.

Reference: https://pkg.go.dev/go/doc#Example
> Playable Examples must be in a package whose name ends in "_test". An Example is "playable" (the Play field is non-nil) in either of these circumstances:
> - The example function is self-contained: the function references only identifiers from other packages (or predeclared identifiers, such as "int") and the test file does not include a dot import.
> - The entire test file is the example: the file contains exactly one example function, zero test, fuzz test, or benchmark function, and at least one top-level function, type, variable, or constant declaration other than the example function.

Using godoc to open the docs for local packages:
`go install -v golang.org/x/tools/cmd/godoc@latest`
`godoc -goroot . -play`

Resolves: #218 

Signed-off-by: Zoey Li <zoeyli@microsoft.com>


registry/example_const_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package registry_test gives examples code of functions in the registry package.
+package registry_test
```
wangxiaoxuan273, line 17: Maybe` example_test_const.go` is a better filename? `example_const_test.go` is misleading.

registry/remote/auth/example_const_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+const (
+	username     = "test_user"
+	password     = "test_password"
+	accessToken  = "test/access/token"
+	refreshToken = "test/refresh/token"
+)
+
+var (
+	host                  string
```
wangxiaoxuan273, line 25: Maybe it's better for variables to remain in the test file, while putting the consts in the const file?

lizMSFT: Some test cases did not use the constants we moved to the const file, they only use those variables. So if we put these variables back in the test file, the example will be playable then. Or do you have any suggestion?

registry/example_const_test.go
lizMSFT: Both files need to be end with `_test.go` in order to be successfully compiled.

lizMSFT:
Hi @m5i-work , that's a brilliant idea! Thank you for your suggestion.

qweeah (approved):
LGTM

internal/playable/playable.go
```diff
@@ -0,0 +1,3 @@
+package playable
+
+const ExampleUnplayable = true
```
shizhMSFT, line 3: Please document this.

internal/playable/playable.go
```diff
@@ -0,0 +1,3 @@
+package playable
```
shizhMSFT, line 1: Is `doc` a better package name?

shizhMSFT, line 1: Missing license header.

shizhMSFT (changes requested):

internal/playable/playable.go
lizMSFT: Added. Thanks for pointing this out.

internal/playable/playable.go
lizMSFT: Agree. Which one do you think is better, `doc` or `godoc`?

shizhMSFT (approved):
LGTM with suggestions

internal/playable/playable.go
shizhMSFT: It's a valid concern.

shizhMSFT: How about moving the package to `registry/remote/internal/doc` so that the `doc` package is specific.

lizMSFT: `registry/example_test.go` is also using this `ExampleUnplayable` value. If we move the package to `registry/remote/internal/doc`, will raise `invalid use of internal package` error in `registry/example_test.go`. How about moving to `registry/internal/doc`?

shizhMSFT: > How about moving to `registry/internal/doc`?

That makes sense.

shizhMSFT (approved):
LGTM

internal/playable/playable.go
lizMSFT: Rebased and made requested change.

shizhMSFT (approved):
LGTM

Issue #201, opened 2022-06-27
Support filtering predecessors by artifact type on ExtendedCopy


Wwwsylvia:
When using [ExtendedCopy](https://github.com/oras-project/oras-go/blob/main/extendedcopy.go#L62), clients may want to filter predecessors of a node by their artifact type. Specifically,

- Include predecessors whose artifact type matches a given regex
- Exclude predecessors whose artifact type matches a given regex

To support this, we can provide some utility methods that configure the [FindPredecessors](https://github.com/oras-project/oras-go/blob/main/extendedcopy.go#L54) option for clients.

Original issue: #29

Wwwsylvia:
@wangxiaoxuan273 Could you help with this?

wangxiaoxuan273:
Sure. Please assign this to me.

shizhMSFT:
Scenario: https://github.com/oras-project/oras/issues/307

wangxiaoxuan273:
There are several things we need to clarify:
1. `ArtifactType` is only supported by `remote.repository` and `Referrers()`. `ExtendedCopyOptions` only works on OCIDescriptors which does not support `ArtifactType`. Given this fact, what should be the scope of this requirement? **Should the filtering only work on `remote.repository`?** And if so, where should we check the source type?

wangxiaoxuan273:
2. Should we support both inclusion and exclusion?
3. Should we support regex?

@shizhMSFT

shizhMSFT:
> There are several things we need to clarify:
> 
> 1. `ArtifactType` is only supported by `remote.repository` and `Referrers()`. `ExtendedCopyOptions` only works on `OCIDescriptors` which does not support `ArtifactType`. Given this fact, what should be the scope of this requirement? **Should the filtering only work on `remote.repository`?** And if so, where should we check the source type? And what should be the functions' behavior on non-repository source?

It is clear if the Target implements `Referrers()`.

The question is how we can filter without referrer API for generic Targets. The solution is as simple as
1. Get the successors of a certain node
2. For each successor, check if it is an artifact manifest type. If so, read it and check its artifact type.

> 2. Should we support both inclusion and exclusion?
Yes. That's intersection.

> 3. Should we support regex?
Why not 😃

shizhMSFT:
Resolved by #258

PR #258, opened 2022-07-28
added filtering by artifact type with regex

wangxiaoxuan273:

Resolves issue #201

### DIFF
--- a/extendedcopy.go
+++ b/extendedcopy.go
@@ -254,9 +254,10 @@ func findReferrersAndFilter(rf registry.ReferrerFinder, ctx context.Context, des
 		for _, referrer := range referrers {
 			if regex.MatchString(referrer.ArtifactType) {
 				predecessors = append(predecessors, ocispec.Descriptor{
-					MediaType: referrer.MediaType,
-					Digest:    referrer.Digest,
-					Size:      referrer.Size,
+					MediaType:   referrer.MediaType,
+					Digest:      referrer.Digest,
+					Size:        referrer.Size,
+					Annotations: referrer.Annotations,
 				})
 			}
 		}
--- a/extendedcopy.go
+++ b/extendedcopy.go
@@ -212,6 +212,9 @@ func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
 			rf, ok := src.(registry.ReferrerFinder)
 			if ok {
 				predecessors, err = findReferrersAndFilter(rf, ctx, desc, regex)
+				if err != nil {
+					return predecessors, nil
+				}
 			} else {
 				predecessors, err = src.Predecessors(ctx, desc)
 			}
@@ -222,6 +225,7 @@ func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
 			return nil, err
 		}
 		var filtered []ocispec.Descriptor
+		// for each predecessor, decode the manifest and check its artifact type.
 		for _, p := range predecessors {
 			if p.MediaType == artifactspec.MediaTypeArtifactManifest {
 				if err = func() error {
@@ -247,6 +251,7 @@ func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
 	}
 }
 
+// findReferrersAndFilter filters the predecessors with Referrers.
 func findReferrersAndFilter(rf registry.ReferrerFinder, ctx context.Context, desc ocispec.Descriptor, regex *regexp.Regexp) ([]ocispec.Descriptor, error) {
 	var predecessors []ocispec.Descriptor
 	if err := rf.Referrers(ctx, desc, "", func(referrers []artifactspec.Descriptor) error {
--- a/extendedcopy.go
+++ b/extendedcopy.go
@@ -17,10 +17,12 @@ package oras
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"regexp"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	artifactspec "github.com/oras-project/artifacts-spec/specs-go/v1"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/internal/copyutil"
 	"oras.land/oras-go/v2/internal/descriptor"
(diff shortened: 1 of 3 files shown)


wangxiaoxuan273:
Further clarification needed: do we need to support doing both inclusion and exclusion at the same time (intersection) ? Right now it is either inclusion or exclusion.

wangxiaoxuan273:
To generalize the above question: should this filtering feature modify the existing `FindPredecessors`, or creating new `FindPredecessors`? My current implementation is creating a new `FindPredecessors` function that either includes or excludes on artifact type with a regex. What should this filtering feature do with existing `FindPredecessors` ? @shizhMSFT

extendedcopy.go
```diff
@@ -170,3 +173,45 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
...
 	return roots, nil
 }
+
+func (opts *ExtendedCopyGraphOptions) IncludeArtifactType(exp string) {
+	opts.FindPredecessors = matchRegexArtifactType(true, exp)
+}
+
+func (opts *ExtendedCopyGraphOptions) ExcludeArtifactType(exp string) {
+	opts.FindPredecessors = matchRegexArtifactType(false, exp)
+}
```
shizhMSFT, line 183: @wangxiaoxuan273 I was expecting something like
```go
func (opts *ExtendedCopyGraphOptions) FilterArtifactType(filter *regexp.Regexp)
```

shizhMSFT:
> To generalize the above question: should this filtering feature modify the existing `FindPredecessors`, or creating new `FindPredecessors`? My current implementation is creating a new `FindPredecessors` function that either includes or excludes on artifact type with a regex. What should this filtering feature do with existing `FindPredecessors` ? @shizhMSFT

We should invoke the existing `FindPredecessors` first, and then do filtering.

extendedcopy.go
```diff
@@ -170,3 +173,58 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
...
+		} else {
+			predecessors, err = fp(ctx, src, desc)
+		}
+		if err != nil {
+			return nil, err
+		}
+		var filtered []ocispec.Descriptor
+		for _, p := range predecessors {
+			if p.MediaType == artifactspec.MediaTypeArtifactManifest {
+				rc, err := src.Fetch(ctx, p)
```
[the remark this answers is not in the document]
wangxiaoxuan273: Yes, I forgot to close `rc`. Added.

extendedcopy.go
```diff
@@ -170,3 +173,58 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
...
+				}
+				pulledContent, err := content.ReadAll(rc, p)
+				if err != nil {
+					return nil, err
+				}
+				var manifest artifactspec.Manifest
+				if err := json.Unmarshal(pulledContent, &manifest); err != nil {
+					return nil, err
+				}
+				matched, err := regexp.MatchString(exp, manifest.ArtifactType)
```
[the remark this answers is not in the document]
wangxiaoxuan273: Changed to pre-compile.

extendedcopy.go
```diff
@@ -170,3 +173,58 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
...
+// FilterArtifactType will configure opts.FindPredecessors to filter the predecessors
+// whose artifact type matches a given regex pattern. The regex syntax supported
+// is RE2. Reference: https://github.com/google/re2/wiki/Syntax.
+func (opts *ExtendedCopyGraphOptions) FilterArtifactType(exp string) {
+	opts.FindPredecessors = matchRegexArtifactType(exp, opts.FindPredecessors)
+}
+
+// matchRegexArtifactType takes an old opts.FindPredecessors function and return
+// a new one with regex filter applied.
+func matchRegexArtifactType(exp string,
```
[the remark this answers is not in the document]
wangxiaoxuan273: Inlined

extendedcopy.go
```diff
@@ -170,3 +173,48 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
 	}
 	return roots, nil
 }
+
+// FilterArtifactType will configure opts.FindPredecessors to filter the predecessors
+// whose artifact type matches a given regex pattern. The regex syntax supported
+// is RE2. Reference: https://github.com/google/re2/wiki/Syntax.
+func (opts *ExtendedCopyGraphOptions) FilterArtifactType(pattern string) {
```
shizhMSFT, line 180: Compiling regular expressions takes time. Therefore, it is better to reuse the compiled one by passing `filter *regexp.Regexp` instead of `pattern string`.
```suggestion
func (opts *ExtendedCopyGraphOptions) FilterArtifactType(filter *regexp.Regexp) {
```

shizhMSFT (commented):
There is a performance issue on obtaining the artifact type. We should avoid unnecessary calls when `Referrer` API is available.

extendedcopy.go
```diff
@@ -170,3 +173,48 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
...
+			return nil, err
+		}
+		var filtered []ocispec.Descriptor
+		for _, p := range predecessors {
+			if p.MediaType == artifactspec.MediaTypeArtifactManifest {
+				rc, err := src.Fetch(ctx, p)
+				if err != nil {
+					return nil, err
+				}
+				defer rc.Close()
```
shizhMSFT, line 204: `defer` in a for loop is dangerous. The deferred function will not be executed at the end of each iteration of the for loop but when the function is returned.

extendedcopy.go
```diff
@@ -170,3 +173,48 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
...
+				pulledContent, err := content.ReadAll(rc, p)
```
shizhMSFT, line 205: `json.NewDecoder` is more efficient than dumping everything into the memory.

shizhMSFT (changes requested):

extendedcopy.go
```diff
@@ -170,3 +173,48 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
 	}
 	return roots, nil
 }
+
+// FilterArtifactType will configure opts.FindPredecessors to filter the predecessors
+// whose artifact type matches a given regex pattern. The regex syntax supported
```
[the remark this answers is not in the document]
wangxiaoxuan273: Now we use type regexp as input, so this problem is resolved.

extendedcopy.go
wangxiaoxuan273: changed accordingly.

extendedcopy.go
wangxiaoxuan273: Wrapped defer with an anonymous function.

extendedcopy.go
wangxiaoxuan273: Used json.NewDecoder

registry/repository.go
```diff
@@ -89,6 +90,11 @@ type ReferenceTagger interface {
 	TagReference(ctx context.Context, src, dst string) error
 }
 
+// ReferrerFinder provides the Referrers API.
```
shizhMSFT, line 93: It would be better if a reference link is provided.

extendedcopy.go
```diff
@@ -196,3 +199,75 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
...
+			// if src is a ReferrerFinder, use Referrers() to filter the predecessors.
+			rf, ok := src.(registry.ReferrerFinder)
+			if ok {
+				predecessors, err = findReferrersAndFilter(rf, ctx, desc, regex)
+				if err != nil {
+					return predecessors, nil
+				}
+			} else {
+				predecessors, err = src.Predecessors(ctx, desc)
+			}
```
shizhMSFT, line 220: The error handling is incorrect.
```suggestion
			if rf, ok := src.(registry.ReferrerFinder); ok {
				return findReferrersAndFilter(rf, ctx, desc, regex)
			}
			predecessors, err = src.Predecessors(ctx, desc)
```

registry/repository.go
wangxiaoxuan273: Added the link.

extendedcopy.go
wangxiaoxuan273: Changed accordingly.

shizhMSFT (approved):
LGTM

PR #262, opened 2022-08-04
Upload empty blob when packing

### LINKED ISSUE #260: Upload empty blob when packing
The current `Pack()` method packs a `{}` config blob for the target OCI manifest.

https://github.com/oras-project/oras-go/blob/6a09a65fcc0b96c3448e988c1727ed154c2388ea/pack.go#L82

When using OCI manifest as the medium of the artifact, the content of the config blob is unused. Therefore, it is better to have an empty blob instead of `{}`. That is,
```json
{
  "digest": "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
  "size": 0
}
```

m5i-work:
Resolves #260 

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -79,16 +79,15 @@ func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descripto
 	if opts.ConfigDescriptor != nil {
 		configDesc = *opts.ConfigDescriptor
 	} else {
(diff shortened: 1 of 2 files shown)


shizhMSFT (approved):
LGTM

PR #266, opened 2022-08-11
fix: added manifest parsing when annotation is not in the descriptor

### LINKED ISSUE #265: `FilterAnnotation` in `ExtendedCopy` does not work as expected
The `FilterAnnotation` in `ExtendedCopy` relies on the `Predecessor()`, which may be implemented by `Referrers()`. However, those APIs does not return annotations as expected and thus filter annotation will always fail.

wangxiaoxuan273:
Fix the bug that filtering does not work when no annotations exists in the descriptor. In this case, we fetch and parse the manifest to get the annotations.

Resolves #265 

### DIFF
--- a/extendedcopy.go
+++ b/extendedcopy.go
@@ -177,7 +177,9 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
 }
 
 // FilterAnnotation will configure opts.FindPredecessors to filter the predecessors
-// whose annotation matches a given regex pattern.
+// whose annotation matches a given regex pattern. For performance consideration,
+// when using both FilterArtifactType and FilterAnnotation, it's recommended to call
+// FilterArtifactType first.
 func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp.Regexp) {
 	fp := opts.FindPredecessors
 	opts.FindPredecessors = func(ctx context.Context, src content.GraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
@@ -230,7 +232,9 @@ func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp
 }
 
 // FilterArtifactType will configure opts.FindPredecessors to filter the predecessors
-// whose artifact type matches a given regex pattern.
+// whose artifact type matches a given regex pattern. For performance consideration,
+// when using both FilterArtifactType and FilterAnnotation, it's recommended to call
+// FilterArtifactType first.
 func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
 	fp := opts.FindPredecessors
 	opts.FindPredecessors = func(ctx context.Context, src content.GraphStorage, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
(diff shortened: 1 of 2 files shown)


wangxiaoxuan273:
Should we also optimize `FilterAnnotation()` with Referrers? I don't see why not.

wangxiaoxuan273:
Currently `FilterAnnotation` filters one annotation at a time. Will implement multiple filtering in another pr. This pr is for bug fix when no annotation is available in descriptor.

extendedcopy.go
```diff
@@ -194,7 +194,25 @@ func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp
 		}
 		var filtered []ocispec.Descriptor
 		for _, p := range predecessors {
-			if regex.MatchString(p.Annotations[key]) {
+			if p.Annotations == nil || p.Annotations[key] == "" {
```
Wwwsylvia, line 197: In Golang, this can be simplified as:
```suggestion
			if p.Annotations[key] == "" {
```

Or consider,
` if val, ok := p.Annotations[key]; ok {}`
See https://go.dev/blog/maps.

But what do we want to do for the 1.b scenario in below?
1. The descriptor has annotations
    a. The annotation map contains the desired key
     - Check if the value matches the regex
   
    b. The annotation map does not contain the desired key
     - Do we still need to parse the content in this case?
2. The descriptor has no annotation
     - Parse the annotations from the content

shizhMSFT, line 197: This logic is indeed interesting and well perf optimized.

Wwwsylvia:
> Should we also optimize `FilterAnnotation()` with `Referrers`? I don't see why not.

What differences will that make?

extendedcopy.go
wangxiaoxuan273: Good question. The current implementation is to parse the content, since `if p.Annotations[key] == ""` represents both cases of no annotation and annotation not having the desired key. I think this is fine? @shizhMSFT

wangxiaoxuan273:
> > Should we also optimize `FilterAnnotation()` with `Referrers`? I don't see why not.
> 
> How many differences will that make?

We used Referrers in `FilterArtifactType`. Since these 2 functions do essentially the same thing, I wonder if we need Referrers in this function too.

extendedcopy.go
wangxiaoxuan273: Used the other logic posted in the group, so this issue does not exist anymore.

extendedcopy.go
```diff
@@ -176,11 +177,9 @@ func findRoots(ctx context.Context, storage content.GraphStorage, node ocispec.D
 }
 
 // FilterAnnotation will configure opts.FindPredecessors to filter the predecessors
-// whose annotation matches a given regex pattern. For performance consideration,
-// when using both FilterArtifactType and FilterAnnotation, it's recommended to call
```
Wwwsylvia, line 180: We still need these comments as per the offline discussion.

Wwwsylvia (approved):
LGTM with suggestions on doc

extendedcopy.go
wangxiaoxuan273: Added them back :p

Wwwsylvia (approved):
LGTM

Issue #225, opened 2022-07-04
oras pull error: empty response Docker-Content-Digest


johankees:
Pulling a container from **AWS ECR** fails with error `empty response Docker-Content-Digest`.

Steps to reproduce:
* login as per https://oras.land/implementors/#amazon-elastic-container-registry-ecr
* `oras pull ${ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/${IMAGE}:${TAG}`

Expected result:
* Image pulled successfully, and response:
    ```text
    Pulled ACCOUNT_ID.dkr.ecr.AWS_REGION.amazonaws.com/IMAGE:TAG
    Digest: sha256:<SOME_SHA>
    ```

Actual result:
* Image failed to be pulled:
    ```text
    Error: GET "https://ACCOUNT_ID.dkr.ecr.REGION.amazonaws.com/v2/IMAGE/manifests/TAG": empty response Docker-Content-Digest
    ```

Verbose error message:
```
-> oras pull ACCOUNT_ID.dkr.ecr.AWS_REGION.amazonaws.com/IMAGE:TAG -d
DEBU[0000]  Request URL: "https://ACCOUNT_ID.dkr.ecr.AWS_REGION.amazonaws.com/v2/IMAGE/manifests/TAG"
DEBU[0000]  Request method: "GET"
DEBU[0000]  Request headers:
DEBU[0000]    "Accept": "application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, application/vnd.cncf.oras.artifact.manifest.v1+json"
DEBU[0000]    "User-Agent": "oras/0.13.0"
DEBU[0000]  Response Status: "401 Unauthorized"
DEBU[0000]  Response headers:
DEBU[0000]    "Docker-Distribution-Api-Version": "registry/2.0"
DEBU[0000]    "Sizes": ""
DEBU[0000]    "Www-Authenticate": "Basic realm=\"https://ACCOUNT_ID.dkr.ecr.AWS_REGION.amazonaws.com/\",service=\"ecr.amazonaws.com\""
DEBU[0000]    "Date": "Mon, 04 Jul 2022 13:06:58 GMT"
DEBU[0000]    "Content-Length": "15"
DEBU[0000]    "Content-Type": "text/plain; charset=utf-8"
[...]

PR #279, opened 2022-08-23
refactor: Small refactor of reference module and tests

nima:
This is the nice-to-have changes that was split out of #225.

Signed-off-by: Nima Talebi <nima@users.noreply.github.com>

### DIFF
--- a/registry/reference.go
+++ b/registry/reference.go
@@ -29,8 +29,8 @@ import (
 var (
 	// repositoryRegexp is adapted from the distribution implementation. The
 	// repository name set under OCI distribution spec is a subset of the docker
-	// spec. For maximum compability, the docker spec is verified at the client
-	// side. Further check is left to the server side.
+	// spec. For maximum compatability, the docker spec is verified client-side.
+	// Further checks are left to the server-side.
 	// References:
 	// - https://github.com/distribution/distribution/blob/v2.7.1/reference/regexp.go#L53
 	// - https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests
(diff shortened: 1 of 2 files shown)


registry/reference.go
```diff
@@ -56,33 +56,65 @@ type Reference struct {
...
 
-// ParseReference parses a string into a artifact reference.
-// If the reference contains both the tag and the digest, the tag will be
-// dropped.
-// Digest is recognized only if the corresponding algorithm is available.
-func ParseReference(raw string) (Reference, error) {
-	parts := strings.SplitN(raw, "/", 2)
+// ParseReference parses a string (artifact) into an `artifact reference`.
+//
+// Note: An "image" is an "artifact", however, an "artifact" is not necessarily
```
Wwwsylvia, line 61: It looks like this sentence is incomplete.

shizhMSFT, line 61: It seems this sentence is not complete.

registry/reference.go
```diff
@@ -56,33 +56,65 @@ type Reference struct {
 	Reference string
 }
 
-// ParseReference parses a string into a artifact reference.
-// If the reference contains both the tag and the digest, the tag will be
-// dropped.
-// Digest is recognized only if the corresponding algorithm is available.
-func ParseReference(raw string) (Reference, error) {
-	parts := strings.SplitN(raw, "/", 2)
+// ParseReference parses a string (artifact) into an `artifact reference`.
```
shizhMSFT, line 59: image

The doc looks nice.

shizhMSFT (changes requested):

shizhMSFT (commented):
It would be even better if we have a [testable example](https://go.dev/blog/examples) for `ParseReference()`

shizhMSFT:
> It would be even better if we have a [testable example](https://go.dev/blog/examples) for `ParseReference()`

We can do this in another PR.

registry/reference.go
nima: 🤣 Fixed.

nima: Fixed 😄

Wwwsylvia (approved):
LGTM

Wwwsylvia:
@nima Can you fix the DCO?

shizhMSFT (approved):
LGTM

PR #288, opened 2022-09-01
fix: remove `+json` encoding from the default config media type

### LINKED ISSUE #285: Default mediatype should not include encoding
Based on OCI call discussion today, we should remove +json as a default

Wwwsylvia:
Resolves #285 


shizhMSFT (approved):
LGTM

PR #296, opened 2022-09-05
fix: use a default value when `Concurrency` is not specified

### LINKED ISSUE #289: The default `Concurrency` limit should not be infinity
Currently, when [`Concurrency`](https://github.com/oras-project/oras-go/blob/main/copy.go#L79-L82) is not specified, its value will be considered as infinity. However, this can be insecure. We should use a default value instead of infinity.

Wwwsylvia:
Fixes #289 

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -33,15 +33,16 @@ import (
 	"oras.land/oras-go/v2/registry"
 )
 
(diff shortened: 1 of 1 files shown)


shizhMSFT (approved):
LGTM

PR #299, opened 2022-09-05
feat!: Support specifying `MaxMetadataBytes` option for `oras.Resolve` and `oras.Copy`

### LINKED ISSUE #291: `oras.Resolve` and `oras.Copy` should have a `MaxMetadataBytes` option
Since `oras.Resolve` and `oras.Copy` may cache content, they should have a `MaxMetadataBytes` option so that the memory won't be blew up.

### LINKED ISSUE #298: Bug: Failing to push to the cache of `internal.Proxy` will result in a deadlock
When using [`internal.Proxy`](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go#L32), we fetch data from the base storage while pushing it to the cache storage.
If any errors occur when pushing to the cache storage, there will be a deadlock. This is because the [pipe ](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go#L62)is not properly closed when there is a [push error](https://github.com/oras-project/oras-go/blob/main/internal/cas/proxy.go#L68).

Wwwsylvia:
Resolves #291 
Fixes #298
BREAKING CHANGE: Rename `oras.TagBytesOptions` to `oras.TagBytesNOptions`

### DIFF
--- a/content.go
+++ b/content.go
@@ -56,7 +56,12 @@ func Tag(ctx context.Context, target Target, src, dst string) error {
 }
 
 // DefaultResolveOptions provides the default ResolveOptions.
-var DefaultResolveOptions ResolveOptions
+var DefaultResolveOptions = ResolveOptions{
+	MaxMetadataBytes: defaultMaxMetadataBytes,
+}
+
+// defaultMaxMetadataBytes is the default value of MaxMetadataBytes.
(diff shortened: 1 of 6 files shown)


shizhMSFT (approved):
LGTM

PR #322, opened 2022-09-19
feat: return errdef.ErrUnsupported when resp.StatusCode equals 405 on delete()

### LINKED ISSUE #317: Return `errdef.ErrUnsupported` on `Delete()` if remote repository does not support delete
When the remote does not support delete or has delete disabled, it will return `405 not supported`.
```
Error: failed to delete localhost:12900/0818hello@sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855: DELETE "http://localhost:12900/v2/0818hello/blobs/sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855": unexpected status code 405: unsupported: The operation is unsupported.
```

We should return `errdef.ErrUnsupported` as a known error.

Original discussion: https://github.com/oras-project/oras/pull/493#issuecomment-1244930279

yuehaoliang-microsoft:
Resolve #317 

Signed-off-by: Haoliang Yue <yuehaoliang@microsoft.com>

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -456,6 +456,8 @@ func (r *Repository) delete(ctx context.Context, target ocispec.Descriptor, isMa
 		return verifyContentDigest(resp, target.Digest)
 	case http.StatusNotFound:
 		return fmt.Errorf("%s: %w", target.Digest, errdef.ErrNotFound)
+	case http.StatusMethodNotAllowed:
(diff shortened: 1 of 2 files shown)


yuehaoliang-microsoft:
Why Codecov shows that misses + 2? I think new lines should be reached by unit test.

PR #326, opened 2022-09-19
fix: Scrutinize references for digests posing as tags

### LINKED ISSUE #303: Review `registry.ParseReference` function
There might be some invalid forms that `registry.ParseReference` does not validate. We need to review this function and test it against more test cases.
https://github.com/oras-project/oras-go/blob/d9a8f8245088a920e910f27c0fb4046ac53b84f8/registry/reference.go#L93-L130

nima:
This change prevents a token sneaking in as a *tag*, then later getting validated as a *digest*.  In order to do that, a new struct member has been introduced to `registry.Reference`, namely, `ReferenceType`.  This can take one of the small finite set of values: `Undefined`, `None`, `Tag`, and `Digest`.  `None` is applicable only to *valid form D*, and `Undefined` is the `zero/null/empty` default value to force explicitness.

~~Also note that in *valid form B*, we're still allowing the redundant `tag`, is there a reason to allow for this (as opposed to considering it ambiguous, and deeming it invalid)?~~ just noticed this was already answered by @Wwwsylvia in #303.

Resolves #303.

Signed-off-by: Nima Talebi <github@nima.id.au>

### DIFF
--- a/registry/reference.go
+++ b/registry/reference.go
@@ -81,12 +81,12 @@ type Reference struct {
 // That is, of all the possible forms that `path` can take, there are exactly 4
 // that are considered valid.  Expanding the BNF notation for `path`, the 4 are:
 //
-//	<---------------------- path ----------------------> |  - Decode `path`
-//	<=== REPOSITORY ===><---------- reference ---------> |    - Decode `reference`
-//	<=== REPOSITORY ===><=================== digest ===> |      - Valid Form A
-//	<=== REPOSITORY ===><!!! TAG !!!> @ <=== digest ===> |      - Valid Form B
-//	<=== REPOSITORY ===><=== TAG ======================> |      - Valid Form C
-//	<=== REPOSITORY ===================================> |    - Valid Form D
+//	<--- path --------------------------------------------> |  - Decode `path`
+//	<=== REPOSITORY ===> <--- reference ------------------> |    - Decode `reference`
+//	<=== REPOSITORY ===> @ <=================== digest ===> |      - Valid Form A
+//	<=== REPOSITORY ===> : <!!! TAG !!!> @ <=== digest ===> |      - Valid Form B (tag is dropped)
+//	<=== REPOSITORY ===> : <=== TAG ======================> |      - Valid Form C
+//	<=== REPOSITORY ======================================> |    - Valid Form D
 //
 // Note: In the case of Valid Form B, TAG is dropped without any validation or
 // further consideration.
@@ -98,54 +98,75 @@ func ParseReference(artifact string) (Reference, error) {
 	}
 	registry, path := parts[0], parts[1]
 
+	var isTag bool
 	var repository string
 	var reference string
 	if index := strings.Index(path, "@"); index != -1 {
 		// `digest` found; Valid Form A (if not B)
+		isTag = false
 		repository = path[:index]
 		reference = path[index+1:]
 
-		if index := strings.Index(repository, ":"); index != -1 {
+		if index = strings.Index(repository, ":"); index != -1 {
 			// `tag` found (and now dropped without validation) since `the
 			// `digest` already present; Valid Form B
 			repository = repository[:index]
 		}
-	} else if index := strings.Index(path, ":"); index != -1 {
+	} else if index = strings.Index(path, ":"); index != -1 {
 		// `tag` found; Valid Form C
+		isTag = true
 		repository = path[:index]
 		reference = path[index+1:]
 	} else {
 		// empty `reference`; Valid Form D
 		repository = path
 	}
-	res := Reference{
+	ref := Reference{
 		Registry:   registry,
 		Repository: repository,
 		Reference:  reference,
 	}
-	if err := res.Validate(); err != nil {
+
+	if err := ref.ValidateRegistry(); err != nil {
 		return Reference{}, err
 	}
-	return res, nil
+
+	if err := ref.ValidateRepository(); err != nil {
+		return Reference{}, err
+	}
+
+	if len(ref.Reference) == 0 {
+		return ref, nil
+	}
+
+	validator := ref.ValidateReferenceAsDigest
+	if isTag {
+		validator = ref.ValidateReferenceAsTag
+	}
+	if err := validator(); err != nil {
+		return Reference{}, err
+	}
+
+	return ref, nil
 }
 
-// Validate validates the entire reference.
+// Validate the entire reference object; the registry, the repository, and the
+// reference.
 func (r Reference) Validate() error {
-	err := r.ValidateRegistry()
-	if err != nil {
+	if err := r.ValidateRegistry(); err != nil {
 		return err
 	}
-	err = r.ValidateRepository()
-	if err != nil {
+
+	if err := r.ValidateRepository(); err != nil {
 		return err
 	}
+
 	return r.ValidateReference()
 }
 
 // ValidateRegistry validates the registry.
 func (r Reference) ValidateRegistry() error {
-	uri, err := url.ParseRequestURI("dummy://" + r.Registry)
-	if err != nil || uri.Host != r.Registry {
+	if uri, err := url.ParseRequestURI("dummy://" + r.Registry); err != nil || uri.Host != r.Registry {
 		return fmt.Errorf("%w: invalid registry", errdef.ErrInvalidReference)
 	}
 	return nil
@@ -159,18 +180,34 @@ func (r Reference) ValidateRepository() error {
 	return nil
 }
 
-// ValidateReference validates the reference.
-func (r Reference) ValidateReference() error {
-	if r.Reference == "" {
-		return nil
-	}
-	if _, err := r.Digest(); err == nil {
-		return nil
-	}
+// ValidateReferenceAsTag validates the reference as a tag.
+func (r Reference) ValidateReferenceAsTag() error {
 	if !tagRegexp.MatchString(r.Reference) {
 		return fmt.Errorf("%w: invalid tag", errdef.ErrInvalidReference)
 	}
 	return nil
+}
+
+// ValidateReferenceAsDigest validates the reference as a digest.
+func (r Reference) ValidateReferenceAsDigest() error {
+	if _, err := r.Digest(); err != nil {
+		return fmt.Errorf("%w: invalid digest; %v", errdef.ErrInvalidReference, err)
+	}
+	return nil
+}
+
+// ValidateReference where the reference is first tried as an ampty string, then
+// as a digest, and if that fails, as a tag.
+func (r Reference) ValidateReference() error {
+	if len(r.Reference) == 0 {
+		return nil
+	}
+
+	if index := strings.IndexByte(r.Reference, ':'); index != -1 {
+		return r.ValidateReferenceAsDigest()
+	}
+
+	return r.ValidateReferenceAsTag()
 }
 
 // Host returns the host name of the registry.
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -194,25 +194,35 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
 func (r *Repository) ParseReference(reference string) (registry.Reference, error) {
 	ref, err := registry.ParseReference(reference)
 	if err != nil {
-		// reference is not a FQDN
-		if index := strings.IndexByte(reference, '@'); index != -1 {
-			// drop tag since the digest is present
-			reference = reference[index+1:]
-		}
 		ref = registry.Reference{
 			Registry:   r.Reference.Registry,
 			Repository: r.Reference.Repository,
 			Reference:  reference,
 		}
-		if err = ref.ValidateReference(); err != nil {
+
+		// reference is not a FQDN
+		if index := strings.IndexByte(reference, '@'); index != -1 {
+			// `@` implies *digest*, so drop the *tag* (irrespective of what it is).
+			ref.Reference = reference[index+1:]
+			err = ref.ValidateReferenceAsDigest()
+		} else {
+			err = ref.ValidateReference()
+		}
+
+		if err != nil {
 			return registry.Reference{}, err
 		}
 	} else if ref.Registry != r.Reference.Registry || ref.Repository != r.Reference.Repository {
-		return registry.Reference{}, fmt.Errorf("%w %q: expect %q", errdef.ErrInvalidReference, ref, r.Reference)
+		return registry.Reference{}, fmt.Errorf(
+			"%w: mismatch between received %q and expected %q",
+			errdef.ErrInvalidReference, ref, r.Reference,
+		)
 	}
-	if ref.Reference == "" {
-		return registry.Reference{}, fmt.Errorf("%w %q: empty reference", errdef.ErrInvalidReference, ref)
+
+	if len(ref.Reference) == 0 {
+		return registry.Reference{}, errdef.ErrInvalidReference
 	}
+
 	return ref, nil
 }
 
@@ -985,7 +995,7 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref registry.Ref
 	if serverHeaderDigestStr := resp.Header.Get(dockerContentDigestHeader); serverHeaderDigestStr != "" {
 		if serverHeaderDigest, err = digest.Parse(serverHeaderDigestStr); err != nil {
 			return ocispec.Descriptor{}, fmt.Errorf(
-				"%s %q: invalid response header value `%s`: `%s`; %w",
+				"%s %q: invalid response header value: `%s: %s`; %w",
 				resp.Request.Method,
 				resp.Request.URL,
 				dockerContentDigestHeader,
@@ -1004,7 +1014,7 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref registry.Ref
 				// HEAD without server `Docker-Content-Digest` header is an
 				// immediate fail
 				return ocispec.Descriptor{}, fmt.Errorf(
-					"HTTP %s request missing required header `%s`",
+					"HTTP %s request missing required header %q",
 					httpMethod, dockerContentDigestHeader,
 				)
 			}
@@ -1025,7 +1035,7 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref registry.Ref
 
 	if len(refDigest) > 0 && refDigest != contentDigest {
 		return ocispec.Descriptor{}, fmt.Errorf(
-			"%s %q: invalid response; digest mismatch: `%s: %s` vs expected `%s`",
+			"%s %q: invalid response; digest mismatch in %s: received %q when expecting %q",
 			resp.Request.Method, resp.Request.URL,
 			dockerContentDigestHeader, contentDigest,
 			refDigest,
@@ -1076,7 +1086,7 @@ func verifyContentDigest(resp *http.Response, expected digest.Digest) error {
 
 	if contentDigest != expected {
 		return fmt.Errorf(
-			"%s %q: invalid response; digest mismatch: `%s: %s` vs expected `%s`",
+			"%s %q: invalid response; digest mismatch in %s: received %q when expecting %q",
 			resp.Request.Method, resp.Request.URL,
 			dockerContentDigestHeader, contentDigest,
 			expected,
--- a/registry/reference_test.go
+++ b/registry/reference_test.go
@@ -22,6 +22,9 @@ import (
 	"testing"
 )
 
+const ValidDigest = "sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9"
+const InvalidDigest = "sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde"
+
 // For a definition of what a "valid form [ABCD]" means, see reference.go.
 func TestParseReferenceGoodies(t *testing.T) {
 	tests := []struct {
@@ -31,18 +34,26 @@ func TestParseReferenceGoodies(t *testing.T) {
 	}{
 		{
 			name:  "digest reference (valid form A)",
-			image: "hello-world@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9",
+			image: fmt.Sprintf("hello-world@%s", ValidDigest),
 			wantTemplate: Reference{
 				Repository: "hello-world",
-				Reference:  "sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9",
+				Reference:  ValidDigest,
 			},
 		},
 		{
 			name:  "tag with digest (valid form B)",
-			image: "hello-world:v2@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9",
+			image: fmt.Sprintf("hello-world:v2@%s", ValidDigest),
 			wantTemplate: Reference{
 				Repository: "hello-world",
-				Reference:  "sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9",
+				Reference:  ValidDigest,
+			},
+		},
+		{
+			name:  "empty tag with digest (valid form B)",
+			image: fmt.Sprintf("hello-world:@%s", ValidDigest),
+			wantTemplate: Reference{
+				Repository: "hello-world",
+				Reference:  ValidDigest,
 			},
 		},
 		{
@@ -109,6 +120,22 @@ func TestParseReferenceUglies(t *testing.T) {
 		{
 			name: "invalid port",
 			raw:  "localhost:v1/hello-world",
+		},
+		{
+			name: "invalid digest",
+			raw:  fmt.Sprintf("registry.example.com/foobar@%s", InvalidDigest),
+		},
+		{
+			name: "invalid digest prefix: colon instead of the at sign",
+			raw:  fmt.Sprintf("registry.example.com/hello-world:foobar:%s", ValidDigest),
+		},
+		{
+			name: "invalid digest prefix: double at sign",
+			raw:  fmt.Sprintf("registry.example.com/hello-world@@%s", ValidDigest),
+		},
+		{
+			name: "invalid digest prefix: space",
+			raw:  fmt.Sprintf("registry.example.com/hello-world @%s", ValidDigest),
 		},
 	}
 
--- a/registry/remote/repository_test.go
+++ b/registry/remote/repository_test.go
@@ -3590,18 +3590,6 @@ func TestRepository_ParseReference(t *testing.T) {
 			wantErr: errdef.ErrInvalidReference,
 		},
 		{
-			name: "missing reference after @",
-			repoRef: registry.Reference{
-				Registry:   "registry.example.com",
-				Repository: "hello-world",
-			},
-			args: args{
-				reference: "registry.example.com/hello-world@",
-			},
-			want:    registry.Reference{},
-			wantErr: errdef.ErrInvalidReference,
-		},
-		{
 			name: "registry mismatch",
 			repoRef: registry.Reference{
 				Registry:   "registry.example.com",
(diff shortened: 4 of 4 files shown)


registry/reference.go
```diff
@@ -54,8 +54,19 @@ type Reference struct {
 	// Reference is the reference of the object in the repository.
 	// A reference can be a tag or a digest.
 	Reference string
+
+	ReferenceType ReferenceType
```
shizhMSFT, line 58: We should not introduce `ReferenceType` since the sets of tags and digests do not have intersection. Introducing `ReferenceType` will make the design unnecessarily overcomplex.

registry/remote/repository.go
```diff
@@ -828,7 +837,7 @@ func (s *manifestStore) Resolve(ctx context.Context, reference string) (ocispec.
 
 // FetchReference fetches the manifest identified by the reference.
 // The reference can be a tag or digest.
-func (s *manifestStore) FetchReference(ctx context.Context, reference string) (desc ocispec.Descriptor, rc io.ReadCloser, err error) {
+func (s *manifestStore) FetchReference(ctx context.Context, reference string) (ocispec.Descriptor, io.ReadCloser, error) {
```
shizhMSFT, line 840: The return parameters are named for their special purposes. Removing the names will introduce bugs in line 858-862.

shizhMSFT (changes requested):
I was expecting that updating the implementation of the `ParseReference()` function only without touching others should fix the bug.

registry/reference.go
nima: Taking a different approach in order to avoid modification of this struct.

errdef/errors.go
```diff
@@ -19,12 +19,13 @@ import "errors"
...
-	ErrInvalidReference   = errors.New("invalid reference")
-	ErrInvalidDigest      = errors.New("invalid digest")
-	ErrNotFound           = errors.New("not found")
-	ErrUnsupportedVersion = errors.New("unsupported version")
-	ErrMissingReference   = errors.New("missing reference")
-	ErrSizeExceedsLimit   = errors.New("size exceeds limit")
+	ErrAlreadyExists            = errors.New("already exists")
+	ErrUnsupported              = errors.New("unsupported")
+	ErrInvalidReference         = errors.New("invalid reference")
+	ErrInvalidReferenceTerminal = errors.New("terminally invalid reference")
```
nima, line 25: In `Repository.ParseReference`, the error `InvalidReference` signals an error, but one where a recovery is attempted (`if err != nil { // recovery }`.  This new error type lets the caller know that a recovery should not be attempted, and this is a terminal state.

registry/reference.go
```diff
@@ -40,6 +40,8 @@ var (
 	// The docker and OCI spec have the same regular expression.
 	// Reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests
 	tagRegexp = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
+
+	pathRegex = regexp.MustCompile(`^[a-z-][-\w.]*(:\w[\w.-]{0,127})?(@\w+:[0-9a-f]+)?$`)
```
nima, line 44: This can be tightened further, but it aims to be an improvement over the current validation.

registry/reference.go
```diff
@@ -114,6 +120,9 @@ func ParseReference(artifact string) (Reference, error) {
 		// `tag` found; Valid Form C
 		repository = path[:index]
 		reference = path[index+1:]
+		if !tagRegexp.MatchString(reference) {
+			return Reference{}, errdef.ErrInvalidReferenceTerminal
```
nima, line 124: Here we have a digest that snuck into this code block under the pretense of being a tag.

registry/remote/repository.go
```diff
@@ -195,6 +195,10 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
 // wrapped error ErrInvalidReference.
 func (r *Repository) ParseReference(reference string) (registry.Reference, error) {
 	ref, err := registry.ParseReference(reference)
+	if errors.Is(err, errdef.ErrInvalidReferenceTerminal) {
```
nima, line 198: Exit immediately on the new terminal error; do not attempt recovery.

registry/remote/repository.go
```diff
@@ -195,6 +195,10 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
...
+		return registry.Reference{}, err
+	}
+
 	if err != nil {
```
nima, line 202: This is what is referred to by *error recovery* in the statements above.

nima (commented):
Annotated the changes in an attempt to add more clarity.

nima:
Added a couple more tests in the last push; no code-changes otherwise.

errdef/errors.go
shizhMSFT: How do you define `ErrInvalidReferenceTerminal`? Users will be confused with `ErrInvalidReference` and `ErrInvalidReferenceTerminal`.

We should only have one error. That is, `ErrInvalidReference`.

registry/remote/repository.go
shizhMSFT: The error recovery is by design since a `reference` passed to `Repository.ParseReference()` can be a FQDN like `localhost:5000/alpine:latest` or just the tag `latest`.

If we exit immediately, we will miss that later case.

shizhMSFT (changes requested):

registry/reference.go
```diff
@@ -160,16 +163,19 @@ func (r Reference) ValidateRepository() error {
 }
 
 // ValidateReference validates the reference.
-func (r Reference) ValidateReference() error {
-	if r.Reference == "" {
-		return nil
-	}
-	if _, err := r.Digest(); err == nil {
+func (r Reference) ValidateReference(isTag bool) error {
```
shizhMSFT, line 166: `isTag` sounds reasonable. Could we have 3 separated explicit methods like
```go
func (r Reference) ValidateReference() error
func (r Reference) ValidateReferenceAsTag() error
func (r Reference) ValidateReferenceAsDigest() error
```
?

shizhMSFT (changes requested):

registry/reference.go
nima: Is this to reduce cyclomatic complexity?  I've made the change, but not sure if it's what you're expecting; let me know what you think.

registry/reference.go
```diff
@@ -164,13 +175,26 @@ func (r Reference) ValidateReference() error {
 	if r.Reference == "" {
 		return nil
 	}
-	if _, err := r.Digest(); err == nil {
+
+	return fmt.Errorf("%w: invalid non-tag, non-digest reference: %v", errdef.ErrInvalidReference, r.Reference)
```
shizhMSFT, line 179: `ValidateReference()` validates the reference as an empty string or a digest or a tag with smart detection. That is, this function accepts all forms, not just empty strings.

shizhMSFT: Based on that, `Validate()` should not take a `isTag` parameter as it will auto detect the reference type.

registry/reference.go
```diff
@@ -164,13 +175,26 @@ func (r Reference) ValidateReference() error {
...
+
+// ValidateReferenceAsTag validates the reference as a tag.
+func (r Reference) ValidateReferenceAsTag() error {
+	if tagRegexp.MatchString(r.Reference) {
 		return nil
 	}
-	if !tagRegexp.MatchString(r.Reference) {
-		return fmt.Errorf("%w: invalid tag", errdef.ErrInvalidReference)
+
+	return fmt.Errorf("%w: invalid tag", errdef.ErrInvalidReference)
```
shizhMSFT, line 188: It is more readable and maintainable to have code blocks

```go
if err := doStuff(); err != nil {
    // handle error
}
return nil
```

over

```go
if err := doStuff(); err == nil {
    return nil
}

// handle error
```

registry/reference.go
```diff
@@ -164,13 +175,26 @@ func (r Reference) ValidateReference() error {
...
+}
+
+// ValidateReferenceAsDigest validates the reference as a digest.
+func (r Reference) ValidateReferenceAsDigest() (err error) {
+	if _, err = r.Digest(); err == nil {
+		return nil
 	}
-	return nil
+
+	return fmt.Errorf("%w: invalid digest; %v", errdef.ErrInvalidReference, err)
```
shizhMSFT, line 197: same here

registry/remote/repository.go
```diff
@@ -193,28 +193,38 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
...
+		if index := strings.IndexByte(reference, '@'); index != -1 {
+			// drop tag since the digest is present
+			ref.Reference = reference[index+1:]
+			err = ref.ValidateReferenceAsDigest()
+		} else if index = strings.IndexByte(reference, ':'); index != -1 {
+			// no @ but :, so it can only be a digest
+			err = ref.ValidateReferenceAsDigest()
+		} else {
+			err = ref.ValidateReferenceAsTag()
+		}
```
shizhMSFT, line 215: Simplifying the code since `ValidateReference()` will auto detect the reference type.
```suggestion
		// reference is not a FQDN
		if index := strings.IndexByte(reference, '@'); index != -1 {
			// drop tag since the digest is present
			ref.Reference = reference[index+1:]
			err = ref.ValidateReferenceAsDigest()
		} else {
			err = ref.ValidateReference()
		}
```

shizhMSFT (changes requested):
Let's try not to have any breaking changes unless it is necessary with justifications.

registry/remote/repository.go
```diff
@@ -193,28 +193,38 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
...
+
+		if err != nil {
+			return registry.Reference{}, fmt.Errorf("%w: invalid reference; %v", errdef.ErrInvalidReference, err)
```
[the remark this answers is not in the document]
shizhMSFT: Actually, we should do
```suggestion
			return registry.Reference{}, err
```
as we don't to do double wrapping.

errdef/errors.go
nima: `ErrInvalidReferenceTerminal` has been removed.

registry/reference.go
nima: This has been reverted.

registry/reference.go
nima: I'm a bit confused on how you're proposing we deal with the bug found in that case, maybe you can help me understand:

Given an input of a form such as this for example: `registry.example.com:5000/hello-world:<valid-sha256>`

In ParseReference (the function in `registry/reference.go`):

1. First we break it up on the `/`, and get `registry` on the LHS and `path` on the RHS.
2. We look for `@` in `path`; not finding one, we deem that it cannot be valid form A or B.
3. We look for a `:` in path, and finding (at least) one, we deem that it must be a *tag*, valid form C.
4. Then we call `Validate()`, without an `isTag` as you've suggested.
5. Auto-detection kicks in, and finds that `<valid-sha256>` is indeed a valid *digest*, because the information about it having been deemed a *tag* earlier is by this stage lost.

registry/reference.go
nima: I didn't know that, I'll make the change.

registry/reference.go
nima: I've amended all occurrences of this pattern that I introduced.

registry/remote/repository.go
```diff
@@ -193,28 +193,38 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
...
-	ref, err := registry.ParseReference(reference)
-	if err != nil {
-		// reference is not a FQDN
-		if index := strings.IndexByte(reference, '@'); index != -1 {
-			// drop tag since the digest is present
-			reference = reference[index+1:]
-		}
+func (r *Repository) ParseReference(reference string) (ref registry.Reference, err error) {
+	ref, err = registry.ParseReference(reference)
+	if err != nil && len(reference) > 0 {
```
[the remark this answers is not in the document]
nima: Right; fixed.

registry/remote/repository.go
nima: See my earlier comment.  If `ValidateReference` is not given the boolean `isTag`, it will effectively suffer from amnesia.  When it doesn't know *how* the call was made (that there was no `@`), it cannot make the call that what it is about to assess *must not be a digest*.  It then looks at it, decides that it can't be a *tag*, but that it can be a *digest*.  Something akin to `TOCTOU` (https://cwe.mitre.org/data/definitions/367.html)

registry/remote/repository.go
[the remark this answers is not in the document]
nima: Ah, nice; fixed.

registry/reference.go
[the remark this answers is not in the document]
nima: This wasn't a type; it's of the general form:

`<errorClass>: <description>`, where description contains usual grammatical syntax, which includes the semicolon (join ideas that are related and equal in grammatical structure; such as this very sentence :)

registry/reference.go
shizhMSFT: I think the following two things are mixed:

1. Parsing and validating a reference string like `registry.example.com:5000/hello-world:latest`.
2. Validating a `Reference` object.

For `#2`, an example usage is
```go
ref := registry.Reference{
    Registry:   "localhost:5000",
    Repository: "hello-world",
    Reference:  "latest",
}
err := ref.Validate()
check(err)
```
Unlike `#1`, there is no symbols like `@`, `:` in `#2`.

shizhMSFT: In other words, `Validate()` is not meant to be invoked in `ParseReference` but be used in other scenarios like `#2`.

registry/reference.go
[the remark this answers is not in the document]
shizhMSFT: Do we have a doc for `;`? If yes, please provide a link for that. Usually, `:` means a chain of errors where `;` means multiple unrelated errors.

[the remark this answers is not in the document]
nima: Oh I see, you're saying that in the context of errors, `;` already has a special meaning.  I wasn't aware of that; will change the error so that it doesn't rely in a semicolon.

registry/remote/repository.go
shizhMSFT: Since `ValidateReference` is not invoked in `ParseReference` as mentioned in my earlier comment, there is no TOCTOU issue.

registry/reference.go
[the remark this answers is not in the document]
nima: Also the wrapping here was unnecessary; now just `return registry.Reference{}, err`.  Fixed.

registry/reference.go
nima: Hmm, right now I get why I was confused.  But it seems that the only caller (in this package) that calls `Validate` was `ParseReference`.  I guess we're keeping it for callers outside the library.  I'll make this change and send a new PR.

---
I had written this but not posted, then received and read your response; the the following is just for the record, but you can ignore it.

> Take the example where the ParseReference is called with the following input:
> 
> `registry.example.com:5000/hello-world:sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
> 
> This function checks for `@`, and does not find one, it would hence rule out *digest* as a possibility.  Subsequently, it discovers a `:`, and immediately asserts that this can only be a *tag*, leaving the deep verification to the `Validate` call.  So, it has determined that this token can only be a `tag` -- **TOC**.
> 
> Next, it will call the proposed `Validate`, the one without the boolean `isTag` in the function prototype.  This function inspects what it has available to it, which is:
> 
> ```
> registry.Reference{
>     Registry:   "localhost:5000",
>     Repository: "hello-world",
>     Reference:  "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
> }
> ```
> 
> Auto-detection at this point can only make the following assertions: a) that this is not a *tag*, but a *digest*, and b) that the digest is valid -- **TOU**.
> 
> So either we have to perform complete validation within `ParseReference` while we still have access to `isTag`, or we need to forward it.

registry/remote/repository.go
```diff
@@ -193,28 +193,40 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
...
+			err = ref.ValidateReferenceAsTag()
+		}
+
+		if err != nil {
 			return registry.Reference{}, err
 		}
 	} else if ref.Registry != r.Reference.Registry || ref.Repository != r.Reference.Repository {
-		return registry.Reference{}, fmt.Errorf("%w %q: expect %q", errdef.ErrInvalidReference, ref, r.Reference)
+		return registry.Reference{}, fmt.Errorf(
+			"%w: mismatch between received `%q` and expected `%q`",
```
shizhMSFT, line 221: `%q` is already double quoted.
```suggestion
			"%w: mismatch between received %q and expected %q",
```

registry/reference.go
```diff
@@ -156,20 +187,25 @@ func (r Reference) ValidateRepository() error {
 	if !repositoryRegexp.MatchString(r.Repository) {
 		return fmt.Errorf("%w: invalid repository", errdef.ErrInvalidReference)
 	}
+
 	return nil
 }
 
-// ValidateReference validates the reference.
-func (r Reference) ValidateReference() error {
```
shizhMSFT, line 163: Can we keep `ValidateReference()`, which is basically `ValidateReferenceAsTagOrDigestOrEmptyString()`?

shizhMSFT: That is,
```
// ValidateReference validates the reference where the reference can be a tag, a digest, or an empty string.
```

shizhMSFT: We also need to revisit the documentation of `registry.Reference` (#332).

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
 	}
 	registry, path := parts[0], parts[1]
 
+	var isTag bool
 	var repository string
 	var reference string
 	if index := strings.Index(path, "@"); index != -1 {
 		// `digest` found; Valid Form A (if not B)
+		isTag = false
```
shizhMSFT, line 106: nit: This line is redundant since `isTag` is default to `false` at line 101.

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
...
 		// `tag` found; Valid Form C
+		isTag = true
 		repository = path[:index]
 		reference = path[index+1:]
 	} else {
 		// empty `reference`; Valid Form D
 		repository = path
 	}
-	res := Reference{
+	var ref = Reference{
```
shizhMSFT, line 124: nit: Let's do it per golang conversion.
```suggestion
	ref := Reference{
```

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
...
 		Registry:   registry,
 		Repository: repository,
 		Reference:  reference,
 	}
-	if err := res.Validate(); err != nil {
+
+	var err error
+
+	if err = ref.ValidateRegistry(); err != nil {
```
shizhMSFT, line 132: It is better to have
```suggestion
	if err := ref.ValidateRegistry(); err != nil {
```

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
...
+
+	if err = ref.ValidateRepository(); err != nil {
+		return Reference{}, err
+	}
+
+	if len(ref.Reference) > 0 && isTag {
+		err = ref.ValidateReferenceAsTag()
+	} else if len(ref.Reference) > 0 {
+		err = ref.ValidateReferenceAsDigest()
+	}
```
shizhMSFT, line 144: The logic can be simplified as follows, and it is clearer.

```suggestion
	if len(ref.Reference) > 0 {
		return ref, nil
	}
	
	if isTag {
		err = ref.ValidateReferenceAsTag()
	} else {
		err = ref.ValidateReferenceAsDigest()
	}
```

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
...
+	if len(r.Reference) == 0 {
+		return nil
+	}
+
+	if err := r.ValidateReferenceAsTag(); err != nil {
+		return err
+	}
+
+	return r.ValidateReferenceAsDigest()
+}
```
shizhMSFT, line 165: Could you move this function code block back to its original position around line 198?

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
...
 }
 
 // ValidateRegistry validates the registry.
 func (r Reference) ValidateRegistry() error {
-	uri, err := url.ParseRequestURI("dummy://" + r.Registry)
-	if err != nil || uri.Host != r.Registry {
+	if uri, err := url.ParseRequestURI("dummy://" + r.Registry); err != nil || uri.Host != r.Registry {
 		return fmt.Errorf("%w: invalid registry", errdef.ErrInvalidReference)
 	}
+
```
shizhMSFT, line 186: nit: remove empty line.

registry/reference.go
```diff
@@ -156,20 +192,25 @@ func (r Reference) ValidateRepository() error {
 	if !repositoryRegexp.MatchString(r.Repository) {
 		return fmt.Errorf("%w: invalid repository", errdef.ErrInvalidReference)
 	}
+
```
shizhMSFT, line 195: nit: remove empty line.

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
...
 		return err
 	}
-	err = r.ValidateRepository()
-	if err != nil {
+
+	if err := r.ValidateRepository(); err != nil {
 		return err
 	}
+
 	return r.ValidateReference()
```
shizhMSFT, line 178: nit: remove empty lines.

registry/reference.go
```diff
@@ -156,20 +192,25 @@ func (r Reference) ValidateRepository() error {
...
-	}
-	if _, err := r.Digest(); err == nil {
-		return nil
-	}
+// ValidateReferenceAsTag validates the reference as a tag.
+func (r Reference) ValidateReferenceAsTag() error {
 	if !tagRegexp.MatchString(r.Reference) {
 		return fmt.Errorf("%w: invalid tag", errdef.ErrInvalidReference)
 	}
+
```
shizhMSFT, line 204: nit: remove empty line.

registry/reference.go
```diff
@@ -156,20 +192,25 @@ func (r Reference) ValidateRepository() error {
...
+	return nil
+}
+
+// ValidateReferenceAsDigest validates the reference as a digest.
+func (r Reference) ValidateReferenceAsDigest() error {
+	if _, err := r.Digest(); err != nil {
+		return fmt.Errorf("%w: invalid digest; %v", errdef.ErrInvalidReference, err)
+	}
+
```
shizhMSFT, line 213: nit: remove empty line.

registry/reference.go
```diff
@@ -81,12 +81,12 @@ type Reference struct {
...
-//	<=== REPOSITORY ===================================> |    - Valid Form D
+//	<--- path --------------------------------------------> |  - Decode `path`
+//	<=== REPOSITORY ===> <--- reference ------------------> |    - Decode `reference`
+//	<=== REPOSITORY ===> @ <=================== digest ===> |      - Valid Form A
+//	<=== REPOSITORY ===> : <!!! TAG !!!> @ <=== digest ===> |      - Valid Form B (tag is dropped)
+//	<=== REPOSITORY ===> : <=== TAG ======================> |      - Valid Form C
+//	<=== REPOSITORY ======================================> |    - Valid Form D
 //
 // Note: In the case of Valid Form B, TAG is dropped without any validation or
 // further consideration.
```
shizhMSFT, line 92: We should add more tests cases in the unit tests for `ParseReference()`.

registry/reference.go
```diff
@@ -98,56 +98,92 @@ func ParseReference(artifact string) (Reference, error) {
...
+func (r Reference) ValidateReference() error {
+	if len(r.Reference) == 0 {
+		return nil
+	}
+
+	if err := r.ValidateReferenceAsTag(); err != nil {
+		return err
+	}
+
+	return r.ValidateReferenceAsDigest()
```
shizhMSFT, line 164: Let's follow the original behavior. That is, try validate as digest first, and then validate as tag for fallback.

registry/remote/repository.go
shizhMSFT: With better understanding in `reference.go`, you may consider above suggestion again.

shizhMSFT (changes requested):

registry/remote/repository.go
nima: Fixed.

registry/reference.go
nima: Fixed (with the exclusion of the #332 reference; I'll look at that after this issue has been resolved).

registry/reference.go
nima: I get that it's redundant, but being explicit helps with readability (explicit communication of intent, vs accidental); that's why it's there.  Let me know if you think it's better removed though.

registry/reference.go
nima: Fixed.

registry/reference.go
nima: Fixed.

registry/reference.go
nima: Moved.

registry/reference.go
nima: Removed.

registry/reference.go
nima: Removed.

registry/reference.go
nima: Ah, didn't pay attention to the order; good call -- fixed.

registry/reference.go
nima: Added more tests.

registry/reference.go
```diff
@@ -98,54 +98,75 @@ func ParseReference(artifact string) (Reference, error) {
...
+
+	if err := ref.ValidateRepository(); err != nil {
+		return Reference{}, err
+	}
+
+	if len(ref.Reference) == 0 {
+		return ref, nil
+	}
+
+	var validator = ref.ValidateReferenceAsDigest
```
shizhMSFT, line 142: nit:
```suggestion
	validator := ref.ValidateReferenceAsDigest
```

registry/reference.go
```diff
@@ -159,20 +180,36 @@ func (r Reference) ValidateRepository() error {
...
+// ValidateReference where the reference is first tried as an ampty string, then
+// as a digest, and if that fails, as a tag.
+func (r Reference) ValidateReference() error {
+	if len(r.Reference) == 0 {
+		return nil
+	}
+
+	if err := r.ValidateReferenceAsDigest(); err != nil {
+		return err
+	}
```
shizhMSFT, line 208:
```suggestion
	if err := r.ValidateReferenceAsDigest(); err == nil {
		return nil
	}
```

registry/remote/repository.go
```diff
@@ -194,27 +194,43 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
...
-	if err != nil {
-		// reference is not a FQDN
-		if index := strings.IndexByte(reference, '@'); index != -1 {
-			// drop tag since the digest is present
-			reference = reference[index+1:]
-		}
+	var err error
+	var ref registry.Reference
+
+	if ref, err = registry.ParseReference(reference); err != nil {
```
shizhMSFT, line 200: The original code is much simpler.
```suggestion
	ref, err := registry.ParseReference(reference)
	if err != nil {
```

registry/reference.go
```diff
@@ -159,20 +180,36 @@ func (r Reference) ValidateRepository() error {
...
+
+	return r.ValidateReferenceAsTag()
```
shizhMSFT, line 210: Inspired by your code change in `repository.go`, we can have a smarter implementation of `ValidateReference()`.

```suggestion
	if index = strings.IndexByte(reference, ':'); index != -1 {
		return r.ValidateReferenceAsDigest()
	}
	return r.ValidateReferenceAsTag()
```

registry/remote/repository.go
```diff
@@ -194,27 +194,43 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
...
+			// `@` implies *digest*, so drop the *tag* (irrespective of what it is).
+			ref.Reference = reference[index+1:]
+			err = ref.ValidateReferenceAsDigest()
+		} else if index = strings.IndexByte(reference, ':'); index != -1 {
+			// no `@` but `:` again implies that it can only (legally) be a *digest*
+			err = ref.ValidateReferenceAsDigest()
+		} else {
+			// finally, the only possibility now is that it's a *tag*
+			err = ref.ValidateReferenceAsTag()
+		}
```
shizhMSFT, line 218: Can be simplified as follows.
```suggestion
		} else {
			err = ref.ValidateReference()
		}
```

The `len(ref.Reference) == 0` case will be caught be line 230.

registry/remote/repository.go
```diff
@@ -1070,15 +1086,15 @@ func verifyContentDigest(resp *http.Response, expected digest.Digest) error {
 	contentDigest, err := digest.Parse(digestStr)
 	if err != nil {
 		return fmt.Errorf(
-			"%s %q: invalid response header: `%s: %s`",
+			"%s %q: invalid response header `%s=%q`",
```
shizhMSFT, line 1089: Let's revert this change since `%s: %s` is the original form in the HTTP protocol.
```suggestion
			"%s %q: invalid response header: `%s: %s`",
```

registry/remote/repository.go
```diff
@@ -987,7 +1003,7 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref registry.Ref
 	if serverHeaderDigestStr := resp.Header.Get(dockerContentDigestHeader); serverHeaderDigestStr != "" {
 		if serverHeaderDigest, err = digest.Parse(serverHeaderDigestStr); err != nil {
 			return ocispec.Descriptor{}, fmt.Errorf(
-				"%s %q: invalid response header value `%s`: `%s`; %w",
+				"%s %q: invalid response header value %s=%q; %w",
```
shizhMSFT, line 1006: `%s: %s` is the original form in the HTTP protocol.
```suggestion
				"%s %q: invalid response header value `%s: %s`; %w",
```

shizhMSFT (changes requested):

registry/reference.go
nima: I'm confused now... isn't this exactly the opposite of the earlier suggestion [here](https://github.com/oras-project/oras-go/pull/326#discussion_r975953379)?

registry/reference.go
nima: Nice, fixed :)

registry/remote/repository.go
nima: Nice!

registry/remote/repository.go
nima: Fixed.

shizhMSFT (approved):
LGTM

PR #327, opened 2022-09-20
fix: add nil checks for some functions

### LINKED ISSUE #318: Missing `nil` checks
For functions that take a pointer parameter, we should check if the parameter is a `nil`. If so, return early or throw an error, depending on the scenario.

For instance,

- [oras.WithTargetPlatform](https://github.com/oras-project/oras-go/blob/main/copy.go#L66)
- [oras.FilterAnnotation](https://github.com/oras-project/oras-go/blob/main/extendedcopy.go#L183)
- [oras.FilterArtifactType](https://github.com/oras-project/oras-go/blob/main/extendedcopy.go#L238)

jasminetMSFT:
Add `nil` checks for functions `oras.WithTargetPlatform`, `oras.FilterAnnotation` and `oras.FilterArtifactType`.

Resolves #318 
Signed-off-by: JasmineTang <jasminetang@microsoft.com>


copy_test.go
```diff
@@ -1058,6 +1058,14 @@ func TestCopy_WithTargetPlatformOptions(t *testing.T) {
...
 		t.Fatalf("Copy() error = %v, wantErr %v", err, expected)
 	}
+
+	// test copy with no platform filter
+	// opts.MapRoot should be nil
+	opts = oras.CopyOptions{}
+	opts.WithTargetPlatform(nil)
+	if opts.MapRoot != nil {
+		t.Fatal("opts.MapRoot not equal to nil when platform is not provided")
+	}
```
Wwwsylvia, line 1068: We should also test the case where there is a custom `opts.MapRoot`.

extendedcopy.go
```diff
@@ -181,6 +181,9 @@ func findRoots(ctx context.Context, storage content.ReadOnlyGraphStorage, node o
 // when using both FilterArtifactType and FilterAnnotation, it's recommended to call
 // FilterArtifactType first.
 func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp.Regexp) {
+	if key == "" {
```
shizhMSFT, line 184: Should we also check `regex`? BTW, is empty key a valid input?

extendedcopy_test.go
```diff
@@ -675,6 +675,120 @@ func TestExtendedCopyGraph_FilterAnnotationWithRegex(t *testing.T) {
...
+	opts := oras.ExtendedCopyGraphOptions{}
+	exp := "black."
+	regex := regexp.MustCompile(exp)
+	opts.FilterAnnotation("", regex)
+	if opts.FindPredecessors != nil {
+		t.Fatal("FindPredecessors not nil!")
+	}
+}
+
+func TestExtendedCopyGraph_FilterAnnotationWithNoRegex(t *testing.T) {
```
Wwwsylvia, line 725: Maybe we can merge `TestExtendedCopyGraph_FilterAnnotationWithNoKey` and `TestExtendedCopyGraph_FilterAnnotationWithNoRegex` into `TestExtendedCopyGraph_FilterAnnotation`?

copy.go
```diff
@@ -64,6 +64,9 @@ type CopyOptions struct {
 // matching manifest if exists, otherwise ErrNotFound will be returned.
 // - Otherwise ErrUnsupported will be returned.
 func (opts *CopyOptions) WithTargetPlatform(p *ocispec.Platform) {
+	if p == nil {
```
Wwwsylvia, line 67: We need to document the nil behavior.

Wwwsylvia (changes requested):

extendedcopy.go
```diff
@@ -236,6 +239,9 @@ func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp
 // when using both FilterArtifactType and FilterAnnotation, it's recommended to call
 // FilterArtifactType first.
 func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
+	if regex == nil {
```
wangxiaoxuan273, line 242: In the case of `regex == nil`, why does `FilterAnnotation` proceeds filtering, while `FilterArtifactType` just returns ? Should they have the same behavior?

extendedcopy_test.go
```diff
@@ -889,6 +1003,54 @@ func TestExtendedCopyGraph_FilterArtifactTypeWithRegex(t *testing.T) {
...
+	src := memory.New()
+	for i := range blobs {
+		err := src.Push(ctx, descs[i], bytes.NewReader(blobs[i]))
+		if err != nil {
+			t.Errorf("failed to push test content to src: %d: %v", i, err)
+		}
+	}
+
+	opts := oras.ExtendedCopyGraphOptions{}
+	opts.FilterArtifactType(nil)
```
wangxiaoxuan273, line 1048: I think we can also test the cases of calling `FilterArtifactType` and `FilterAnnotation` **multiple times**, where one or some of the calls have nil arguments.

extendedcopy.go
```diff
@@ -236,6 +238,9 @@ func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp
...
```
shizhMSFT, line 241: Let's document this `regex == nil` behavior.

extendedcopy.go
```diff
@@ -223,8 +223,10 @@ func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp
 					}
 				}
 			}
-			if regex.MatchString(p.Annotations[key]) {
-				filtered = append(filtered, p)
+			if _, ok := p.Annotations[key]; ok {
+				if regex == nil || regex.MatchString(p.Annotations[key]) {
+					filtered = append(filtered, p)
+				}
 			}
```
shizhMSFT, line 230:
```suggestion
			if value, ok := p.Annotations[key]; ok && (regex == nil || regex.MatchString(value)) {
				filtered = append(filtered, p)
			}
```

shizhMSFT (changes requested):

shizhMSFT (approved):
LGTM

copy_test.go
```diff
@@ -1058,6 +1058,32 @@ func TestCopy_WithTargetPlatformOptions(t *testing.T) {
...
+		MapRoot: func(ctx context.Context, src content.ReadOnlyStorage, root ocispec.Descriptor) (ocispec.Descriptor, error) {
+			if root.MediaType == "test" {
+				return root, nil
+			} else {
+				return ocispec.Descriptor{}, errdef.ErrNotFound
+			}
+		},
+		CopyGraphOptions: oras.DefaultCopyGraphOptions,
+	}
+
```
Wwwsylvia, line 1082: Did you forget to call `opts.WithTargetPlatform(nil)`?

extendedcopy.go
```diff
@@ -223,7 +223,7 @@ func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp
 					}
 				}
 			}
-			if regex.MatchString(p.Annotations[key]) {
+			if value, ok := p.Annotations[key]; ok && (regex == nil || regex.MatchString(value)) {
```
Wwwsylvia, line 226: We need to document this behavior.

copy.go
```diff
@@ -58,12 +58,16 @@ type CopyOptions struct {
 // WithTargetPlatform configures opts.MapRoot to select the manifest whose
 // platform matches the given platform. When MapRoot is provided, the platform
 // selection will be applied on the mapped root node.
+// - If the given platform is nil, no operations will be done.
```
Wwwsylvia, line 61: nit: how about "no platform selection will be applied"?

extendedcopy.go
```diff
@@ -232,10 +232,14 @@ func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp
 }
 
 // FilterArtifactType will configure opts.FindPredecessors to filter the predecessors
-// whose artifact type matches a given regex pattern. For performance consideration,
-// when using both FilterArtifactType and FilterAnnotation, it's recommended to call
+// whose artifact type matches a given regex pattern. When the regex pattern is nil,
+// no operations will be done. For performance consideration, when using both
```
Wwwsylvia, line 236: nit: how about "no artifact type filter will be applied"?

Wwwsylvia (approved):
LGTM

extendedcopy.go
```diff
@@ -177,9 +177,15 @@ func findRoots(ctx context.Context, storage content.ReadOnlyGraphStorage, node o
...
-// whose annotation matches a given regex pattern. For performance consideration,
-// when using both FilterArtifactType and FilterAnnotation, it's recommended to call
-// FilterArtifactType first.
+// whose annotation matches a given regex pattern. For each predecessor,
+// - If key is available in the predecessor's annotation and regex is not nil,
+// return the predecessor if the annotation value matches the regex.
+// - If key is available in the predecessor's annotation and regex is nil,
+// return the predecessor.
+// - If key is unavailable in the predecessor's annotation, the predecessor
+// will not be returned.
```
shizhMSFT, line 186: A predecessor is not filtered out if the key is in its annotation and matches the regex if present.

shizhMSFT (approved):
LGTM

PR #335, opened 2022-09-28
refactor: optimize extended copy filters

### LINKED ISSUE #310: Improve performance of filters of `oras.ExtendedCopy()` caused by information loss due to descriptor conversions

Wwwsylvia:
According to [Open Container Initiative Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers), the `Referrers API` must include the artifact type and annotations of the referrers in the returned descriptor list. 

> The descriptors MUST include an artifactType field that is set to the value of artifactType for an artifact manifest if present, or the configuration descriptor's mediaType for an image manifest. The descriptors MUST include annotations from the image or artifact manifest.

However, the `Predecessors` function implemented by file, OCI, and memory store is not guaranteed to include the artifact type and annotations in the returned descriptors.

This PR is migrated from https://github.com/oci-playground/oras-go/pull/3.

Resolves #310 

### DIFF
--- a/content/graph.go
+++ b/content/graph.go
@@ -50,13 +50,22 @@ type ReadOnlyGraphStorage interface {
 // In other words, returns the "children" of the current descriptor.
 func Successors(ctx context.Context, fetcher Fetcher, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
 	switch node.MediaType {
-	case docker.MediaTypeManifest, ocispec.MediaTypeImageManifest:
+	case docker.MediaTypeManifest:
 		content, err := FetchAll(ctx, fetcher, node)
 		if err != nil {
 			return nil, err
 		}
-
-		// docker manifest and oci manifest are equivalent for successors.
+		// OCI manifest schema can be used to marshal docker manifest
+		var manifest ocispec.Manifest
+		if err := json.Unmarshal(content, &manifest); err != nil {
+			return nil, err
+		}
+		return append([]ocispec.Descriptor{manifest.Config}, manifest.Layers...), nil
+	case ocispec.MediaTypeImageManifest:
+		content, err := FetchAll(ctx, fetcher, node)
+		if err != nil {
+			return nil, err
+		}
 		var manifest ocispec.Manifest
 		if err := json.Unmarshal(content, &manifest); err != nil {
 			return nil, err
--- a/content/graph.go
+++ b/content/graph.go
@@ -62,10 +62,10 @@ func Successors(ctx context.Context, fetcher Fetcher, node ocispec.Descriptor) (
 			return nil, err
 		}
 		var nodes []ocispec.Descriptor
-		nodes = append(nodes, manifest.Config)
 		if manifest.Subject != nil {
 			nodes = append(nodes, *manifest.Subject)
 		}
+		nodes = append(nodes, manifest.Config)
 		return append(nodes, manifest.Layers...), nil
 	case docker.MediaTypeManifestList, ocispec.MediaTypeImageIndex:
 		content, err := FetchAll(ctx, fetcher, node)
--- a/content/graph.go
+++ b/content/graph.go
@@ -61,7 +61,12 @@ func Successors(ctx context.Context, fetcher Fetcher, node ocispec.Descriptor) (
 		if err := json.Unmarshal(content, &manifest); err != nil {
 			return nil, err
 		}
-		return append([]ocispec.Descriptor{manifest.Config}, manifest.Layers...), nil
+		var nodes []ocispec.Descriptor
+		nodes = append(nodes, manifest.Config)
+		if manifest.Subject != nil {
+			nodes = append(nodes, *manifest.Subject)
+		}
+		return append(nodes, manifest.Layers...), nil
 	case docker.MediaTypeManifestList, ocispec.MediaTypeImageIndex:
 		content, err := FetchAll(ctx, fetcher, node)
 		if err != nil {
--- a/extendedcopy.go
+++ b/extendedcopy.go
@@ -175,31 +175,30 @@ func findRoots(ctx context.Context, storage content.ReadOnlyGraphStorage, node o
 	return roots, nil
 }
 
-// FilterAnnotation will configure opts.FindPredecessors to filter the
-// predecessors whose annotation matches a given regex pattern. A predecessor is
-// kept if the key is in its annotation and matches the regex if present.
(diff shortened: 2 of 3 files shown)


Wwwsylvia:
This PR is migrated from https://github.com/oci-playground/oras-go/pull/3.

extendedcopy.go
```diff
@@ -182,52 +181,64 @@ func findRoots(ctx context.Context, storage content.ReadOnlyGraphStorage, node o
...
-		for _, p := range predecessors {
-			if p.Annotations == nil {
-				switch p.MediaType {
-				case docker.MediaTypeManifest, ocispec.MediaTypeImageManifest,
-					docker.MediaTypeManifestList, ocispec.MediaTypeImageIndex,
-					artifactspec.MediaTypeArtifactManifest:
-					if err = func() error {
-						rc, err := src.Fetch(ctx, p)
+
+		if !supportReferrers {
```
shizhMSFT, line 219: `supportReferrers == true` does not imply `predecessors` comes from the referrer API. They may come from user specified `opts.FindPredecessors`.

extendedcopy.go
```diff
@@ -241,62 +252,90 @@ func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
...
 	}
-	return predecessors, nil
+	defer rc.Close()
+	var manifest struct {
+		Annotations map[string]string `json:"annotations"`
+	}
+	if err := json.NewDecoder(rc).Decode(&manifest); err != nil {
+		return nil, err
+	}
+	return manifest.Annotations, nil
```
shizhMSFT, line 326:
```suggestion
	if manifest.Annotations == nil {
		return make(map[string]string), nil
	}
	return manifest.Annotations, nil
```

extendedcopy.go
```diff
@@ -182,52 +181,64 @@ func findRoots(ctx context.Context, storage content.ReadOnlyGraphStorage, node o
...
 					}
 				}
 			}
-			if value, ok := p.Annotations[key]; ok && (regex == nil || regex.MatchString(value)) {
-				filtered = append(filtered, p)
-			}
 		}
+		var filtered []ocispec.Descriptor
+		filtered = filter(filtered, predecessors)
 		return filtered, nil
```
shizhMSFT, line 242: nit:
```suggestion
		return filter(nil, predecessors), nil
```

extendedcopy.go
```diff
@@ -241,62 +252,90 @@ func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
...
-			if p.MediaType == artifactspec.MediaTypeArtifactManifest {
-				if err = func() error {
-					rc, err := src.Fetch(ctx, p)
+
+		if !supportReferrers {
+			// For src that does not support Referrers, Predecessors is used.
+			// However, the descriptors returned by Predecessors may not include
+			// the artifact type of the corresponding manifests.
+			for i, p := range predecessors {
+				if p.MediaType == ocispec.MediaTypeArtifactManifest && p.ArtifactType == "" {
```
shizhMSFT, line 295: We might need to review the spec. For OCI image manifest, the config media type is used as artifact type.

shizhMSFT (changes requested):

extendedcopy.go
Wwwsylvia: updated.

extendedcopy.go
```diff
@@ -182,56 +181,89 @@ func findRoots(ctx context.Context, storage content.ReadOnlyGraphStorage, node o
 // For performance consideration, when using both FilterArtifactType and
 // FilterAnnotation, it's recommended to call FilterArtifactType first.
 func (opts *ExtendedCopyGraphOptions) FilterAnnotation(key string, regex *regexp.Regexp) {
+	filter := func(desc ocispec.Descriptor) bool {
+		if value, ok := desc.Annotations[key]; ok && (regex == nil || regex.MatchString(value)) {
+			return true
+		}
+		return false
+	}
```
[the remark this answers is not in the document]
shizhMSFT: We probably also need to update the documentation to be less ambiguous.

content/graph.go
```diff
@@ -61,7 +61,12 @@ func Successors(ctx context.Context, fetcher Fetcher, node ocispec.Descriptor) (
 		if err := json.Unmarshal(content, &manifest); err != nil {
 			return nil, err
 		}
-		return append([]ocispec.Descriptor{manifest.Config}, manifest.Layers...), nil
+		var nodes []ocispec.Descriptor
+		if manifest.Subject != nil {
+			nodes = append(nodes, *manifest.Subject)
+		}
```
shizhMSFT, line 67: `docker.MediaTypeManifest` should not have `manifest.Subject`.

shizhMSFT: That is, the above statement

> // docker manifest and oci manifest are equivalent for successors.

is no longer true.

extendedcopy.go
```diff
@@ -241,14 +273,29 @@ func (opts *ExtendedCopyGraphOptions) FilterArtifactType(regex *regexp.Regexp) {
 	if regex == nil {
 		return
 	}
+	filter := func(desc ocispec.Descriptor) bool {
+		return regex.MatchString(desc.ArtifactType)
+	}
```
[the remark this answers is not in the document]
Wwwsylvia: How about
```go
func (opts *ExtendedCopyGraphOptions) MatchArtifactType(artifactType string) {
```
?

extendedcopy.go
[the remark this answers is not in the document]
Wwwsylvia: updated

content/graph.go
Wwwsylvia: fixed

shizhMSFT (approved):
LGTM

extendedcopy.go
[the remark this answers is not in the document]
Wwwsylvia: We can add this into `rc.5`.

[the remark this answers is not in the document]
shizhMSFT: > select: v. carefully choose as being the best or most suitable.
> match: v. be equal to (something) in quality or strength.

Basically, to select, you need a list of items first :D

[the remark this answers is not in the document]
shizhMSFT: Let's create an issue in `rc.5` to track it.

Wwwsylvia:
Tested against https://github.com/oci-playground/distribution/tree/shizh/artifact.

shizhMSFT (approved):
LGTM

PR #340, opened 2022-10-09
refactor: update implementation of `Referrers` API

### LINKED ISSUE #311: Update implementation of `Referers` API

Wwwsylvia:
Resolves #311

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -339,6 +339,7 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
 	}
 	if !tagSchemaUsed && r.ReferrerListPageSize > 0 {
 		// Referrers tag schema does not support paging
+		// reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#backwards-compatibility
 		q := req.URL.Query()
 		q.Set("n", strconv.Itoa(r.ReferrerListPageSize))
 		req.URL.RawQuery = q.Encode()
@@ -387,6 +388,9 @@ func isReferrersFilterApplied(annotation string, filter string) bool {
 		return true
 	}
 	filters := strings.Split(annotation, ",")
+	if len(filters) < 2 {
+		return false
+	}
 	for _, f := range filters {
 		if f == filter {
 			return true
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -348,6 +348,11 @@ func (r *Repository) Referrers(ctx context.Context, desc ocispec.Descriptor, art
 	for err == nil {
 		url, err = r.referrers(ctx, artifactType, fn, url, tagSchemaUsed)
 	}
+	if err == errdef.ErrNotFound && tagSchemaUsed {
+		// If the tag schema API does not return a valid image index,
+		// consider that there are no referrers to the manifest.
+		return nil
+	}
 	if err != errNoLink {
 		return err
 	}
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -348,15 +348,15 @@ func (r *Repository) Referrers(ctx context.Context, desc ocispec.Descriptor, art
 	for err == nil {
 		url, err = r.referrers(ctx, artifactType, fn, url, tagSchemaUsed)
 	}
-	if err == errdef.ErrNotFound && tagSchemaUsed {
+	if err == errNoLink {
+		return nil
+	}
+	if errors.Is(err, errdef.ErrNotFound) && tagSchemaUsed {
 		// If the tag schema API does not return a valid image index,
 		// consider that there are no referrers to the manifest.
 		return nil
 	}
-	if err != errNoLink {
-		return err
-	}
-	return nil
+	return err
 }
 
 // referrers returns a single page of the manifest descriptors directly
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync/atomic"
 
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -93,6 +94,11 @@ type Repository struct {
 	// list, and referrers list.
 	// If less than or equal to zero, a default (currently 4MiB) is used.
 	MaxMetadataBytes int64
+
+	// referrersUnsupported indicates if Referrers API is unsupported for the
+	// repository.
+	//  0: false (supported), 1: true (unsupported).
+	referrersUnsupported int32
 }
 
 // NewRepository creates a client to the remote repository identified by a
@@ -106,6 +112,18 @@ func NewRepository(reference string) (*Repository, error) {
 	return &Repository{
 		Reference: ref,
 	}, nil
+}
+
+// setReferrersUnsupported sets referrersUnsupported to indicate that
+// Referrers API is unsupported for the repository.
+func (r *Repository) setReferrersUnsupported() {
+	atomic.StoreInt32(&r.referrersUnsupported, 1)
+}
+
+// isReferrersUnsupported returns true if Referrers API is unsupported for the
+// repository, otherwise returns false.
+func (r *Repository) isReferrersUnsupported() bool {
+	return atomic.LoadInt32(&r.referrersUnsupported) == 1
 }
 
 // client returns an HTTP client used to access the remote repository.
@@ -310,8 +328,14 @@ func (r *Repository) Referrers(ctx context.Context, desc ocispec.Descriptor, art
 
 	var err error
 	var tagSchemaUsed bool
-	url, err = r.referrers(ctx, artifactType, fn, url, tagSchemaUsed)
-	if errors.Is(err, errdef.ErrNotFound) {
+	if !r.isReferrersUnsupported() {
+		url, err = r.referrers(ctx, artifactType, fn, url, tagSchemaUsed)
+		if errors.Is(err, errdef.ErrNotFound) {
+			// to save future requests to Referrers API
+			r.setReferrersUnsupported()
+		}
+	}
+	if r.isReferrersUnsupported() {
 		// Referrers API is unavailable, fallback to the referrers tag schema
 		// reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#backwards-compatibility
 		url, err = buildReferrersTagSchemaURL(r.PlainHTTP, ref)
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -24,11 +24,9 @@ import (
 	"io"
 	"mime"
 	"net/http"
-	"regexp"
 	"strconv"
 	"strings"
 
-	"github.com/opencontainers/distribution-spec/specs-go/v1/extensions"
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"oras.land/oras-go/v2/errdef"
@@ -46,10 +44,6 @@ import (
 // See https://docs.docker.com/registry/spec/api/#digest-header
 // See https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull
 const dockerContentDigestHeader = "Docker-Content-Digest"
-
-// referrersApiRegex checks referrers API version.
-// Reference: https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#versioning
-var referrersApiRegex = regexp.MustCompile(`^oras/1\.(0|[1-9]\d*)$`)
 
 // Client is an interface for a HTTP client.
 type Client interface {
@@ -288,17 +282,14 @@ func (r *Repository) tags(ctx context.Context, last string, fn func(tags []strin
 	return parseLink(resp)
 }
 
-// Predecessors returns the descriptors of ORAS Artifact manifests directly
-// referencing the given manifest descriptor.
-// Predecessors internally leverages Referrers, and converts the result ORAS
-// Artifact descriptors to OCI descriptors.
-// Reference: https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md
+// Predecessors returns the descriptors of manifests directly referencing the
+// given manifest descriptor.
+// Predecessors internally leverages Referrers.
+// Reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
 func (r *Repository) Predecessors(ctx context.Context, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
 	var res []ocispec.Descriptor
 	if err := r.Referrers(ctx, desc, "", func(referrers []ocispec.Descriptor) error {
-		for _, referrer := range referrers {
-			res = append(res, referrer)
-		}
+		res = append(res, referrers...)
 		return nil
 	}); err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


Wwwsylvia:
Tested against https://github.com/oci-playground/distribution/tree/shizh/artifact.

registry/remote/repository.go
```diff
@@ -99,6 +94,11 @@ type Repository struct {
 	// list, and referrers list.
 	// If less than or equal to zero, a default (currently 4MiB) is used.
 	MaxMetadataBytes int64
+
+	// referrersUnsupported indicates if Referrers API is unsupported for the
+	// repository.
+	//  0: false (supported), 1: true (unsupported).
+	referrersUnsupported int32
```
shizhMSFT, line 101: `oras-go` should provide the ability to pre-set whether the remote server supports referrers API without access it.

Basically, it can have 3 states: unknown (detection required), known that remote server has support, and known that remote server has no support on referrers API. Let's discuss this part in detail.

wangxiaoxuan273:
LGTM

registry/remote/repository.go
Wwwsylvia: Updated.

registry/remote/repository.go
```diff
@@ -288,57 +323,69 @@ func (r *Repository) tags(ctx context.Context, last string, fn func(tags []strin
...
+		if err == errNoLink {
+			// no more pages
+			r.SetReferrersCapability(true)
+			return nil
+		}
+		if errors.Is(err, errdef.ErrNotFound) && state == referrersStateUnknown {
+			// A 404 returned by Referrers API represents that Referrers API is
+			// not supported. If the repository Referrers state is unknown,
+			// fallback to referrers tag schema.
+			r.SetReferrersCapability(false)
```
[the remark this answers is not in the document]
Wwwsylvia: Do you mean we should allow `SetReferrersCapability()` only when state is unknown?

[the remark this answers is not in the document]
Wwwsylvia: Or should we have something like
```go
NewRepositoryWithReferrersCapability(reference string, capable bool)`?
```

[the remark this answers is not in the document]
Wwwsylvia: Took the first approach.

registry/remote/repository.go
```diff
@@ -47,9 +47,19 @@ import (
...
 
-// referrersApiRegex checks referrers API version.
-// Reference: https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#versioning
-var referrersApiRegex = regexp.MustCompile(`^oras/1\.(0|[1-9]\d*)$`)
+// referrersState represents the state of Referrers API.
+type referrersState = int32
+
+const (
+	// referrersStateUnknown represents an unknown state of Referrers API.
+	referrersStateUnknown referrersState = 0
```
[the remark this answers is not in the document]
Wwwsylvia: updated

registry/remote/url.go
```diff
@@ -87,10 +87,10 @@ func buildRepositoryBlobUploadURL(plainHTTP bool, ref registry.Reference) string
 	return buildRepositoryBaseURL(plainHTTP, ref) + "/blobs/uploads/"
 }
 
-// buildArtifactReferrerURLLegacy builds the URL for accessing the manifest referrers API in artifact spec v1.0.0-draft.1.
-// Format: <scheme>://<registry>/oras/artifacts/v1/<repository>/manifests/<digest>/referrers?artifactType=<artifactType>
-// Reference: https://github.com/oras-project/artifacts-spec/blob/v1.0.0-draft.1/manifest-referrers-api.md
-func buildArtifactReferrerURLLegacy(plainHTTP bool, ref registry.Reference, artifactType string) string {
+// buildReferrersURL builds the URL for querying the Referrers API.
+// Format: <scheme>://<registry>/v2/<repository>/referrers/<digest>?artifactType=<artifactType>
+// Reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
```
shizhMSFT, line 92: Let's use a tagged version instead of `main`.

registry/remote/repository.go
```diff
@@ -114,6 +128,28 @@ func NewRepository(reference string) (*Repository, error) {
...
+// SetReferrersCapabilityOnce sets the capability of Referrers API for the
+// repository. true: capable; false: not capable.
+// SetReferrersCapabilityOnce is valid only when it is called for the first time.
+// - When the capability is set to true, the Referrers() function will always
+// request the Referrers API. Reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
+// - When the capability is set to false, the Referrers() function will always
+// request the Referrers Tag. Reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#referrers-tag-schema
+// - When the capability is not set, the Referrers() function will automatically
+// determine which API to use.
+func (r *Repository) SetReferrersCapabilityOnce(capable bool) {
```
shizhMSFT, line 140: The term `Once` is a bit confusing.
```suggestion
func (r *Repository) SetReferrersCapability(capable bool) {
```

registry/remote/repository.go
```diff
@@ -288,57 +324,69 @@ func (r *Repository) tags(ctx context.Context, last string, fn func(tags []strin
...
 }
 
-// Predecessors returns the descriptors of ORAS Artifact manifests directly
+// Predecessors returns the descriptors of image or artifact manifests directly
 // referencing the given manifest descriptor.
-// Predecessors internally leverages Referrers, and converts the result ORAS
-// Artifact descriptors to OCI descriptors.
-// Reference: https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md
+// Predecessors internally leverages Referrers.
+// Reference: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
```
shizhMSFT, line 330: Let's use a tagged version instead of `main`. Please also apply the update to other docs.

registry/remote/repository.go
```diff
@@ -361,38 +409,92 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#unavailable-referrers-api
+func buildReferrersTag(desc ocispec.Descriptor) string {
+	alg := desc.Digest.Algorithm().String()
+	encoded := desc.Digest.Encoded()
+	return alg + "-" + encoded
+}
+
+// isReferrersFilterApplied checks the filtersApplied annotation to see if
+// filter is in the applied filter list.
+func isReferrersFilterApplied(annotation string, filter string) bool {
```
shizhMSFT, line 479: nit:
```suggestion
func isReferrersFilterApplied(applied string, requested string) bool {
```

shizhMSFT, line 479: It is more readable to pass the annotation map instead of passing the annotation value.

registry/remote/repository.go
```diff
@@ -361,38 +409,92 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
+func (r *Repository) referrersTagSchema(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error {
+	referrersTag := buildReferrersTag(desc)
+	desc, rc, err := r.FetchReference(ctx, referrersTag)
+	if err != nil {
+		if errors.Is(err, errdef.ErrNotFound) {
+			// no referrers to the manifest
+			return nil
+		} else {
+			return err
+		}
```
shizhMSFT, line 447: nit:
```suggestion
		return err
```

registry/remote/repository.go
```diff
@@ -361,38 +409,92 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
+	}
+	defer rc.Close()
+
+	var index ocispec.Index
+	lr := limitReader(rc, r.MaxMetadataBytes)
+	indexJson, err := content.ReadAll(lr, desc)
```
shizhMSFT, line 453: nit:
```suggestion
	indexJSON, err := content.ReadAll(lr, desc)
```

registry/remote/repository.go
```diff
@@ -114,6 +128,28 @@ func NewRepository(reference string) (*Repository, error) {
...
+// SetReferrersCapability is valid only when it is called for the first time.
+// - When the capability is set to true, the Referrers() function will always
+// request the Referrers API. Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
+// - When the capability is set to false, the Referrers() function will always
+// request the Referrers Tag. Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema
+// - When the capability is not set, the Referrers() function will automatically
+// determine which API to use.
+func (r *Repository) SetReferrersCapability(capable bool) {
+	if capable {
+		atomic.CompareAndSwapInt32(&r.referrersState, referrersStateUnknown, referrersStateSupported)
```
shizhMSFT, line 142: The return value of `CompareAndSwapInt32()` is ignored but it should not. We may consider returning an error instead.

registry/remote/repository.go
```diff
@@ -288,57 +324,69 @@ func (r *Repository) tags(ctx context.Context, last string, fn func(tags []strin
...
 
-// referrers returns a single page of the manifest descriptors directly
-// referencing the given manifest descriptor with the next link.
-func (r *Repository) referrers(ctx context.Context, artifactType string, fn func(referrers []ocispec.Descriptor) error, url string, legacyAPI bool) (string, error) {
+// referrers lists a single page of the descriptors of manifests directly
+// referencing the given manifest descriptor. fn is called for a page of
+// referrers result. If artifactType is not empty, only referrers of the same
+// artifact type are fed to fn.
+// referrers returns the link url for the next page.
+func (r *Repository) referrers(ctx context.Context, artifactType string, fn func(referrers []ocispec.Descriptor) error, url string) (string, error) {
```
qweeah, line 389: if `legacyAPI` is removed, should this be a breaking-change (`refactor!`) PR?

qweeah: oh just found that it's private method. Please ignore this.

registry/remote/repository.go
```diff
@@ -361,38 +409,92 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
-	var refs []ocispec.Descriptor
-	if legacyAPI {
-		refs = page.References
-	} else {
-		refs = page.Referrers
+	referrers := index.Manifests
+	if artifactType != "" &&
+		!isReferrersFilterApplied(index.Annotations[ocispec.AnnotationReferrersFiltersApplied], "artifactType") {
+		// perform client side filtering if the filter is not applied on the
+		// server side
```
qweeah, line 422: I guess the comment length limitation should not applied to this line since it will not be converted to any documentation?

registry/remote/repository.go
```diff
@@ -114,6 +128,28 @@ func NewRepository(reference string) (*Repository, error) {
 	}, nil
 }
 
+// SetReferrersCapability sets the capability of Referrers API for the
+// repository. true: capable; false: not capable.
+// SetReferrersCapability is valid only when it is called for the first time.
```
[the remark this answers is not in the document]
Wwwsylvia: No it isn't. Revised the doc.

registry/remote/repository.go
```diff
@@ -288,57 +324,69 @@ func (r *Repository) tags(ctx context.Context, last string, fn func(tags []strin
...
-// the referrers result. If artifactType is not empty, only referrers of the
-// same artifact type are fed to fn.
-// Reference: https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md
+// Referrers lists the descriptors of image or artifact manifests directly
+// referencing the given manifest descriptor.
+// fn is called for each page of the referrers result.
+// If artifactType is not empty, only referrers of the same artifact type are
+// fed to fn.
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers
 func (r *Repository) Referrers(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error {
```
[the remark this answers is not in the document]
Wwwsylvia: This is nice!

registry/remote/repository.go
Wwwsylvia: Added an error.

registry/remote/repository.go
Wwwsylvia: Merged into one line.

registry/remote/repository.go
```diff
@@ -114,6 +132,37 @@ func NewRepository(reference string) (*Repository, error) {
...
+		swapped = atomic.CompareAndSwapInt32(&r.referrersState, referrersStateUnknown, referrersStateSupported)
+	} else {
+		swapped = atomic.CompareAndSwapInt32(&r.referrersState, referrersStateUnknown, referrersStateUnsupported)
+	}
+	if !swapped {
+		return fmt.Errorf("%w: current capability = %v",
+			ErrReferrersCapabilityAlreadySet,
+			r.loadReferrersState() == referrersStateSupported)
+	}
+	return nil
```
shizhMSFT, line 158: How about
```suggestion
	state := referrersStateUnsupported
	if capable {
		state = referrersStateSupported
	}
	if swapped := atomic.CompareAndSwapInt32(&r.referrersState, referrersStateUnknown, state); !swapped {
		if fact := r.loadReferrersState(); fact != state {
			return errors.New("capability cannot be changed once set")
		}
	}
	return nil
```

registry/remote/repository.go
```diff
@@ -361,38 +436,86 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
 		return "", fmt.Errorf("%s %q: failed to decode response: %w", resp.Request.Method, resp.Request.URL, err)
 	}
-	var refs []ocispec.Descriptor
-	if legacyAPI {
-		refs = page.References
-	} else {
-		refs = page.Referrers
+	referrers := index.Manifests
+	if artifactType != "" &&
+		!isReferrersFilterApplied(index.Annotations, "artifactType") {
```
shizhMSFT, line 447: nit: it is actually more readable if it is one line. It won't be longer than line 443.
```suggestion
	if artifactType != "" && !isReferrersFilterApplied(index.Annotations, "artifactType") {
```

registry/remote/repository.go
```diff
@@ -361,38 +436,86 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
 	return parseLink(resp)
 }
 
+// referrersByTagSchema lists the descriptors of manifests directly referencing
+// the given manifest descriptor by requesting referrers tag.
+// fn is called for the referrers result. If artifactType is not empty,
+// only referrers of the same artifact type are fed to fn.
+// reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#backwards-compatibility
+func (r *Repository) referrersByTagSchema(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error {
+	ctx = registryutil.WithScopeHint(ctx, r.Reference, auth.ActionPull)
```
shizhMSFT, line 465: We can remove this line as the hint will be added in `FetchReference()`.

registry/remote/repository.go
```diff
@@ -361,38 +436,86 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
+
+	var index ocispec.Index
+	lr := limitReader(rc, r.MaxMetadataBytes)
+	indexJSON, err := content.ReadAll(lr, desc)
+	if err != nil {
+		return fmt.Errorf("failed to read referrers index from referrers tag %s: %w", referrersTag, err)
+	}
+	if err := json.Unmarshal(indexJSON, &index); err != nil {
+		return fmt.Errorf("failed to decode referrers index from referrers tag %s: %w", referrersTag, err)
+	}
```
shizhMSFT, line 485: We can follow the same logic in the `referrersPageByAPI()`.
```suggestion
	if err := json.NewDecoder(lr).Decode(&index); err != nil {
		return fmt.Errorf("failed to read referrers index from referrers tag %s: %w", referrersTag, err)
	}
```

registry/remote/repository.go
Wwwsylvia: right!

registry/remote/repository.go
Wwwsylvia: Used `content.VerifyReader`

registry/remote/repository.go
Wwwsylvia: This makes sense.

registry/remote/repository.go
```diff
@@ -361,38 +439,86 @@ func (r *Repository) referrers(ctx context.Context, artifactType string, fn func
...
+		return fmt.Errorf("failed to read referrers index from referrers tag %s: %w", referrersTag, err)
+	}
+	vr := content.NewVerifyReader(rc, desc)
+	var index ocispec.Index
+	if err := json.NewDecoder(vr).Decode(&index); err != nil {
+		return fmt.Errorf("failed to decode referrers index from referrers tag %s: %w", referrersTag, err)
+	}
+	if err := vr.Verify(); err != nil {
+		return fmt.Errorf("failed to verify referrers index: referrers tag: %s: %w", referrersTag, err)
+	}
```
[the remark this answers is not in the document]
shizhMSFT: Then I find an edge case. If the index manifest is valid but not well formed (e.g. have trailing spaces). The index will be decoded properly but the `Verify()` operation will fail with error message like "early verification".

There are two possible solutions:
1. Use `content.ReadAll` like in previous iterations of this PR.
2. Use `io.Copy(io.Discard, vr)` to skip rest data.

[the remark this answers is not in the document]
Wwwsylvia: Refactored and took solution 1.

registry/remote/utils.go
```diff
@@ -61,3 +66,25 @@ func limitReader(r io.Reader, n int64) io.Reader {
...
+}
+
+// decodeJSON safely reads the JSON content described by desc, and
+// decodes it into v.
+func decodeJSON(r io.Reader, desc ocispec.Descriptor, v any) error {
+	indexJSON, err := content.ReadAll(r, desc)
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(indexJSON, v)
```
shizhMSFT, line 89: nit:
```suggestion
	jsonBytes, err := content.ReadAll(r, desc)
	if err != nil {
		return err
	}
	return json.Unmarshal(jsonBytes, v)
```

Wwwsylvia: fixed

shizhMSFT (approved):
LGTM

Issue #341, opened 2022-10-09
Push/Pull file example


yuvalturg:
Hi,

Is there an example for a simple push-pull of a single, large (can't load it's entire content to memory) file to a repository?

Thanks!

Wwwsylvia:
Hi @yuvalturg, can you elaborate your scenario a little bit?
- Do you want to push/pull the file as a single blob?
- Do you want to have a manifest to reference this blob (file)?

Basically, how do you want to access this file after you pushing it to a repository?

yuvalturg:
Yes a single layer is fine. Similar to what is described here https://oras.land/cli/1_pushing/

Oras push repo:tag /tmp/bigfile

shizhMSFT:
@Wwwsylvia Could you write an example for @yuvalturg. Besides, we probably need `content.NewDescriptorFromReader()` to simplify the experience.

yuvalturg:
Thanks, I used oras-py, and it's as simple as
OrasClient().push(target, filenames)

Wwwsylvia:
@yuvalturg OK you will need a manifest to reference your file blob if you want to access it via a tag.
Can you check if this helps? https://github.com/Wwwsylvia/demo-push-file/blob/main/main/main.go
Please let us know if you find any bugs or any bad user experiences.

Wwwsylvia:
> @Wwwsylvia Could you write an example for @yuvalturg? Besides, we probably need `content.NewDescriptorFromReader()` to simplify the experience.

Not really, I think a file store is needed anyway in this case 🤔

yuvalturg:
@Wwwsylvia is there a way to get the local filename(s) from the filestore after pulling?

Wwwsylvia:
> @Wwwsylvia is there a way to get the local filename(s) from the filestore after pulling?

@yuvalturg Regarding "get the local filename(s) from the filestore", my understanding is that you want to check the filenames associated with the pulled tag? Does the below code work for you?

```go
	desc, err := fs.Resolve(ctx, tag)
	if err != nil {
		panic(err)
	}
	blobs, err := content.Successors(ctx, fs, desc)
	if err != nil {
		panic(err)
	}
	for _, b := range blobs {
		if name := b.Annotations[v1.AnnotationTitle]; name != "" {
			fmt.Println("filename: ", name)
		}
	}
```

yuvalturg:
> > @Wwwsylvia is there a way to get the local filename(s) from the filestore after pulling?
> 
[...]

PR #343, opened 2022-10-17
refactor!: merge `oras.PackArtifact()` into `oras.Pack()`

### LINKED ISSUE #309: Merge `oras.PackArtifact()` into `oras.Pack()`

Wwwsylvia:
1. Removed `oras.PackArtifact()`
2. Refactored `oras.Pack()` to pack artifact manifest or image manifest
3. Removed the dependency of `github.com/oras-project/artifacts-spec`

Resolves #309
BREAKING CHANGE: Remove `oras.PackArtifact()`, and change the signature of `oras.Pack()`

### DIFF
--- a/content/file/file.go
+++ b/content/file/file.go
@@ -440,7 +440,7 @@ func (s *Store) PackFiles(ctx context.Context, names []string) (ocispec.Descript
 		layers = append(layers, desc)
 	}
 
-	return oras.Pack(ctx, s, layers, oras.PackOptions{})
+	return oras.Pack(ctx, s, "", layers, oras.PackOptions{})
 }
 
 // saveFile saves content matching the descriptor to the given file.
--- a/errdef/errors.go
+++ b/errdef/errors.go
@@ -20,11 +20,12 @@ import "errors"
 // Common errors used in ORAS
 var (
 	ErrAlreadyExists      = errors.New("already exists")
-	ErrUnsupported        = errors.New("unsupported")
-	ErrInvalidReference   = errors.New("invalid reference")
+	ErrBadRequest         = errors.New("bad request")
 	ErrInvalidDigest      = errors.New("invalid digest")
-	ErrNotFound           = errors.New("not found")
-	ErrUnsupportedVersion = errors.New("unsupported version")
+	ErrInvalidReference   = errors.New("invalid reference")
 	ErrMissingReference   = errors.New("missing reference")
+	ErrNotFound           = errors.New("not found")
 	ErrSizeExceedsLimit   = errors.New("size exceeds limit")
+	ErrUnsupported        = errors.New("unsupported")
+	ErrUnsupportedVersion = errors.New("unsupported version")
 )
--- a/errdef/errors.go
+++ b/errdef/errors.go
@@ -20,7 +20,6 @@ import "errors"
 // Common errors used in ORAS
 var (
 	ErrAlreadyExists      = errors.New("already exists")
-	ErrBadRequest         = errors.New("bad request")
 	ErrInvalidDigest      = errors.New("invalid digest")
 	ErrInvalidReference   = errors.New("invalid reference")
 	ErrMissingReference   = errors.New("missing reference")
--- a/pack.go
+++ b/pack.go
@@ -60,12 +60,12 @@ type PackOptions struct {
 	// Default: false.
 	PackImageManifest bool
 	// ConfigDescriptor is a pointer to the descriptor of the config blob.
-	// If specified, artifactType will be implied by the media type of the
-	// specified descriptor, and ConfigAnnotations will be ignored.
+	// If not nil, artifactType will be implied by the mediaType of the
+	// specified ConfigDescriptor, and ConfigAnnotations will be ignored.
 	// This option is valid only when PackImageManifest is true.
 	ConfigDescriptor *ocispec.Descriptor
 	// ConfigAnnotations is the annotation map of the config descriptor.
-	// This option is valid only when PackImageManifest is true,
+	// This option is valid only when PackImageManifest is true
 	// and ConfigDescriptor is nil.
 	ConfigAnnotations map[string]string
 }
@@ -106,6 +106,7 @@ func packArtifact(ctx context.Context, pusher content.Pusher, artifactType strin
 		return ocispec.Descriptor{}, fmt.Errorf("failed to marshal manifest: %w", err)
 	}
 	manifestDesc := content.NewDescriptorFromBytes(ocispec.MediaTypeArtifactManifest, manifestJSON)
+	// populate ArtifactType and Annotations of the manifest into manifestDesc
 	manifestDesc.ArtifactType = manifest.ArtifactType
 	manifestDesc.Annotations = manifest.Annotations
 
@@ -165,6 +166,7 @@ func packImage(ctx context.Context, pusher content.Pusher, configMediaType strin
 		return ocispec.Descriptor{}, fmt.Errorf("failed to marshal manifest: %w", err)
 	}
 	manifestDesc := content.NewDescriptorFromBytes(ocispec.MediaTypeImageManifest, manifestJSON)
+	// populate ArtifactType and Annotations of the manifest into manifestDesc
 	manifestDesc.ArtifactType = manifest.Config.MediaType
 	manifestDesc.Annotations = manifest.Annotations
 
--- a/pack.go
+++ b/pack.go
@@ -63,11 +63,12 @@ type PackOptions struct {
 	PackImageManifest bool
 	// ConfigDescriptor is a pointer to the descriptor of the config blob.
 	// If specified, artifactType will be implied by the media type of the
-	// specified descriptor.
+	// specified descriptor, and ConfigAnnotations will be ignored.
 	// This option is valid only when PackImageManifest is true.
 	ConfigDescriptor *ocispec.Descriptor
 	// ConfigAnnotations is the annotation map of the config descriptor.
-	// This option is valid only when PackImageManifest is true.
+	// This option is valid only when PackImageManifest is true,
+	// and ConfigDescriptor is nil.
 	ConfigAnnotations map[string]string
 }
 
--- a/pack.go
+++ b/pack.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/image-spec/specs-go"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"oras.land/oras-go/v2/content"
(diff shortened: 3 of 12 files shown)


pack.go
```diff
@@ -128,62 +187,26 @@ func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descripto
...
-		// artifactType is required for ORAS Artifact Manifest
-		return ocispec.Descriptor{}, ErrMissingArtifactType
-	}
-
-	if createdTime, ok := opts.ManifestAnnotations[artifactspec.AnnotationArtifactCreated]; ok {
-		// if AnnotationArtifactCreated is provided, validate its format
+// ensureAnnotationCreated ensures that annotationCreatedKey is in annotations,
+// and that its value conforms to RFC 3339. Otherwise returns a new annotation
+// map with annotationCreatedKey created.
+func ensureAnnotationCreated(annotations map[string]string, annotationCreatedKey string) (map[string]string, error) {
```
Wwwsylvia, line 193: The creation time annotation is not mandatory in the [spec](https://github.com/opencontainers/image-spec/blob/main/artifact.md), but I think it's good to add it as a built-in annotation of oras-go.

pack.go
```diff
@@ -26,53 +26,111 @@ import (
...
+		artifactType = MediaTypeUnknownArtifact
+	}
+
+	annotations, err := ensureAnnotationCreated(opts.ManifestAnnotations, ocispec.AnnotationArtifactCreated)
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	if blobs == nil {
+		blobs = []ocispec.Descriptor{} // make it an empty array to prevent potential server-side bugs
+	}
```
shizhMSFT, line 99: Unlike the image manifest. we don't need this for artifact manifest. It is `OPTIONAL` in the spec.

pack.go
```diff
@@ -26,53 +26,111 @@ import (
...
+	if err != nil {
+		return ocispec.Descriptor{}, fmt.Errorf("failed to marshal manifest: %w", err)
+	}
+	manifestDesc := ocispec.Descriptor{
+		MediaType:    ocispec.MediaTypeArtifactManifest,
+		Digest:       digest.FromBytes(manifestBytes),
+		Size:         int64(len(manifestBytes)),
+		ArtifactType: manifest.ArtifactType,
+		Annotations:  manifest.Annotations,
+	}
```
shizhMSFT, line 117: nit: do we want to use `content.NewDescriptorFromBytes`?

pack.go
```diff
@@ -26,53 +26,111 @@ import (
...
-// MediaTypeUnknownConfig is the default mediaType used when no
-// config media type is specified.
-const MediaTypeUnknownConfig = "application/vnd.unknown.config.v1+json"
+const (
+	// MediaTypeUnknownConfig is the default mediaType used when no
+	// config media type is specified.
+	MediaTypeUnknownConfig = "application/vnd.unknown.config.v1+json"
+	// MediaTypeUnknownArtifact is the default artifactType used when no
+	// artifact type is specified.
+	MediaTypeUnknownArtifact = "application/vnd.unknown.artifact.v1"
```
wangxiaoxuan273, line 39: This variable is for artifact type. Should it be named _**ArtifactType**UnknownArtifact_ instead of _**MediaType**UnknownArtifact_?

content/file/file.go
```diff
@@ -440,7 +440,7 @@ func (s *Store) PackFiles(ctx context.Context, names []string) (ocispec.Descript
 		layers = append(layers, desc)
 	}
 
-	return oras.Pack(ctx, s, layers, oras.PackOptions{})
+	return oras.Pack(ctx, s, "", layers, oras.PackOptions{})
```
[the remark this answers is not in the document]
Wwwsylvia: This is just a convenient method for basic users. We can consider whether to support it in the GA version. 🤔

pack_test.go
```diff
@@ -37,124 +35,128 @@ func Test_Pack_Default(t *testing.T) {
 	s := memory.New()
 
 	// prepare test content
-	layer_1 := []byte("hello world")
+	blob_1 := []byte("hello world")
 	desc_1 := ocispec.Descriptor{
 		MediaType: "test",
-		Digest:    digest.FromBytes(layer_1),
-		Size:      int64(len(layer_1)),
+		Digest:    digest.FromBytes(blob_1),
```
wangxiaoxuan273, line 41: nit: should we use `content.NewDescriptorFromBytes()` to construct the descriptors?

pack.go
```diff
@@ -84,40 +142,41 @@ func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descripto
...
 	manifestDesc := ocispec.Descriptor{
-		MediaType: ocispec.MediaTypeImageManifest,
-		Digest:    digest.FromBytes(manifestBytes),
-		Size:      int64(len(manifestBytes)),
+		MediaType:    ocispec.MediaTypeImageManifest,
+		Digest:       digest.FromBytes(manifestBytes),
+		Size:         int64(len(manifestBytes)),
+		ArtifactType: manifest.Config.MediaType,
+		Annotations:  manifest.Annotations,
 	}
```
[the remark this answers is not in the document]
Wwwsylvia: yes!😄

pack.go
```diff
@@ -26,53 +26,111 @@ import (
...
-	// ManifestAnnotations is the annotation map of the manifest.
-	ManifestAnnotations map[string]string
+// Pack packs the given blobs, generates a manifest for the pack,
+// and pushes it to a content storage.
+// If succeeded, returns a descriptor of the manifest.
+func Pack(ctx context.Context, pusher content.Pusher, artifactType string, blobs []ocispec.Descriptor, opts PackOptions) (ocispec.Descriptor, error) {
+	if opts.PackImageManifest {
+		return packImage(ctx, pusher, artifactType, blobs, opts)
+	}
+	return packArtifact(ctx, pusher, artifactType, blobs, opts)
```
wangxiaoxuan273, line 82: nit: Should we rename the variable `artifactType`? As it is artifact type in `PackImage` and media type in `PackArtifact`. A more generic name like `type` may be less confusing.

pack.go
Wwwsylvia: Good point! My considerations are:
1. `artifactType` is actually a media type
> **`artifactType`** _string_
This property SHOULD be used and contain the mediaType of the referenced artifact. 
Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md#artifact-manifest

2.  When `PackImageManifest` is true, `artifactType` will be used as the config media type
(See https://github.com/Wwwsylvia/oras-go/blob/be6588adc63a903be64ebb9ae6d5d06a1b7aad21/pack.go#L145)
> **`artifactType`** _string_
This OPTIONAL property contains the type of an artifact when the descriptor points to an artifact. This is the value of artifactType when the descriptor references an [artifact manifest](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/artifact.md). This is the value of the config descriptor mediaType when the descriptor references an [image manifest](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md).
Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/descriptor.md

pack.go
Wwwsylvia: Yeah this is confusing... But `type` is also confusing😂 (naming is hard!).
Maybe we should add more docs here.

pack_test.go
Wwwsylvia: Refactored.

errdef/errors.go
```diff
@@ -20,11 +20,12 @@ import "errors"
 // Common errors used in ORAS
 var (
 	ErrAlreadyExists      = errors.New("already exists")
-	ErrUnsupported        = errors.New("unsupported")
-	ErrInvalidReference   = errors.New("invalid reference")
+	ErrBadRequest         = errors.New("bad request")
```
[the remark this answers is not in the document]
Wwwsylvia: The problem is that we can't tell the reason of the 400 response, it can be because the manifest media type is not supported, or the manifest format is invalid. What if some users push an artifact manifest in an invalid form, and the server returns 400 while it actually supports artifact manifest?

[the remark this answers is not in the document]
Wwwsylvia: Mmmm, it may not be a problem after we implement referrers compatibility for manifest push, as the content will be parsed before being pushed anyway.

pack.go
```diff
@@ -23,56 +23,108 @@ import (
...
+	//   - If true, pack an image manifest; artifactType will be used as the
+	// the config descriptor mediaType of the image manifest.
+	// media type.
+	//   - If false, pack an artifact manifest.
+	// Default: false.
+	PackImageManifest bool
 	// ConfigDescriptor is a pointer to the descriptor of the config blob.
+	// If specified, artifactType will be implied by the media type of the
+	// specified descriptor, and ConfigAnnotations will be ignored.
+	// This option is valid only when PackImageManifest is true.
```
qweeah, line 66: Can we use `ConfigDescriptor==nil` to imply`PackImageManifest==true`?

qweeah: Please resolve this, in the fallback case, we might need to use `PackImageManifest=true` and `ConfigDescriptor!=nil` to store artifact type

pack.go
```diff
@@ -23,56 +23,108 @@ import (
...
+// packImage packs the given blobs, generates an image manifest for the pack,
+// and pushes it to a content storage. artifactType will be used as the config
+// descriptor mediaType of the image manifest.
 // If succeeded, returns a descriptor of the manifest.
-func Pack(ctx context.Context, pusher content.Pusher, layers []ocispec.Descriptor, opts PackOptions) (ocispec.Descriptor, error) {
-	if opts.ConfigMediaType == "" {
-		opts.ConfigMediaType = MediaTypeUnknownConfig
+func packImage(ctx context.Context, pusher content.Pusher, configMediaType string, layers []ocispec.Descriptor, opts PackOptions) (ocispec.Descriptor, error) {
+	if configMediaType == "" {
+		configMediaType = MediaTypeUnknownConfig
```
qweeah, line 127: if `opts.PackImageManifest` is `true`, should we set this to `MediaTypeUnknownArtifact`?

Wwwsylvia: Sorry I don't get it 🤔

qweeah: Filled the gap offline. If oci artifact manifest is not supported, it's `Pack` callee's duty to repack a manifest and repush.

shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

content/file/file.go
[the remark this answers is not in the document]
Wwwsylvia: Here is an use case of this method: https://github.com/oras-project/oras-go/issues/341

PR #351, opened 2022-10-27
refactor: return `ErrBadRequest` for manifest push

### LINKED ISSUE #344: How to distinguish "unsupported manifest" error from other `400 Bad Request` errors on manifest push?
Remote registries return `400 Bad Request` error on manifest push when:
1. The pushed manifest is malformed
2. The layers referenced by the manifest do not exist
3. The registry does not support the media type of the pushed manifest

Currently, `oras-go` is unable to distinguish case 3 from case 1 and 2 on manifest push.
This blocks the below scenario:

1. A client builds an artifact manifest and pushes is to a remote registry
2. The client finds that the registry does not support artifact manifest
3. The client fallback to push an image manifest to the registry

Related code:
https://github.com/oras-project/oras-go/blob/1969551cc3c77b46f09090f00360e8465871fe8f/registry/remote/repository.go#L1048-L1051

Wwwsylvia:
Resolves #344

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -1092,7 +1092,7 @@ func (s *manifestStore) push(ctx context.Context, expected ocispec.Descriptor, c
 		return verifyContentDigest(resp, expected.Digest)
 	case http.StatusBadRequest:
 		err := errutil.ParseErrorResponse(resp)
-		return fmt.Errorf("%w: %s: %s: %v", ErrBadRequest, expected.Digest, expected.MediaType, err)
+		return fmt.Errorf("%s: %s: %w: %v", expected.Digest, expected.MediaType, ErrBadRequest, err)
 	default:
 		return errutil.ParseErrorResponse(resp)
 	}
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -63,9 +63,16 @@ const (
 	referrersStateUnsupported
 )
 
-// ErrReferrersCapabilityAlreadySet is returned by SetReferrersCapability()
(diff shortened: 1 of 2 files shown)


registry/remote/repository.go
```diff
@@ -1080,10 +1087,15 @@ func (s *manifestStore) push(ctx context.Context, expected ocispec.Descriptor, c
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusCreated {
+	switch resp.StatusCode {
+	case http.StatusCreated:
+		return verifyContentDigest(resp, expected.Digest)
+	case http.StatusBadRequest:
+		err := errutil.ParseErrorResponse(resp)
+		return fmt.Errorf("%s: %s: %w: %v", expected.Digest, expected.MediaType, ErrBadRequest, err)
```
shizhMSFT, line 1095: It is weird to have `ErrBadRequest` here as it cannot indicate invalid manifest.

That is, `MANIFEST_INVALID` implies but is not implied by `400 Bad Request`.

shizhMSFT: Furthermore, the `ErrBadRequest` does not contain the context. In other words, we are not able to distinguish a real bad request with an unsupported manifest in a `oras.Copy()` scenario with lots of manifests.

shizhMSFT (changes requested):

Wwwsylvia:
Closing this PR as there is a better approach. See https://github.com/opencontainers/distribution-spec/issues/359.

PR #352, opened 2022-10-28
feat: include error code in error response

### LINKED ISSUE #344: How to distinguish "unsupported manifest" error from other `400 Bad Request` errors on manifest push?

Wwwsylvia:
Resolves #344


shizhMSFT (approved):
LGTM

PR #359, opened 2022-11-09
fix: fix potential deadlock in `oras.Copy`

### LINKED ISSUE #349: Bug: Potential dead lock in `oras.Copy`
Assuming there is a DAG, where multiple nodes reference the same node, as shown in the chart below.

```mermaid
flowchart TD
B --> F;
C --> F;
D --> F;
E --> F;
A --> B;
A --> C;
A --> D;
A --> E;
```

When copying such DAGs from a Target to another, there is a chance that `oras.Copy()` will encounter dead lock and never finish.
Repro: https://github.com/Wwwsylvia/oras-go/blob/repro_copy_deadlock/repro_test.go

Wwwsylvia:
1. Refactored `copyGraph` and internal graph utilities
2. Added a test case that could cause dead locks 

Fixes #349 


internal/syncutil/limit.go
```diff
@@ -0,0 +1,84 @@
...
+
+// LimitRegion provides a way to bound concurrent access to a code block.
+type LimitRegion struct {
+	ctx     context.Context
+	limiter *semaphore.Weighted
+	ended   bool
+}
+
+// NewLimitRegion creates a new LimitRegion.
+func NewLimitRegion(ctx context.Context, limiter *semaphore.Weighted) *LimitRegion {
```
shizhMSFT, line 33: Let's follow the name of `LimitedReader` and `LimitReader()` in the package `io`.

Then we have
- `LimitRegion` --> `LimitedRegion`
- `NewLimitRegion` --> `LimitRegion`

Wwwsylvia: Sounds good.

shizhMSFT (approved):
LGTM

PR #363, opened 2022-11-20
fix: avoid panic from releasing unacquired semaphore

abursavich:
When using this pattern:

```go
eg, egCtx := errgroup.WithContext(ctx)
for /* ... */ {
	limiter.Acquire(ctx, 1)
	eg.Go(func() error {
		defer limiter.Release(1)
		// ...
		return nil
	})
}
if err := eg.Wait(); err != nil {
	return err
}
```

If `ctx` is cancelled then `Acquire` may return an error, which will then cause `Release` to panic after it's called more times than the semaphore was successfully acquired.

This change uses the [SetLimit](https://pkg.go.dev/golang.org/x/sync@v0.1.0/errgroup#Group.SetLimit) functionality which has been added to `errgroup.Group` and provides a small wrapper around it in the internal/syncutil package to improve the ergonomics of its usage.


shizhMSFT:
@abursavich Thank you for your contribution.

I thought about using `SetLimit` instead of `semaphore`. However, when I look into the implementation of [SetLimit()](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.1.0:errgroup/errgroup.go;l=131), I find
```go
func (g *Group) SetLimit(n int) {
	if n < 0 {
		g.sem = nil
		return
	}
	if len(g.sem) != 0 {
		panic(fmt.Errorf("errgroup: modify limit while %v goroutines in the group are still active", len(g.sem)))
	}
	g.sem = make(chan token, n)
}
```
The statement `g.sem = make(chan token, n)` makes `sync.Group` with `SetLimit(n int)` having a space complexity of `O(n)` while the `semaphore.Weighted` has a space complexity of `O(1)`. Therefore, I think the original implementation is better.

However, your concern is still valid. The pattern
```go
eg, egCtx := errgroup.WithContext(ctx)
for /* ... */ {
	limiter.Acquire(ctx, 1)
	eg.Go(func() error {
		defer limiter.Release(1)
		// ...
		return nil
	})
}
if err := eg.Wait(); err != nil {
	return err
}
```
is flawed since the error of `limiter.Acquire(ctx, 1)` is never checked. I think the fix will be as simple as checking the error as follows.
```go
eg, egCtx := errgroup.WithContext(ctx)
for /* ... */ {
	if err := limiter.Acquire(ctx, 1); err != nil {
		return err
	}
	eg.Go(func() error {
		defer limiter.Release(1)
		// ...
		return nil
	})
}
if err := eg.Wait(); err != nil {
	return err
}
```

shizhMSFT:
BTW, thank you for informing `golang.org/x/sync` has a tagged version `v0.1.0`.

abursavich:
> The statement `g.sem = make(chan token, n)` makes `sync.Group` with `SetLimit(n int)` having a space complexity of `O(n)` while the `semaphore.Weighted` has a space complexity of `O(1)`. Therefore, I think the original implementation is better.

The type used in the channel is actually 0 bytes: https://go.dev/play/p/ctojkieS8zW

```go
type token struct{}

// ...

func (g *Group) SetLimit(n int) {
	// ...
	g.sem = make(chan token, n)
}
```

So its space complexity ends up being `O(1)` too :)


> I think the fix will be as simple as checking the error as follows.
> 
> ```go
> eg, egCtx := errgroup.WithContext(ctx)
> for /* ... */ {
> 	if err := limiter.Acquire(ctx, 1); err != nil {
> 		return err
> 	}
> 	eg.Go(func() error {
> 		defer limiter.Release(1)
> 		// ...
> 		return nil
> 	})
> }
> if err := eg.Wait(); err != nil {
> 	return err
> }
> ```

This was also my initial fix, but I ended up changing it because:
- If you use ctx to acquire the limiter, it will keep going after an error has occurred and egCtx has been cancelled.
- If you use egCtx to acquire the limiter to fix that problem, it will return a "context cancelled" error instead of the underlying error that triggered the cancellation.

shizhMSFT:
@abursavich You are right. The `struct{}` is indeed special and has zero cost in the go [channel](https://go.dev/src/runtime/chan.go) implementation.
```go
c.buf = mallocgc(mem, elem, true)
```

content.go
```diff
@@ -119,13 +118,10 @@ func TagN(ctx context.Context, target Target, srcReference string, dstReferences
 			return err
 		}
 
-		limiter := semaphore.NewWeighted(opts.Concurrency)
-		eg, egCtx := errgroup.WithContext(ctx)
+		eg, egCtx := syncutil.NewLimitGroup(ctx, int(opts.Concurrency))
```
shizhMSFT, line 121: There is an issue on casting `int64` to `int`. What if we lose precisions?

@Wwwsylvia Should we change the type of all occurrences of `Concurrency` to `int` from `int64`?

internal/syncutil/limitgroup.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// A LimitGroup is a collection of goroutines working on subtasks that are part of
+// the same overall task.
+type LimitGroup struct {
```
shizhMSFT, line 26: Instead of `LimitGroup` / `NewLimitGroup()` pair, it is better to align the name to `LimitedGroup` / `LimitGroup()`, just like `io.LimitedReader` / `io.LimitReader()` and `LimitedRegion` / `LimitRegion`.

shizhMSFT:
@abursavich @Wwwsylvia Do you also want to revisit [internal/syncutil/limit.go](https://github.com/oras-project/oras-go/blob/main/internal/syncutil/limit.go) for context cancellation?

content.go
Wwwsylvia: Yeah `int` should be enough.

internal/syncutil/limitgroup.go
abursavich: I disagree with the change to the constructor's name, but I made the change as requested.

`LimitReader` is taking in a `Reader` and limiting it, so its name is like a command "limit this reader". I think `NewLimitedGroup` would be better than `LimitGroup`, since it's creating the `LimitedGroup` and not just wrapping an existing `Group` with limits.

abursavich:
> @abursavich @Wwwsylvia Do you also want to revisit [internal/syncutil/limit.go](https://github.com/oras-project/oras-go/blob/main/internal/syncutil/limit.go) for context cancellation?

I think that this type of concurrent code is tricky and you should probably use whatever you're most comfortable with. I'm more comfortable with `for { select {} }` loops that manage worker goroutines (kinda like [this talk](https://www.youtube.com/watch?v=QDDwwePbDtw)). 

`LimitedRegion` has the `ended` field and guards all over the place to protect the `Semaphore` from being released when it isn't held. That's nice to avoid panics, but it makes me concerned that the code using it may be invalid.

Could probably tweak the `Go` function a little to use the `egCtx` and short-circuit on error:
```go
func Go[T any](ctx context.Context, limiter *semaphore.Weighted, fn GoFunc[T], items ...T) error {
	eg, egCtx := errgroup.WithContext(ctx)
	for _, item := range items {
		region := LimitRegion(egCtx, limiter)
			eg.Go(func() error { return err })
			break
		}
		eg.Go(func(t T) func() error {
			return func() error {
				defer region.End()
				return fn(egCtx, region, t)
			}
		}(item))
	}
	return eg.Wait()
}
```

TerryHowe:
rebsae?

abursavich:
I've rebased this a couple times to keep it up to date with main. It's unclear to me if there are any more requested changes to this PR, or if the more recent discussion is about other related changes that should be made separately?

Wwwsylvia (approved):
LGTM

Wwwsylvia:
> I've rebased this a couple times to keep it up to date with main. It's unclear to me if there are any more requested changes to this PR, or if the more recent discussion is about other related changes that should be made separately?

LGTM. Let's wait for review from @shizhMSFT.

> I disagree with the change to the constructor's name, but I made the change as requested.

LimitReader is taking in a Reader and limiting it, so its name is like a command "limit this reader". I think NewLimitedGroup would be better than LimitGroup, since it's creating the LimitedGroup and not just wrapping an existing Group with limits.

+1 for `NewLimitedGroup`.

content.go
Wwwsylvia: Sent a PR for this. See #376

PR #364, opened 2022-11-20
fix: avoid copying Repository mutex state

abursavich:
The `Repository` struct contains mutex state (in the `sync.Mutex` and `syncutil.Pool`). When converting `RepositoryOptions` values to `Repository` values, this mutex state is copied. This change prevents the internal state from being copied by passing `RepositoryOptions` around by reference and explicitly copying the visible fields.

### DIFF
--- a/registry/remote/registry.go
+++ b/registry/remote/registry.go
@@ -168,11 +168,5 @@ func (r *Registry) Repository(ctx context.Context, name string) (registry.Reposi
 		Registry:   r.Reference.Registry,
 		Repository: name,
 	}
-	if err := ref.ValidateRepository(); err != nil {
-		return nil, err
-	}
-
-	repo := Repository(r.RepositoryOptions)
-	repo.Reference = ref
-	return &repo, nil
+	return newRepositoryWithOptions(ref, &r.RepositoryOptions)
 }
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -100,6 +100,8 @@ type Repository struct {
 	// If less than or equal to zero, a default (currently 4MiB) is used.
 	MaxMetadataBytes int64
 
+	// NOTE: Must keep fields in sync with newRepositoryWithOptions function.
+
 	// referrersState represents that if the repository supports Referrers API.
 	// default: referrersStateUnknown
 	referrersState referrersState
@@ -124,6 +126,33 @@ func NewRepository(reference string) (*Repository, error) {
 	return &Repository{
 		Reference: ref,
 	}, nil
+}
+
+// newRepositoryWithOptions returns a Repository with the given reference and options.
+// RepositoryOptions are part of the Registry struct and may set defaults for the Registry.
+// RepositoryOptions share the same struct definition as Repository and therfore contain
+// unexported state that we must avoid copying around to different Repositories.
+// To handle this we explicitly copy only the fields that we want to reproduce.
+func newRepositoryWithOptions(ref registry.Reference, opts *RepositoryOptions) (*Repository, error) {
+	if err := ref.ValidateRepository(); err != nil {
+		return nil, err
+	}
+	return &Repository{
+		Client:               opts.Client,
+		Reference:            ref,
+		PlainHTTP:            opts.PlainHTTP,
+		ManifestMediaTypes:   cloneSlice(opts.ManifestMediaTypes),
+		TagListPageSize:      opts.TagListPageSize,
+		ReferrerListPageSize: opts.ReferrerListPageSize,
+		MaxMetadataBytes:     opts.MaxMetadataBytes,
+	}, nil
+}
+
+func cloneSlice[S ~[]E, E any](s S) S {
+	if s == nil {
+		return nil
+	}
+	return append(make(S, 0, len(s)), s...)
 }
 
 // SetReferrersCapability indicates the Referrers API capability of the remote
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -38,6 +38,7 @@ import (
 	"oras.land/oras-go/v2/internal/httputil"
 	"oras.land/oras-go/v2/internal/ioutil"
 	"oras.land/oras-go/v2/internal/registryutil"
+	"oras.land/oras-go/v2/internal/slices"
 	"oras.land/oras-go/v2/internal/syncutil"
 	"oras.land/oras-go/v2/registry"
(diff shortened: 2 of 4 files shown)


abursavich:
I considered also copying the `referrersState`, but I ultimately decided against it since `RepositoryOptions` blocks the methods on `Repository` that can mutate it.

Wwwsylvia:
> The Repository struct contains mutex state (in the sync.Mutex and syncutil.Pool). When converting RepositoryOptions values to Repository values, this mutex state is copied. 

Nice catch!

registry/remote/registry.go
```diff
@@ -168,11 +168,5 @@ func (r *Registry) Repository(ctx context.Context, name string) (registry.Reposi
...
 		Repository: name,
 	}
-	if err := ref.ValidateRepository(); err != nil {
-		return nil, err
-	}
-
-	repo := Repository(r.RepositoryOptions)
-	repo.Reference = ref
-	return &repo, nil
+	return newRepositoryWithOptions(ref, &r.RepositoryOptions)
```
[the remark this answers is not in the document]
shizhMSFT: The `newRepositoryWithOptions` is necessary here. Different repositories should have different instances of mutex and so on.

registry/remote/repository.go
```diff
@@ -126,6 +129,21 @@ func NewRepository(reference string) (*Repository, error) {
 	}, nil
 }
 
+func newRepositoryWithOptions(ref registry.Reference, opts *RepositoryOptions) (*Repository, error) {
```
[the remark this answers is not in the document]
shizhMSFT: LGTM

shizhMSFT, line 132: It would be better if we can have documentation on `newRepositoryWithOptions`.

go.mod
```diff
@@ -5,5 +5,6 @@ go 1.18
 require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.0-rc2
+	golang.org/x/exp v0.0.0-20221114191408-850992195362
```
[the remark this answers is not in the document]
shizhMSFT: Agree. It is not worthy to add dependencies just for `slices.Clone`, which is easy to implement.
```go
func Clone[S ~[]E, E any](s S) S {
	// Preserve nil in case it matters.
	if s == nil {
		return nil
	}
	return append(S([]E{}), s...)
}
```

registry/remote/repository.go
```diff
@@ -126,6 +129,21 @@ func NewRepository(reference string) (*Repository, error) {
...
+	if err := ref.ValidateRepository(); err != nil {
+		return nil, err
+	}
+	return &Repository{
+		Client:               opts.Client,
+		Reference:            ref,
+		PlainHTTP:            opts.PlainHTTP,
+		ManifestMediaTypes:   slices.Clone(opts.ManifestMediaTypes),
```
shizhMSFT, line 140: Consider the following code to remove the dependency on `golang.org/x/exp`.
```suggestion
		ManifestMediaTypes:   append([]string{}, opts.ManifestMediaTypes...),
```
Of course, we can have a `if` check if `nil` matters.

shizhMSFT:
@abursavich Thank you for your contribution and catching this bug. Could you address the review comments accordingly so that we can merged your PR?

registry/remote/repository.go
abursavich: Done.

go.mod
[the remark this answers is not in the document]
abursavich: Done.

abursavich:
@shizhMSFT, PTAL.

registry/remote/repository.go
```diff
@@ -126,6 +128,33 @@ func NewRepository(reference string) (*Repository, error) {
...
+		Reference:            ref,
+		PlainHTTP:            opts.PlainHTTP,
+		ManifestMediaTypes:   cloneSlice(opts.ManifestMediaTypes),
+		TagListPageSize:      opts.TagListPageSize,
+		ReferrerListPageSize: opts.ReferrerListPageSize,
+		MaxMetadataBytes:     opts.MaxMetadataBytes,
+	}, nil
+}
+
+func cloneSlice[S ~[]E, E any](s S) S {
```
Wwwsylvia, line 151: nit: Can we have some documentation and unit tests for this new method?

registry/remote/repository.go
```diff
@@ -126,6 +128,33 @@ func NewRepository(reference string) (*Repository, error) {
 	}, nil
 }
 
+// newRepositoryWithOptions returns a Repository with the given reference and options.
```
Wwwsylvia, line 131: nit: It would be better if we can limit the line width of comments to 80.

registry/remote/repository.go
```diff
@@ -126,6 +128,33 @@ func NewRepository(reference string) (*Repository, error) {
...
+// RepositoryOptions are part of the Registry struct and may set defaults for the Registry.
+// RepositoryOptions share the same struct definition as Repository and therfore contain
```
Wwwsylvia, line 133: > RepositoryOptions share the same struct definition as Repository and **therfore** contain

nit: typo

registry/remote/repository.go
```diff
@@ -126,6 +128,33 @@ func NewRepository(reference string) (*Repository, error) {
...
+		MaxMetadataBytes:     opts.MaxMetadataBytes,
+	}, nil
+}
+
+func cloneSlice[S ~[]E, E any](s S) S {
+	if s == nil {
+		return nil
+	}
+	return append(make(S, 0, len(s)), s...)
+}
```
shizhMSFT, line 156: nit: it will be better if we can move it to `internal/slices/slice.go` and name it `Clone[S ~[]E, E any](s S) S`.

shizhMSFT (approved):
LGTM

registry/remote/repository.go
abursavich: Done

registry/remote/repository.go
abursavich: Done

registry/remote/repository.go
abursavich: Done

abursavich:
PTAL

Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #304, opened 2022-09-07
Improve godoc


Wwwsylvia:
Make godoc easier to read.
e.g. https://pkg.go.dev/oras.land/oras-go/v2@main#pkg-variables

Wwwsylvia:
According to [Go Doc Comments](https://tip.golang.org/doc/comment#const), 
> Typed constants are displayed next to the declaration of their type and as a result often omit a const group doc comment in favor of the type’s doc comment. 

> The conventions for variables are the same as those for constants.

We may explicitly declare the type for the variables so that they can be displayed close to the corresponding types.

PR #365, opened 2022-11-22
doc: fix testable example name

Wwwsylvia:
Fix the function name of a testable example so that the example can be showed on godoc.

Related to #304


shizhMSFT (approved):
LGTM

PR #369, opened 2022-12-02
fix: attempt range query only if server explicitly supports

### LINKED ISSUE #368: Blob fetch should not attempt range request by default
`oras-go` always attempts `Range` header to seekable http client, which is aggressive.

https://github.com/oras-project/oras-go/blob/511e4d4d72f641580c73d532f5a277e590661981/registry/remote/repository.go#L616-L657

The Docker spec states that
> This endpoint may also support RFC7233 compliant range requests. Support can be detected by issuing a HEAD request. If the header `Accept-Range: bytes` is returned, range requests can be used to fetch partial content.

Therefore, `oras-go` should do normal request by default and attempt range request only if the server explicitly states it support range query in order to be compatible with most registries with best effort.

Known impacted registry provider:
- jFrog Artifactory

shizhMSFT:
Fixes #368


Wwwsylvia:
Should `FetchReference` apply the same?

Wwwsylvia (approved):
LGTM

PR #377, opened 2022-12-07
feat: change foreign layer error message at repository.Fetch()

### LINKED ISSUE #319: feat: improve the error message if non-distributable(foreign) layer is not supported
Some artifacts(e.g., Windows base images) contain layers whose distribution is restricted by license. When these images are pushed to a registry, restricted artifacts are not included.

For now, `oras.Copy` and `oras.Fetch` will fail if the target artifact contains an non-distributable(foreign) layer since the layer blob doesn't exist in the remote registry. We should
1) By default skip blob copying if a layer is non-distributable
2) Provide option to enforce copying a non-distributable layer and follows the foreign link for fetching

wangxiaoxuan273:
Resolves #319

Current result with oras-cli:
```
$ oras copy docker.io/library/hello-world@sha256:314cc0309465c7be2936df274b258c843f5a49eb21f63b719d60b51564070b8f wxxacr.azurecr.io/temp439:test00

$ Error: the artifact with foreign layer sha256:5ead999142ecce15e02523c49706a633fa708374d94bb65a254e3a3c117d609b is not supported: sha256:5ead999142ecce15e02523c49706a633fa708374d94bb65a254e3a3c117d609b: not found
```

### DIFF
--- a/content.go
+++ b/content.go
@@ -26,6 +26,7 @@ import (
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
 	"oras.land/oras-go/v2/internal/cas"
+	"oras.land/oras-go/v2/internal/descriptor"
 	"oras.land/oras-go/v2/internal/docker"
 	"oras.land/oras-go/v2/internal/interfaces"
 	"oras.land/oras-go/v2/internal/platform"
@@ -273,6 +274,9 @@ func Fetch(ctx context.Context, target ReadOnlyTarget, reference string, opts Fe
 		}
 		rc, err := target.Fetch(ctx, desc)
 		if err != nil {
+			if errors.Is(err, errdef.ErrNotFound) && descriptor.IsForeignLayer(desc) {
+				return ocispec.Descriptor{}, nil, fmt.Errorf("the artifact with foreign layer %s is not supported: %w", desc.Digest, err)
+			}
 			return ocispec.Descriptor{}, nil, err
 		}
 		return desc, rc, nil
@@ -291,6 +295,9 @@ func Fetch(ctx context.Context, target ReadOnlyTarget, reference string, opts Fe
 	proxy.StopCaching = true
 	rc, err := proxy.Fetch(ctx, desc)
 	if err != nil {
+		if errors.Is(err, errdef.ErrNotFound) && descriptor.IsForeignLayer(desc) {
+			return ocispec.Descriptor{}, nil, fmt.Errorf("the artifact with foreign layer %s is not supported: %w", desc.Digest, err)
+		}
 		return ocispec.Descriptor{}, nil, err
 	}
 	return desc, rc, nil
--- a/content.go
+++ b/content.go
@@ -26,7 +26,6 @@ import (
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
 	"oras.land/oras-go/v2/internal/cas"
-	"oras.land/oras-go/v2/internal/descriptor"
 	"oras.land/oras-go/v2/internal/docker"
 	"oras.land/oras-go/v2/internal/interfaces"
 	"oras.land/oras-go/v2/internal/platform"
@@ -274,9 +273,6 @@ func Fetch(ctx context.Context, target ReadOnlyTarget, reference string, opts Fe
 		}
 		rc, err := target.Fetch(ctx, desc)
 		if err != nil {
-			if errors.Is(err, errdef.ErrNotFound) && descriptor.IsForeignLayer(desc) {
-				return ocispec.Descriptor{}, nil, fmt.Errorf("the artifact with foreign layer %s is not supported: %w", desc.Digest, err)
-			}
 			return ocispec.Descriptor{}, nil, err
 		}
 		return desc, rc, nil
@@ -295,9 +291,6 @@ func Fetch(ctx context.Context, target ReadOnlyTarget, reference string, opts Fe
 	proxy.StopCaching = true
 	rc, err := proxy.Fetch(ctx, desc)
 	if err != nil {
-		if errors.Is(err, errdef.ErrNotFound) && descriptor.IsForeignLayer(desc) {
-			return ocispec.Descriptor{}, nil, fmt.Errorf("the artifact with foreign layer %s is not supported: %w", desc.Digest, err)
-		}
 		return ocispec.Descriptor{}, nil, err
 	}
 	return desc, rc, nil
--- a/copy.go
+++ b/copy.go
@@ -26,6 +26,7 @@ import (
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
 	"oras.land/oras-go/v2/internal/cas"
+	"oras.land/oras-go/v2/internal/descriptor"
 	"oras.land/oras-go/v2/internal/platform"
 	"oras.land/oras-go/v2/internal/registryutil"
 	"oras.land/oras-go/v2/internal/status"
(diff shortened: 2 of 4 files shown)


errdef/errors.go
```diff
@@ -27,4 +27,5 @@ var (
 	ErrSizeExceedsLimit   = errors.New("size exceeds limit")
 	ErrUnsupported        = errors.New("unsupported")
 	ErrUnsupportedVersion = errors.New("unsupported version")
+	ErrForeignLayer       = errors.New("foreign layer is not supported")
```
wangxiaoxuan273, line 30: Need advice for a better message that matches the error level.

registry/remote/repository.go
```diff
@@ -614,6 +615,9 @@ type blobStore struct {
 
 // Fetch fetches the content identified by the descriptor.
 func (s *blobStore) Fetch(ctx context.Context, target ocispec.Descriptor) (rc io.ReadCloser, err error) {
+	if isForeignLayer(target) {
+		return nil, fmt.Errorf("%s: %w", target.Digest, errdef.ErrForeignLayer)
+	}
```
wangxiaoxuan273, line 620: Should this check be here, or inside its upstream caller (copy)? It affects what error message to display.

registry/remote/repository.go
```diff
@@ -614,6 +615,9 @@ type blobStore struct {
 
 // Fetch fetches the content identified by the descriptor.
 func (s *blobStore) Fetch(ctx context.Context, target ocispec.Descriptor) (rc io.ReadCloser, err error) {
+	if isForeignLayer(target) {
+		return nil, fmt.Errorf("%s: %w", target.Digest, errdef.ErrForeignLayer)
```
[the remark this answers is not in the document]
wangxiaoxuan273: Impossible :)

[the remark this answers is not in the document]
wangxiaoxuan273: After offline discussion, we improved the error message and resolved this comment.

registry/remote/repository.go
wangxiaoxuan273: Put it here is fine.

errdef/errors.go
wangxiaoxuan273: Changed accordingly, after offline discussion.

registry/remote/repository.go
```diff
@@ -1428,3 +1432,14 @@ func generateIndex(manifests []ocispec.Descriptor) (ocispec.Descriptor, []byte,
...
+
+func isForeignLayer(desc ocispec.Descriptor) bool {
+	switch desc.MediaType {
+	case ocispec.MediaTypeImageLayerNonDistributable,
+		ocispec.MediaTypeImageLayerNonDistributableGzip,
+		ocispec.MediaTypeImageLayerNonDistributableZstd,
+		docker.MediaTypeForeignLayer:
+		return true
+	}
+	return false
```
Wwwsylvia, line 1444: nit:
```suggestion
	default:
	    return false
	}
```

registry/remote/repository.go
Wwwsylvia: It's the right place. We should also apply the same logic to `blobStore.FetchReference()`.

errdef/errors.go
Wwwsylvia: I'm wondering would this specific `ErrForeignLayer` be checked by consumers (such as oras and notation)? If not, maybe we can just reuse `ErrUnsupported`.

registry/remote/repository.go
Wwwsylvia: Besides, can we have some small unit tests?

Wwwsylvia (changes requested):

registry/remote/repository.go
wangxiaoxuan273: Moved it down after getting 404.

registry/remote/repository.go
```diff
@@ -614,6 +615,9 @@ type blobStore struct {
 
 // Fetch fetches the content identified by the descriptor.
 func (s *blobStore) Fetch(ctx context.Context, target ocispec.Descriptor) (rc io.ReadCloser, err error) {
+	if isForeignLayer(target) {
+		return nil, fmt.Errorf("the artifact with foreign layer %s is %w", target.Digest, errdef.ErrUnsupported)
```
Wwwsylvia, line 619: Actually, I think "artifact with foreign layer" looks a bit strange for the `blobStore.Fetch` method... 🤔
What if
```suggestion
		return nil, fmt.Errorf("%s: %s: foreign layer is unsupported: %w", target.Digest, target.MediaType, errdef.ErrUnsupported)
```
？

qweeah:
Some repositories may contain foreign layers which pointing to the repository itself, e.g.
```
> oras manifest fetch mcr.microsoft.com/windows/servercore:1607-KB5019964 --pretty --platform windows/amd64
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "config": {
    "mediaType": "application/vnd.docker.container.image.v1+json",
    "size": 787,
    "digest": "sha256:f443e6ff891920258e9c0ce66f7f5472793b26691ea18d7fc87af0472c677430"
  },
  "layers": [
    {
      "mediaType": "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip",
      "size": 4069985900,
      "digest": "sha256:3889bb8d808bbae6fa5a33e07093e65c31371bcf9e4c38c21be6b9af52ad1548",
      "urls": [
        "https://mcr.microsoft.com/v2/windows/servercore/blobs/sha256:3889bb8d808bbae6fa5a33e07093e65c31371bcf9e4c38c21be6b9af52ad1548"
      ]
    },
    {
      "mediaType": "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip",
      "size": 2213621320,
      "digest": "sha256:29226bae1723379e75abf7d7da45bbf67d5d730611c58bf86df104245db30204",
      "urls": [
        "https://mcr.microsoft.com/v2/windows/servercore/blobs/sha256:29226bae1723379e75abf7d7da45bbf67d5d730611c58bf86df104245db30204"
      ]
    }
  ]
}
```
So it's better to return unsupported **only** when a layer is foreign and not found

registry/remote/repository.go
qweeah: The default media type of a blob cannot be foreign, there MUST be an artifact whose manifest is pointing to the to-be-fetched layer.

registry/remote/repository.go
wangxiaoxuan273: changed.

registry/remote/repository.go
wangxiaoxuan273: After discussion, removed the check inside `Fetch()`.

internal/descriptor/descriptor.go
```diff
@@ -45,3 +46,15 @@ func FromOCI(desc ocispec.Descriptor) Descriptor {
 		Size:      desc.Size,
 	}
 }
+
+func IsForeignLayer(desc ocispec.Descriptor) bool {
```
shizhMSFT, line 50: nit: Missing documentation for this method.

copy.go
```diff
@@ -258,6 +259,9 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 func doCopyNode(ctx context.Context, src content.ReadOnlyStorage, dst content.Storage, desc ocispec.Descriptor) error {
 	rc, err := src.Fetch(ctx, desc)
 	if err != nil {
+		if errors.Is(err, errdef.ErrNotFound) && descriptor.IsForeignLayer(desc) {
```
shizhMSFT, line 262: Can we do other way around?
```suggestion
			if descriptor.IsForeignLayer(desc) && errors.Is(err, errdef.ErrNotFound) {
```

content.go
```diff
@@ -273,6 +274,9 @@ func Fetch(ctx context.Context, target ReadOnlyTarget, reference string, opts Fe
 		}
 		rc, err := target.Fetch(ctx, desc)
 		if err != nil {
+			if errors.Is(err, errdef.ErrNotFound) && descriptor.IsForeignLayer(desc) {
```
shizhMSFT, line 277: By analyzing, we don't need to consider foreign layers in `oras.Fetch` as we cannot fetch a foreign layer from a reference string but a descriptor.

Let's revert the changes in the `content.go`.

internal/descriptor/descriptor.go
wangxiaoxuan273: Added the documentation.

content.go
wangxiaoxuan273: Reverted.

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

Issue #373, opened 2022-12-05
feat: support reading OCI layout from tarballs


Wwwsylvia:
We can support reading OCI layout from tarballs once #336 is addressed.

PR #379, opened 2022-12-08
feat: support reading OCI layout from tarballs

Wwwsylvia:
1. Support `oci.NewStorageFromTar`
2. Support `oci.NewFromTar`

Resolves: #373


internal/fs/tarfs/tarfs.go
```diff
@@ -0,0 +1,172 @@
...
+
+	data, err := io.ReadAll(tr)
+	if err != nil {
+		return nil, err
+	}
+	f := &entryFile{
+		header: entry.header,
+		r:      bytes.NewReader(data),
+	}
+	return f, nil
```
shizhMSFT, line 93: We should not dump all data to the memory.

internal/fs/tarfs/tarfs.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// Stat returns a FileInfo describing the file.
+// If there is an error, it should be of type *PathError.
+func (tfs *TarFS) Stat(name string) (fs.FileInfo, error) {
+	entry, err := tfs.getEntry(name)
+	if err != nil {
+		return nil, err
+	}
+	return entry.header.FileInfo(), nil
```
shizhMSFT, line 103: nit:
```suggestion
	return entry.Stat()
```

shizhMSFT (approved):
LGTM

PR #380, opened 2022-12-08
feat: skip foreign layers on `oras.Copy`

### LINKED ISSUE #319: feat: improve the error message if non-distributable(foreign) layer is not supported

shizhMSFT:
Resolves #319
Related to #377

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -26,6 +26,7 @@ import (
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
 	"oras.land/oras-go/v2/internal/cas"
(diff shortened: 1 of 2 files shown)


Wwwsylvia (approved):
LGTM

PR #384, opened 2022-12-12
fix: use proxy when copying node with successors

### LINKED ISSUE #383: node not found when using oras.Copy with certain option
When using `oras.Copy` to copy the below DAG, if `FindSuccessors` is provided and successors of `a` is obtained without fetching , then a not found error will be thrown when copying node `a`.
```
   +-> b
a -+-> c
   +-> d
```

qweeah:
Resolves #383 
Signed-off-by: Billy Zha <jinzha1@microsoft.com>

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -219,8 +219,8 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 		}
 		successors = removeForeignLayers(successors)
 
-		// handle leaf nodes
 		if len(successors) == 0 {
+			// handle leaf nodes
 			exists, err = proxy.Cache.Exists(ctx, desc)
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


copy.go
```diff
@@ -250,7 +250,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 		if err := region.Start(); err != nil {
 			return err
 		}
-		return copyNode(ctx, proxy.Cache, dst, desc, opts)
+		return copyNode(ctx, proxy, dst, desc, opts)
```
shizhMSFT, line 253: If the content is not in the proxy cache, shouldn't copy from the source directly just like the copy operations for leaf node?

qweeah: Copying from the source is the safiest choice, but in that case `proxy` won't be used at all, is what we expected?

qweeah: A manifest needs not to be cached if its successors are externally provided.

shizhMSFT (approved):
LGTM

PR #394, opened 2022-12-23
feat!: Support `Tags` for package `content/oci`

### LINKED ISSUE #391: Enhance `content/oci` for content discovery
Given an OCI layout, the current implementation of `oras-go` has no means to [discover content](https://github.com/opencontainers/distribution-spec/blob/f72e5010d8441a94de8b35ebbe505b827d05fee2/conformance/README.md#content-discovery) so that users cannot copy or extract any information from the OCI layout unless they know the entry point.

Proposal:
1. Refactor https://github.com/oras-project/oras-go/blob/258bfba7524104a5a76a597184ca0d038f4e3086/registry/repository.go#L51-L66 of the [registry.Repository](https://pkg.go.dev/oras.land/oras-go/v2/registry#Repository) interface as `TagFinder`.
2. Let [content/oci.Store](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0-rc.6/content/oci#Store) and [content/oci.ReadOnlyStore](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0-rc.6/content/oci#ReadOnlyStore) implement `TagFinder`.

shizhMSFT:
Resolves #391


registry/repository.go
```diff
@@ -100,8 +84,28 @@ type ReferrerFinder interface {
 	Referrers(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error
 }
 
+// TagFinder discovers tags by the tag service.
+type TagFinder interface {
```
qweeah, line 88: Why this interface is added to `registry` package but not `content`?

Wwwsylvia: How about calling it `TagLister`? 🤔
`ReferrerFinder` is to find the referrers of a descirptor.
`TagLister` is to list tags in the repository.
If it is called `TagFinder` , it sounds like it should find tags for a descriptor?

content/oci/oci_test.go
```diff
@@ -1375,6 +1379,108 @@ func TestCopyGraph_OCIToMemory_PartialCopy(t *testing.T) {
 	}
 }
 
+func TestCopyGraph_Tags(t *testing.T) {
```
Wwwsylvia, line 1382: nit: `TestStore_Tags` ?

content/oci/oci_test.go
```diff
@@ -1375,6 +1379,108 @@ func TestCopyGraph_OCIToMemory_PartialCopy(t *testing.T) {
...
+				t.Errorf("ReadOnlyStore.Tags() error = %v", err)
+			}
+		})
+	}
+
+	wantErr := errors.New("expected error")
+	if err := s.Tags(ctx, "", func(got []string) error {
+		return wantErr
+	}); err != wantErr {
+		t.Errorf("ReadOnlyStore.Tags() error = %v, wantErr %v", err, wantErr)
```
Wwwsylvia, line 1480: nit: "Store.Tags() error = ..."?

content/oci/readonlyoci_test.go
```diff
@@ -34,13 +36,17 @@ import (
...
+	"oras.land/oras-go/v2/registry"
 )
 
 func TestReadonlyStoreInterface(t *testing.T) {
 	var store interface{} = &ReadOnlyStore{}
 	if _, ok := store.(oras.ReadOnlyGraphTarget); !ok {
 		t.Error("&Store{} does not conform oras.ReadOnlyGraphTarget")
 	}
+	if _, ok := store.(registry.TagFinder); !ok {
+		t.Error("&Store{} does not conform registry.TagFinder")
```
Wwwsylvia, line 48: nit: Just realized this it should be "&ReadOnlystore{} ..." here 😂

registry/repository.go
shizhMSFT: If we call it `TagLister`, we should re-name `ReferrerFinder` to `ReferrerLister` since it is [listing referrers](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers).

Wwwsylvia (approved):
LGTM

PR #396, opened 2023-01-03
fix!: `Tag()` and `TagN()` return a descriptor

### LINKED ISSUE #395: design: `Tag()` and `TagN()` should return a descriptor
[oras.Tag()](https://pkg.go.dev/oras.land/oras-go/v2#Tag) and [oras.TagN()](https://pkg.go.dev/oras.land/oras-go/v2#TagN) do not return a descriptor. Thus, there is no way to know which content is re-tagged exactly.

This is a design flaw / bug and should be fixed.

shizhMSFT:
Fix #395

### DIFF
--- a/content.go
+++ b/content.go
@@ -82,9 +82,9 @@ func TagN(ctx context.Context, target Target, srcReference string, dstReferences
 		opts.MaxMetadataBytes = defaultTagNMaxMetadataBytes
 	}
 
-	refFetcher, okFetch := target.(registry.ReferenceFetcher)
-	refPusher, okPush := target.(registry.ReferencePusher)
-	if okFetch && okPush {
+	_, isRefFetcher := target.(registry.ReferenceFetcher)
+	_, isRefPusher := target.(registry.ReferencePusher)
+	if isRefFetcher && isRefPusher {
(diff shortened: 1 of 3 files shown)


content.go
```diff
@@ -82,60 +82,43 @@ func TagN(ctx context.Context, target Target, srcReference string, dstReferences
...
-			eg.Go(func(dst string) func() error {
-				return func() error {
-					r := bytes.NewReader(contentBytes)
-					if err := refPusher.PushReference(egCtx, desc, r, dst); err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
-						return fmt.Errorf("failed to tag %s as %s: %w", srcReference, dst, err)
-					}
-					return nil
-				}
-			}(dstRef))
+		if err := tagBytesN(ctx, target, desc, contentBytes, dstReferences, TagBytesNOptions{
```
qweeah, line 111: Can we remove the reference fetcher and reference pusher check and always use `content.FetchBytes` and `content.tagBytesN`?

qweeah: Please resolve it. Looks like the proposal doesn't perform well when the Target is local, e.g. memory target.

Wwwsylvia (approved):
LGTM

Issue #135, opened 2022-04-27
Extended Copy performance improvement


Wwwsylvia:
We can introduce go-routines to speed up the Extended Copy at here:
https://github.com/oras-project/oras-go/blob/f701e60ea36cd64189c66814f72e4cddcb375abe/extendedcopy.go#L102-L106

PR #397, opened 2023-01-10
refactor: optimize performance for Extended Copy

Wwwsylvia:
Resolves: #135

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -171,23 +171,40 @@ func CopyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 // copyGraph copies a rooted directed acyclic graph (DAG) from the source CAS to
 // the destination CAS with specified caching.
 func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Storage, proxy *cas.Proxy, root ocispec.Descriptor, opts CopyGraphOptions) error {
-	// if FindSuccessors is not provided, use the default one
-	if opts.FindSuccessors == nil {
-		opts.FindSuccessors = content.Successors
-	}
 	// if Concurrency is not set or invalid, use the default concurrency
 	if opts.Concurrency <= 0 {
 		opts.Concurrency = defaultConcurrency
 	}
-	limiter := semaphore.NewWeighted(int64(opts.Concurrency))
-	// track content status
-	tracker := status.NewTracker()
+	return copyGraphWithOptions(ctx, root, copyGraphOptions{
+		CopyGraphOptions: opts,
+		src:              src,
+		dst:              dst,
+		proxy:            proxy,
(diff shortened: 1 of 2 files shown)


copy.go
```diff
@@ -160,28 +165,26 @@ func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, ds
...
 	if opts.FindSuccessors == nil {
 		opts.FindSuccessors = content.Successors
 	}
-	// if Concurrency is not set or invalid, use the default concurrency
-	if opts.Concurrency <= 0 {
-		opts.Concurrency = defaultConcurrency
-	}
-	limiter := semaphore.NewWeighted(int64(opts.Concurrency))
-	// track content status
-	tracker := status.NewTracker()
```
shizhMSFT, line 184: I am thinking if it would be better if we do something like
```go
if proxy == nil {
	// use caching proxy on non-leaf nodes
	if opts.MaxMetadataBytes <= 0 {
		opts.MaxMetadataBytes = defaultCopyMaxMetadataBytes
	}
	proxy = cas.NewProxyWithLimit(src, cas.NewMemory(), opts.MaxMetadataBytes)
}
if limiter == nil {
	// if Concurrency is not set or invalid, use the default concurrency
	if opts.Concurrency <= 0 {
		opts.Concurrency = defaultConcurrency
	}
	limiter = semaphore.NewWeighted(int64(opts.Concurrency))
}
if tracker == nil {
	// track content status
	tracker := status.NewTracker()
}
```
in `copyGraph()`?

copy.go
```diff
@@ -160,28 +165,26 @@ func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, ds
...
+	if opts.Concurrency <= 0 {
+		opts.Concurrency = defaultConcurrency
+	}
 	// use caching proxy on non-leaf nodes
 	if opts.MaxMetadataBytes <= 0 {
 		opts.MaxMetadataBytes = defaultCopyMaxMetadataBytes
 	}
 	proxy := cas.NewProxyWithLimit(src, cas.NewMemory(), opts.MaxMetadataBytes)
-	return copyGraph(ctx, src, dst, proxy, root, opts)
+	return copyGraph(ctx, src, dst, root, proxy, semaphore.NewWeighted(int64(opts.Concurrency)), status.NewTracker(), opts)
```
shizhMSFT, line 177: Based on https://github.com/oras-project/oras-go/pull/397/files#r1065821344, we can do
```suggestion
	return copyGraph(ctx, src, dst, root, nil, nil, nil, opts)
```

copy.go
Wwwsylvia: This makes sense!

shizhMSFT (approved):
LGTM

PR #398, opened 2023-01-11
feat!: retryable http client

### LINKED ISSUE #147: Support for retry?
hey oras developers!

I've been using oras in CI, and I typically need to wrap it in some kind of retry with exponential backoff. To compare to another registry client, when I use docker it will retry pulling blobs, and same with Singularity. How could it work? I suspect there could be some `Retry-After` header returned from a server, and if not, the user could specify a number of retries.  Would this be something oras could support? In practice I have seen a lot of push failures in my CI when using oras, and probably just the fault of the receiving end.

souleb:
fixes #147 

If implemented, this will provide a default http client with retry.

The retry function is an exponential back off 0.25s * 2^n ± 10% and max 5 attempts.

The client is the default client of `auth.Client`

Co-authored-by: Shiwei Zhang <shizh@microsoft.com>
Signed-off-by: Soule BA <bah.soule@gmail.com>

cc @shizhMSFT

### DIFF
--- a/registry/remote/retry/client.go
+++ b/registry/remote/retry/client.go
@@ -16,23 +16,21 @@ limitations under the License.
 package retry
 
 import (
-	"bytes"
-	"io"
 	"net/http"
 	"time"
 )
 
-// // DefaultClient is a client with the default retry policy
+// DefaultClient is a client with the default retry policy.
 var DefaultClient = NewClient()
 
-// NewClient creates an HTTP client with the default retry policy
+// NewClient creates an HTTP client with the default retry policy.
 func NewClient() *http.Client {
 	return &http.Client{
 		Transport: NewTransport(nil),
 	}
 }
 
-// Transport is an HTTP transport with retry policy
+// Transport is an HTTP transport with retry policy.
 type Transport struct {
 	// Base is the underlying HTTP transport to use.
 	// If nil, http.DefaultTransport is used for round trips.
@@ -43,36 +41,51 @@ type Transport struct {
 	Policy func() Policy
 }
 
-// NewTransport creates an HTTP Transport with the default retry policy
+// NewTransport creates an HTTP Transport with the default retry policy.
 func NewTransport(base http.RoundTripper) *Transport {
 	return &Transport{
 		Base: base,
 	}
 }
 
-// RoundTrip executes a single HTTP transaction, returning a Response for the provided Request.
-// It relies on the configured Policy to determine if the request should be retried and to backoff.
+// RoundTrip executes a single HTTP transaction, returning a Response for the
+// provided Request.
+// It relies on the configured Policy to determine if the request should be
+// retried and to backoff.
 func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
 	policy := t.policy()
 	attempt := 0
 	for {
 		resp, respErr := t.roundTrip(req)
-		duration, err := policy.Retry(ctx, attempt, resp, respErr)
+		duration, err := policy.Retry(attempt, resp, respErr)
+		if err != nil {
+			if respErr == nil {
+				resp.Body.Close()
+			}
+			return nil, err
+		}
 		if duration < 0 {
-			return resp, err
+			return resp, respErr
 		}
 
-		// rewind the body if necessary
+		// rewind the body if possible
 		if req.Body != nil {
-			var buf bytes.Buffer
-			if _, err = buf.ReadFrom(resp.Body); err != nil {
-				return resp, err
+			if req.GetBody == nil {
+				// body can't be rewound, so we can't retry
+				return resp, respErr
 			}
-			if err = resp.Body.Close(); err != nil {
-				return resp, err
+			body, err := req.GetBody()
+			if err != nil {
+				// failed to rewind the body, so we can't retry
+				return resp, respErr
 			}
-			req.Body = io.NopCloser(&buf)
+			req.Body = body
+		}
+
+		// close the response body if needed
+		if respErr == nil {
+			resp.Body.Close()
 		}
 
 		timer := time.NewTimer(duration)
--- a/registry/remote/retry/client.go
+++ b/registry/remote/retry/client.go
@@ -1,0 +1,101 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package retry
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"time"
+)
+
+// // DefaultClient is a client with the default retry policy
+var DefaultClient = NewClient()
+
+// NewClient creates an HTTP client with the default retry policy
+func NewClient() *http.Client {
+	return &http.Client{
+		Transport: NewTransport(nil),
+	}
+}
+
+// Transport is an HTTP transport with retry policy
+type Transport struct {
+	// Base is the underlying HTTP transport to use.
+	// If nil, http.DefaultTransport is used for round trips.
+	Base http.RoundTripper
+
+	// Policy returns a retry Policy to use for the request.
+	// If nil, DefaultPolicy is used to determine if the request should be retried.
+	Policy func() Policy
+}
+
+// NewTransport creates an HTTP Transport with the default retry policy
+func NewTransport(base http.RoundTripper) *Transport {
+	return &Transport{
+		Base: base,
+	}
+}
+
+// RoundTrip executes a single HTTP transaction, returning a Response for the provided Request.
+// It relies on the configured Policy to determine if the request should be retried and to backoff.
+func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+	policy := t.policy()
+	attempt := 0
+	for {
+		resp, respErr := t.roundTrip(req)
+		duration, err := policy.Retry(ctx, attempt, resp, respErr)
+		if duration < 0 {
+			return resp, err
+		}
+
+		// rewind the body if necessary
+		if req.Body != nil {
+			var buf bytes.Buffer
+			if _, err = buf.ReadFrom(resp.Body); err != nil {
+				return resp, err
+			}
+			if err = resp.Body.Close(); err != nil {
+				return resp, err
+			}
+			req.Body = io.NopCloser(&buf)
+		}
+
+		timer := time.NewTimer(duration)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return nil, ctx.Err()
+		case <-timer.C:
+		}
+		attempt++
+	}
+}
+
+func (t *Transport) roundTrip(req *http.Request) (*http.Response, error) {
+	if t.Base == nil {
+		return http.DefaultTransport.RoundTrip(req)
+	}
+	return t.Base.RoundTrip(req)
+}
+
+func (t *Transport) policy() Policy {
+	if t.Policy == nil {
+		return DefaultPolicy
+	}
+	return t.Policy()
+}
--- a/registry/remote/retry/policy.go
+++ b/registry/remote/retry/policy.go
@@ -16,10 +16,10 @@ limitations under the License.
 package retry
 
 import (
-	"context"
-	"fmt"
+	"hash/maphash"
 	"math"
 	"math/rand"
+	"net"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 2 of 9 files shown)


registry/remote/retry/client.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+// Transport is an HTTP transport with retry policy
+type Transport struct {
+	// Base is the underlying HTTP transport to use.
+	// If nil, http.DefaultTransport is used.
+	Base http.RoundTripper
+	// Policy returns a retry Policy to use for the request.
+	Policy func() Policy
```
shizhMSFT, line 41: It would be better if we add an empty line between fields. It would be clearer in the godoc.

registry/remote/retry/client.go
```diff
@@ -0,0 +1,95 @@
...
+		Transport: NewTransport(nil),
+	}
+}
+
+// Transport is an HTTP transport with retry policy
+type Transport struct {
+	// Base is the underlying HTTP transport to use.
+	// If nil, http.DefaultTransport is used.
+	Base http.RoundTripper
+	// Policy returns a retry Policy to use for the request.
```
shizhMSFT, line 40: We need to document the default behavior if `Policy` is `nil`.

registry/remote/retry/client.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// RoundTrip executes a single HTTP transaction, returning a Response for the provided Request.
+// It relies on the configured Policy to determine if the request should be retried and to backoff.
+func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := req.Context()
+	policy := t.policy()
+	attempt := 0
+	for {
+		resp, respErr := t.Base.RoundTrip(req)
```
shizhMSFT, line 61: Instead of `t.Base.RoundTrip(req)`, could we have something like
```go
func (t *Transport) roundTrip(req *http.Request) (*http.Response, error) {
    if t.Base == nil {
        return http.DefaultTransport.RoundTrip(req)
    }
    return t.Base.RoundTrip(req)
}
```
Then we call `resp, respErr := t.roundTrip(req)` so that an empty `Transport` is also workable just like other structs in the `remote` package. Maybe, we can even consider to remove `func NewTransport(base http.RoundTripper) *Transport`

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,142 @@
...
+	if resp.StatusCode == 0 || resp.StatusCode >= 500 {
+		return true, fmt.Errorf("unexpected HTTP status %s", resp.Status)
+	}
+
+	return false, nil
+}
+
+// DefaultBackoff is a backoff that uses an exponential backoff with jitter.
+// It uses a base of 250ms, a factor of 2 and a jitter of 10%.
+var DefaultBackoff Backoff = ExponentialBackoff(250*time.Millisecond, 2, 0.1)
```
shizhMSFT, line 68: Nice catch. It should be `ExponentialBackoff` instead of linear backoff

shizhMSFT (commented):
Good work. Just left some initial review comments. I will continue reviewing later.

souleb:
The `examples` tests are failing due to some TLS misconfiguration when I set the `retry.DefaultClient` as `client` for `auth.Client`. Can you tell me where to look?

shizhMSFT:
> The `examples` tests are failing due to some TLS misconfiguration when I set the `retry.DefaultClient` as `client` for `auth.Client`. Can you tell me where to look?

@souleb The root cause of the error is that `retry.Client` relies on the default http transport. Therefore, we need to override the default http transport instead of the client for the mocks.

Please check the following changes:
```diff
diff --git a/example_test.go b/example_test.go
index c1e4c20..82fa126 100644
--- a/example_test.go
+++ b/example_test.go
@@ -182,7 +182,7 @@ func TestMain(m *testing.M) {
                panic(err)
        }
        remoteHost = u.Host
-       http.DefaultClient = httpsServer.Client()
+       http.DefaultTransport = httpsServer.Client().Transport
 
        os.Exit(m.Run())
 }
diff --git a/registry/example_test.go b/registry/example_test.go
index 70f102c..2463667 100644
--- a/registry/example_test.go
+++ b/registry/example_test.go
@@ -65,7 +65,7 @@ func TestMain(m *testing.M) {
                panic(err)
        }
        host = u.Host
-       http.DefaultClient = ts.Client()
+       http.DefaultTransport = ts.Client().Transport
 
        os.Exit(m.Run())
 }
diff --git a/registry/remote/example_test.go b/registry/remote/example_test.go
index bbd292e..370486d 100644
--- a/registry/remote/example_test.go
+++ b/registry/remote/example_test.go
@@ -214,7 +214,7 @@ func TestMain(m *testing.M) {
                panic(err)
        }
        host = u.Host
-       http.DefaultClient = ts.Client()
+       http.DefaultTransport = ts.Client().Transport
 
        os.Exit(m.Run())
 }
```

registry/remote/retry/client.go
```diff
@@ -0,0 +1,101 @@
...
+		if req.Body != nil {
+			var buf bytes.Buffer
+			if _, err = buf.ReadFrom(resp.Body); err != nil {
+				return resp, err
+			}
+			if err = resp.Body.Close(); err != nil {
+				return resp, err
+			}
+			req.Body = io.NopCloser(&buf)
+		}
```
shizhMSFT, line 76: This part is not right. It seems that the code reads the _response body_ of the previous request as the _request body_ of the next request.

To properly rewind the body, `req.GetBody` as documented at [here](https://pkg.go.dev/net/http#Request) is the proper choice.

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+// jitter. The backoff is calculated as:
+//
+//	backoff * factor ^ attempt + rand.Int63n(jitter * backoff)
+//
+// The HTTP response is checked for a Retry-After header. If it is present, the
+// value is used as the backoff duration and jitter is applied.
+func ExponentialBackoff(backoff time.Duration, factor int, jitter float64) Backoff {
+	return func(attempt int, resp *http.Response) time.Duration {
+		// Seed random number generator with nanoseconds
+		rand := rand.New(rand.NewSource(int64(time.Now().Nanosecond())))
```
shizhMSFT, line 94: For the rand seed, we can do it a faster way by internally calling `runtime.fastrand()`.

```suggestion
		var h maphash.Hash
		h.SetSeed(maphash.MakeSeed())
		rand := rand.New(rand.NewSource(int64(h.Sum64())))
```

shizhMSFT: Details see: 
- https://pkg.go.dev/hash/maphash
- https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/hash/maphash/maphash.go;l=257

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+		// check Retry-After
+		if resp != nil && resp.StatusCode == http.StatusTooManyRequests {
+			if v := resp.Header.Get(headerRetryAfter); v != "" {
+				if retryAfter, _ := strconv.ParseInt(v, 10, 64); retryAfter > 0 {
+					return time.Duration(retryAfter + rand.Int63n(int64(jitter*float64(backoff))))
```
shizhMSFT, line 99: Do we require jitter if the retry-after is specified by the server?

shizhMSFT: Again, `Retry-After` is in seconds. Therefore, it should be something like `retryAfter * time.Second`.

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+type Backoff func(attempt int, resp *http.Response) time.Duration
+
+// ExponentialBackoff returns a Backoff that uses an exponential backoff with
+// jitter. The backoff is calculated as:
+//
+//	backoff * factor ^ attempt + rand.Int63n(jitter * backoff)
+//
+// The HTTP response is checked for a Retry-After header. If it is present, the
+// value is used as the backoff duration and jitter is applied.
+func ExponentialBackoff(backoff time.Duration, factor int, jitter float64) Backoff {
```
shizhMSFT, line 91: Can the factor be a float64?

```suggestion
func ExponentialBackoff(backoff time.Duration, factor, jitter float64) Backoff {
```

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Backoff is a function that returns the duration to wait before retrying the
+// request. The attempt, is the next attempt number. The response is the
+// response from the previous request.
+type Backoff func(attempt int, resp *http.Response) time.Duration
+
+// ExponentialBackoff returns a Backoff that uses an exponential backoff with
+// jitter. The backoff is calculated as:
+//
+//	backoff * factor ^ attempt + rand.Int63n(jitter * backoff)
```
shizhMSFT, line 87: Suggesting a new formula for math calculation:
```
temp = backoff * factor ^ attempt
interval = temp * (1 - jitter) + rand.Int63n(2 * jitter * temp)
```

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+	MaxWait:   3 * time.Second,
+	MaxRetry:  5,
+}
+
+// DefaultPredicate is a predicate that retries on 5xx errors, 429 Too Many
+// Requests, 401 Unauthorized and 408 Request Timeout.
+var DefaultPredicate Predicate = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
```
shizhMSFT, line 46: nit: better code style
```suggestion
	if err := ctx.Err(); err != nil {
		return false, err
	}
```

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if err != nil {
+		return false, err
+	}
```
shizhMSFT, line 50: If it is a dial time-out error, we should retry. It is common when the network is not stable.

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+
+	switch resp.StatusCode {
+	case http.StatusUnauthorized:
+		return true, nil
```
shizhMSFT, line 54: We should not retry on `http.StatusUnauthorized`. Expired token should be retried at the auth client level.

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,146 @@
...
+	case http.StatusRequestTimeout, http.StatusTooManyRequests:
+		return true, nil
+	}
+
+	if resp.StatusCode == 0 || resp.StatusCode >= 500 {
+		return true, fmt.Errorf("unexpected HTTP status %s", resp.Status)
```
shizhMSFT, line 60: Since the error is ignored on `true`, should we just `return true, nil`?

registry/remote/retry/policy.go
shizhMSFT, line 46: You may consider
```
	if err != nil {
		return false, err
	}
```
first, and then the `ctx`. If the `ctx` is cancelled, it is most likely error is not nil.

shizhMSFT: Actually, since we check the context in the main `Retry` loop, we don't really need to check the context here.

registry/remote/retry/client.go
souleb: thanks for catching this. Let me rectify that.

registry/remote/retry/policy.go
souleb: The idea of adding the jitter to the `Retry-After` is to avoid a herd.

registry/remote/retry/policy.go
souleb: great, thanks.

registry/remote/retry/policy.go
souleb: TIL 👍

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// MaxRetry is the maximum number of retries.
+	MaxRetry int
+}
+
+// Retry returns the duration to wait before retrying the request.
+// It returns -1 if the request should not be retried.
+func (p *GenericPolicy) Retry(attempt int, resp *http.Response, err error) (time.Duration, error) {
+	if attempt >= p.MaxRetry {
+		return -1, err
```
shizhMSFT, line 138: Should we
```suggestion
		return -1, nil
```
so that we can return the original error from the server side?

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	if ok, err := p.Retryable(resp, err); !ok {
+		return -1, err
+	}
```
shizhMSFT, line 142: We should always check errors before processing any return values.
```suggestion
	if ok, err := p.Retryable(resp, err); err != nil {
		return -1, err
	} else if !ok {
		return -1, nil
	}
```

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,151 @@
...
+	return false, nil
+}
+
+// DefaultBackoff is a backoff that uses an exponential backoff with jitter.
+// It uses a base of 250ms, a factor of 2 and a jitter of 10%.
+var DefaultBackoff Backoff = ExponentialBackoff(250*time.Millisecond, 2, 0.1)
+
+// Policy is a retry policy.
+type Policy interface {
+	// Retry returns the duration to wait before retrying the request.
```
shizhMSFT, line 69: Negative duration implies stop retrying.

registry/remote/retry/client.go
```diff
@@ -0,0 +1,112 @@
...
+				return resp, err
+			} else {
+				if _, err = buf.ReadFrom(body); err != nil {
+					return resp, err
+				}
+				if err = body.Close(); err != nil {
+					return resp, err
+				}
+				req.Body = io.NopCloser(&buf)
+			}
```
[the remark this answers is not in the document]
souleb: Thanks, after reading the code, I see that it does it own buffering.

registry/remote/retry/client.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		// close the response body if necessary
+		if resp != nil {
+			if err := resp.Body.Close(); err != nil {
+				return resp, err
+			}
+		}
```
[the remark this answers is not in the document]
souleb: I see now that `Body` is guaranteed to be non nil. So we close if `respErr == nil` meaning that we have received a response.

registry/remote/retry/client.go
```diff
@@ -0,0 +1,114 @@
...
+		// rewind the body if possible
+		if req.Body != nil {
+			if req.GetBody == nil {
+				// body can't be rewound, so we can't retry
+				return resp, respErr
+			}
+			body, err := req.GetBody()
+			if err != nil {
+				// failed to rewind the body, so we can't retry
+				return resp, err
```
shizhMSFT, line 81: If we can't retry, we should return the original error.
```suggestion
				return resp, respErr
```

registry/remote/retry/policy.go
```diff
@@ -0,0 +1,154 @@
...
+	// It returns -1 if the request should not be retried.
```
shizhMSFT, line 70: nit: Should we state `-1` or a negative value?

registry/remote/retry/client.go
```diff
@@ -0,0 +1,114 @@
...
+*/
+
+package retry
+
+import (
+	"net/http"
+	"time"
+)
+
+// DefaultClient is a client with the default retry policy
```
Wwwsylvia, line 23: nit: Should we have periods (.) at the end of the Go doc comments like we did in other places?

registry/remote/retry/policy.go
souleb: Yes negative value is better based on how we handle it.

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

PR #399, opened 2023-01-12
build: add dependabot for `v1` branch, add GitHub Actions

lucacome:
- Target `v1` branch for golang updates until `v2` is GA
- Add check for GitHub Actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,3 +18,14 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+# update go dependencies for v1 branch
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -18,3 +18,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    target-branch: "v1"
```
shizhMSFT, line 21: Is it possible to target to both `v1` and `main`?

lucacome: looks like it's not possible unfortunately 😞

shizhMSFT: Will merging this PR to the `v1` branch work?

lucacome: I think it only works from the main branch, but let me do tests in another repo.

lucacome: @shizhMSFT Turns out you can specify the same config for multiple branches 🎉

shizhMSFT (approved):
LGTM

Issue #402, opened 2023-01-12
Bug: file store fails to extract folder containing symbolic links


Wwwsylvia:
Original issue: https://github.com/oras-project/oras/issues/740

PR #411, opened 2023-01-16
fix!: fix error when extracting folder containing symbolic links

### LINKED ISSUE #404: Bug: OCI store and File store should record an absolute path instead of a relative path for root

Wwwsylvia:
Fixes part of #404
Fixes: #402
BREAKING CHANGE: `file.New()` and other constructors now return an error


shizhMSFT (approved):
LGTM

PR #412, opened 2023-01-16
fix!: use absolute path for OCI root

### LINKED ISSUE #404: Bug: OCI store and File store should record an absolute path instead of a relative path for root

Wwwsylvia:
Fixes: #404 
BREAKING CHANGE: `oci.NewStorage()` now returns an error

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -70,16 +70,25 @@ func New(root string) (*Store, error) {
 
 // NewWithContext creates a new OCI store.
 func NewWithContext(ctx context.Context, root string) (*Store, error) {
+	rootAbs, err := filepath.Abs(root)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


content/oci/oci.go
```diff
@@ -70,16 +70,25 @@ func New(root string) (*Store, error) {
 
 // NewWithContext creates a new OCI store.
 func NewWithContext(ctx context.Context, root string) (*Store, error) {
+	rootAbs, err := filepath.Abs(root)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve absolute path for %s: %w", root, err)
+	}
```
shizhMSFT, line 76: This check is unnecessary since `NewStorage()` has already done the check.

Wwwsylvia: This is still needed because `indexPath` relies on the root path. See line 85 and line 91.

shizhMSFT (approved):
LGTM

Issue #242, opened 2022-07-18
v1.2.0 is incompatible with docker v20.10.17


Justin-W:
There seems to be a mismatch between the oras-go v1.2.0 `go.mod` and some of the code.

- https://github.com/oras-project/oras-go/blob/v1.2.0/go.mod#L10
  * This requires `github.com/docker/docker v20.10.17+incompatible`
- https://github.com/oras-project/oras-go/blob/v1.2.0/pkg/auth/docker/login_tls.go#L181
  * this line actually uses the `func PingV2Registry` signature matching the [docker v22 API](https://github.com/moby/moby/blob/v22.06.0-beta.0/registry/auth.go#L173)
  * it seems like it _should_ use the `func PingV2Registry` signature matching the [docker v20 API](https://github.com/moby/moby/blob/v20.10.17/registry/auth.go#L203)

As a result, running `go get -u ./...` on my project (which indirectly depends on oras-go via kops v1.24.0 (both kops v1.24.0 and kops latest still vendor github.com/docker/docker v20) results in a compilation error:
```
../../../../../../pkg/mod/oras.land/oras-go@v1.2.0/pkg/auth/docker/login_tls.go:181:30: assignment mismatch: 3 variables but registry.PingV2Registry returns 2 values
```

It seems that either:
1. oras-go v1.2.0 should still be using the docker v20-compatible API [here](https://github.com/oras-project/oras-go/blob/v1.2.0/pkg/auth/docker/login_tls.go#L181); _OR_
2. oras-go v1.2.0 should be depending on a docker v22 version [here](https://github.com/oras-project/oras-go/blob/v1.2.0/go.mod#L10), _**and**_ the [v1.2.0 release notes](https://github.com/oras-project/oras-go/releases/tag/v1.2.0) should have included a breaking change notice about incompatibility with docker v20.

shizhMSFT:
Nice catch. /cc @souleb.

shizhMSFT:
@jasminetMSFT Could you investigate this issue?

jasminetMSFT:
Sure, please assign this to me.

jasminetMSFT:
@Justin-W In oras-go v1.2.0, https://github.com/oras-project/oras-go/blob/v1.2.0/pkg/auth/docker/login_tls.go#L181 `PingV2Registry` returns three parameters which matches `func PingV2Registry` in [docker v20 API](https://github.com/moby/moby/blob/v20.10.17/registry/auth.go#L203) 
[...]

PR #428, opened 2023-02-03
bump: update docker and docker/cli

benschlueter:
since there is a small api change, one has to do the update manually


shizhMSFT:
@benschlueter Thanks for contributing. I am wondering if this change will cause an issue similar to #242

shizhMSFT (approved):
LGTM. We need to mention this version bump up in the release note as concerned by #242. /cc @Wwwsylvia

benschlueter:
Sorry, overlooked the initial mention. 
Yes, probably. I just go the issue from my go.mod file, since helm imported this lib and when dependable-bot upgraded my package it caused the incompatibility with this lib.

PR #431, opened 2023-02-05
fix: Broken tests for mac

TerryHowe:
These tests are broken for mac because `/var/folder` is a sym link to `/private/var/folder`.

Signed-off-by: Terry Howe <tlhowe@amazon.com>


sabre1041 (approved):
LGTM. Existing `main` branch fails on osx. tested successfully with PR contents

shizhMSFT (approved):
LGTM

Issue #425, opened 2023-02-02
Shift Go Support Window to [1.19, 1.20] 


shizhMSFT:
[Go 1.20](https://go.dev/blog/go1.20) is released with [release notes](https://go.dev/doc/go1.20)!

It is time to shift the window for go support to `[1.19, 1.20]` by updating

- `go. mod`
- GitHub Actions

Although we will set `1.19` as the minimum version of `oras-go`, it is still possible to compile `oras-go` with go 1.18.

Asmit2952:
Hey @shizhMSFT if this is a good-first-issue, can I work on it?

shizhMSFT:
@Asmit2952 Sure! Thanks for contributing. BTW, we follow the go [release policy](https://go.dev/doc/devel/release#policy) for the support window.

PR #442, opened 2023-02-11
feat: update go support window to [1.19, 1.20]

Asmit2952:
Update Go support window to [1.19, 1.20]

Fixes: #425 

Cc: @shizhMSFT

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.18, 1.19]
+        go-version: [1.19, 1.20]
       fail-fast: true
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [1.18, 1.19]
```
shizhMSFT, line 27: Since it’s yaml, `1.20` is recognized as a number `1.2` instead of a string. Therefore, we should specify it as a string `'1.20'`.

Asmit2952: Will resolve it by including versions in quotes

shizhMSFT (approved):
LGTM

TerryHowe (approved):

Issue #461, opened 2023-03-14
manifest removed unexpectly from the OCI layout `index.json` file


qweeah:
In an `oci.Store`, if a manifest `m` has its entry in the `index.json` file. Using `Store.Tag` with the digest as the reference will remove `m` from `index.json` file.

Run below go file to reproduce the issue. oras-go version: v2.0.1.
```golang
package main

import (
	"context"
	_ "crypto/sha256"
	"fmt"
	"os"
	"path/filepath"
	"strings"

	"github.com/opencontainers/go-digest"
	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
	"oras.land/oras-go/v2/content/oci"
)

func main() {
	// prepare
	root := os.TempDir()
	defer os.RemoveAll(root)
	oci, err := oci.New(root)
	if err != nil {
		panic(err)
	}
	ctx := context.Background()
	manifestContent := `{"schemaVersion":2,"mediaType":"application/vnd.oci.image.manifest.v1+json","config":{"mediaType":"application/vnd.oci.image.config.v1+json","digest":"sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a","size":2},"layers":[]}`
	desc := ocispec.Descriptor{
		Digest:    digest.FromString(manifestContent),
		Size:      int64(len(manifestContent)),
		MediaType: "application/vnd.oci.image.manifest.v1+json",
	}

	// push a manifest
	oci.Push(ctx, desc, strings.NewReader(manifestContent))
	index, err := os.ReadFile(filepath.Join(root, "index.json"))
	if err != nil {
		panic(err)
	}
	fmt.Println(">>>>>>>>>>> after pushing >>>>>>>>>>>>>>")
	fmt.Println(string(index))

	// tag the manifest with its digest
	oci.Tag(ctx, desc, desc.Digest.String())
	index, err = os.ReadFile(filepath.Join(root, "index.json"))
	if err != nil {
		panic(err)
	}
	fmt.Println(">>>>>>>>>>> after tagging >>>>>>>>>>>>>>")
	fmt.Println(string(index))
}
```

FeynmanZhou:
Do we plan to fix it and cut a release of v2.0.2 in ORAS-go? Then ORAS CLI v1.0.0 could bump up the ORAS-go dep to this version.

PR #462, opened 2023-03-15
fix: refactor tagging implementation for OCI Store

Wwwsylvia:
Fixes: #461

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -142,10 +142,13 @@ func (s *Store) Tag(ctx context.Context, desc ocispec.Descriptor, reference stri
 		return fmt.Errorf("%s: %s: %w", desc.Digest, desc.MediaType, errdef.ErrNotFound)
 	}
 
-	if desc.Annotations == nil {
-		desc.Annotations = map[string]string{}
+	if reference != desc.Digest.String() {
+		// add ref name annotation only when reference is not a digest
+		if desc.Annotations == nil {
+			desc.Annotations = map[string]string{}
+		}
+		desc.Annotations[ocispec.AnnotationRefName] = reference
 	}
-	desc.Annotations[ocispec.AnnotationRefName] = reference
 	return s.tag(ctx, desc, reference)
 }
 
@@ -157,6 +160,12 @@ func (s *Store) tag(ctx context.Context, desc ocispec.Descriptor, reference stri
 		if err := s.tagResolver.Tag(ctx, desc, dgst); err != nil {
 			return err
 		}
+
+		// untag the last manifest tagged by reference
+		lastTagged, err := s.tagResolver.Resolve(ctx, reference)
+		if err == nil {
+			s.untag(ctx, lastTagged)
+		}
 	}
 	if err := s.tagResolver.Tag(ctx, desc, reference); err != nil {
 		return err
@@ -165,6 +174,11 @@ func (s *Store) tag(ctx context.Context, desc ocispec.Descriptor, reference stri
 		return s.SaveIndex()
 	}
 	return nil
+}
+
+func (s *Store) untag(ctx context.Context, desc ocispec.Descriptor) error {
+	delete(desc.Annotations, ocispec.AnnotationRefName)
+	return s.tagResolver.Tag(ctx, desc, desc.Digest.String())
 }
 
 // Resolve resolves a reference to a descriptor.
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -32,7 +32,7 @@ import (
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"oras.land/oras-go/v2/content"
 	"oras.land/oras-go/v2/errdef"
-	"oras.land/oras-go/v2/internal/containers"
+	"oras.land/oras-go/v2/internal/container/set"
 	"oras.land/oras-go/v2/internal/descriptor"
 	"oras.land/oras-go/v2/internal/graph"
 	"oras.land/oras-go/v2/internal/resolver"
@@ -267,7 +267,7 @@ func (s *Store) SaveIndex() error {
 	defer s.indexLock.Unlock()
 
 	var manifests []ocispec.Descriptor
-	tagged := make(containers.Set[digest.Digest])
+	tagged := set.New[digest.Digest]()
 	refMap := s.tagResolver.Map()
 
 	// 1. Add descriptors that are associated with tags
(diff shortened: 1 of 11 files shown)


qweeah (changes requested):

content/oci/oci_test.go
```diff
@@ -694,6 +712,141 @@ func TestStore_RepeatTag(t *testing.T) {
...
+		t.Errorf("len(index.Manifests) = %v, want %v", got, want)
+	}
+	gotDesc, err = s.Resolve(ctx, manifestDesc.Digest.String())
+	if err != nil {
+		t.Fatal("Store.Resolve() error =", err)
+	}
+	if !reflect.DeepEqual(gotDesc, manifestDesc) {
+		t.Errorf("Store.Resolve() = %v, want %v", gotDesc, manifestDesc)
+	}
+
```
qweeah, line 810: Shouldn't we recall the `s, err = New(tempDir)` here to reproduce the bug?

content/oci/oci.go
```diff
@@ -269,14 +266,33 @@ func (s *Store) SaveIndex() error {
...
+
 	refMap := s.tagResolver.Map()
+	// 1. Add descriptors identified by tags
+	for ref, desc := range refMap {
+		if ref != desc.Digest.String() {
+			// Note: One descriptor can be associated with multiple tags
+			if desc.Annotations == nil {
+				desc.Annotations = make(map[string]string)
+			}
+			desc.Annotations[ocispec.AnnotationRefName] = ref
```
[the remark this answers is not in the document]
Wwwsylvia: Nice catch!

content/oci/oci.go
```diff
@@ -269,14 +266,33 @@ func (s *Store) SaveIndex() error {
 	defer s.indexLock.Unlock()
 
 	var manifests []ocispec.Descriptor
+	tagged := make(map[digest.Digest]struct{})
```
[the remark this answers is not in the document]
Wwwsylvia: I think using `desc.Digest` is more convenient

content/oci/oci.go
```diff
@@ -269,14 +266,33 @@ func (s *Store) SaveIndex() error {
...
+			if desc.Annotations == nil {
+				desc.Annotations = make(map[string]string)
+			}
+			desc.Annotations[ocispec.AnnotationRefName] = ref
+			manifests = append(manifests, desc)
+			// Mark digest as tagged for deduplication for step 2
+			tagged[desc.Digest] = struct{}{}
+		}
+	}
+	// 2. Add descriptors identified by digests
```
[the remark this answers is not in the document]
Wwwsylvia: Updated with some adjustments

content/oci/oci_test.go
Wwwsylvia: Tested in `TestStore_ExistingStore` (line 1078-1082).

content/oci/oci.go
```diff
@@ -269,14 +267,35 @@ func (s *Store) SaveIndex() error {
...
+	}
+	// 2. Add descriptors that are not associated with any tag
 	for ref, desc := range refMap {
-		if ref == desc.Digest.String() && desc.Annotations[ocispec.AnnotationRefName] != "" {
-			// skip saving desc if ref is a digest and desc is tagged
-			continue
+		if ref == desc.Digest.String() {
+			// skip tagged ones since they have been added in step 1
+			if !tagged.Contains(desc.Digest) {
+				manifests = append(manifests, desc)
```
Wwwsylvia, line 294: Just realized that, here we assume desc does not contain the annotation ref name. However, there might be issues if we load the index and save it back. We may need to strip the ref name when loading the index.

internal/containers/set.go
shizhMSFT, line 1: Can we make it `internal/container/set/set.go` just like those in the built-in libraries?

Ref: https://pkg.go.dev/container/list

internal/containers/set.go
```diff
@@ -0,0 +1,30 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package containers
+
+// Set represents a set data structure.
+type Set[T comparable] map[T]struct{}
```
shizhMSFT, line 19: We need a constructor for this `Set`.

content/oci/oci.go
Wwwsylvia: Fixed.

content/oci/readonlyoci.go
```diff
@@ -224,3 +224,29 @@ func listTags(ctx context.Context, tagResolver *resolver.Memory, last string, fn
 
 	return fn(tags)
 }
+
+// stripAnnotationRefName strips the AnnotationRefName from the annotation map
+// of desc.
+func stripAnnotationRefName(desc ocispec.Descriptor) ocispec.Descriptor {
```
shizhMSFT, line 230: It is better to name the function to `deleteAnnotationRefName`.

content/oci/readonlyoci.go
```diff
@@ -224,3 +224,29 @@ func listTags(ctx context.Context, tagResolver *resolver.Memory, last string, fn
...
+
+	l := len(desc.Annotations)
+	if l == 1 {
+		// the annotation map contains ref name only
+		desc.Annotations = nil
+		return desc
+	}
+
+	// the annotation map contains more than 1 key
+	annotations := make(map[string]string, l-1)
```
shizhMSFT, line 244: It would be better to have a meaning variable name other than `l`.

```suggestion
	size := len(desc.Annotations) - 1
	if size == 0 {
		desc.Annotations = nil
		return desc
	}
	annotations := make(map[string]string, size)
```

content/oci/oci.go
```diff
@@ -269,14 +267,35 @@ func (s *Store) SaveIndex() error {
...
 	for ref, desc := range refMap {
-		if ref == desc.Digest.String() && desc.Annotations[ocispec.AnnotationRefName] != "" {
-			// skip saving desc if ref is a digest and desc is tagged
-			continue
+		if ref == desc.Digest.String() {
+			// skip tagged ones since they have been added in step 1
+			if !tagged.Contains(desc.Digest) {
+				manifests = append(manifests, stripAnnotationRefName(desc))
+			}
 		}
```
shizhMSFT, line 296: nit: code style
```suggestion
		if ref == desc.Digest.String() && !tagged.Contains(desc.Digest) {
			manifests = append(manifests, deleteAnnotationRefName(desc))
		}
```

shizhMSFT (approved):
LGTM

qweeah (approved):
LGTM

Issue #457, opened 2023-03-07
OCI Store resolver returns plain descriptor


carabasdaniel:
When resolving a descriptor from the OCI store, the returned descriptor is stripped down to contain only the media type, digest and size: https://github.com/oras-project/oras-go/blob/96a37c2b359ac1305f70dc31b28c789688d77d0f/content/oci/oci.go#L185

Is there any particular reason to not have a [complete](https://github.com/opencontainers/image-spec/blob/819aa940cae7c067a8bf89b1745d3255ddaaba1d/specs-go/v1/descriptor.go) descriptor returned ?

Wwwsylvia:
Returning the complete descriptor could be tricky in some scenarios.

I would like to take the chance to discuss this use case:
Suppose there are two descriptors pointing to the same manifest with different tags (`v1.0` and `latest`) in the `index.json`:

```json
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [
      {
        "mediaType": "application/vnd.oci.image.manifest.v1+json",
        "size": 7143,
        "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
        "annotations": {
          "org.opencontainers.image.ref.name": "v1.0",
          "foo":"bar"
        }
      },
      {
        "mediaType": "application/vnd.oci.image.manifest.v1+json",
        "size": 7143,
        "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
        "annotations": {
          "org.opencontainers.image.ref.name": "latest",
          "foo":"bar"
        }
      }
    ]
  }
```
If we resolve the descriptor by its digest:
```go
desc, err := oci.Resolve(ctx, "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f")
```
What should be the result of `oci.Resolve()`?

Should it be the descriptor tagged by `v1.0`?
```json
     {
        "mediaType": "application/vnd.oci.image.manifest.v1+json",
        "size": 7143,
        "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
        "annotations": {
          "org.opencontainers.image.ref.name": "v1.0",
          "foo":"bar"
        }
      },
```
Or the descriptor tagged by `latest`?
```json
      {
[...]

PR #468, opened 2023-03-21
fix: OCI store resolve returns full descriptor

oanatmaria:
Based on discussion on https://github.com/oras-project/oras-go/issues/457, this PR returns a full descriptor on Resolve when the method it's called by tag and it returns a plain descriptor when the method it's called by digest.

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -165,7 +165,11 @@ func (s *Store) tag(ctx context.Context, desc ocispec.Descriptor, reference stri
 	return nil
 }
 
-// Resolve resolves a reference to a descriptor.
+// Resolve resolves a reference to a descriptor. If the reference to be resolved
+// is a tag, the returned descriptor will be a full descriptor declared by
+// github.com/opencontainers/image-spec/specs-go/v1. If the reference is a
+// digest the returned descriptor will be a plain descriptor (containing only
+// the digest, media type and size).
 func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descriptor, error) {
 	if reference == "" {
 		return ocispec.Descriptor{}, errdef.ErrMissingReference
@@ -180,7 +184,12 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
 		}
 		return ocispec.Descriptor{}, err
 	}
-	return descriptor.Plain(desc), nil
+
+	if reference == desc.Digest.String() {
+		return descriptor.Plain(desc), nil
+	}
+
+	return desc, nil
 }
 
 // Predecessors returns the nodes directly pointing to the current node.
--- a/content/oci/oci_test.go
+++ b/content/oci/oci_test.go
@@ -39,6 +39,7 @@ import (
 	"oras.land/oras-go/v2/content/memory"
 	"oras.land/oras-go/v2/errdef"
 	"oras.land/oras-go/v2/internal/cas"
+	"oras.land/oras-go/v2/internal/descriptor"
 	"oras.land/oras-go/v2/registry"
 )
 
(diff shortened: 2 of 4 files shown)


content/oci/oci_test.go
```diff
@@ -1088,7 +1162,7 @@ func TestStore_ExistingStore(t *testing.T) {
 	}
 
 	// test resolving index root by tag
-	gotDesc, err := anotherS.Resolve(ctx, tag)
+	gotDesc, err := anotherS.Resolve(ctx, string(indexRoot.Digest))
```
Wwwsylvia, line 1165: Is it doing the same thing as line 1174?

content/oci/oci.go
```diff
@@ -180,7 +180,12 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
 		}
 		return ocispec.Descriptor{}, err
 	}
-	return descriptor.Plain(desc), nil
+
+	if digest.DigestRegexpAnchored.MatchString(reference) {
```
Wwwsylvia, line 184: Can we just check
```go
if reference == desc.Digest.String()
```
?

content/oci/oci_test.go
```diff
@@ -452,6 +452,80 @@ func TestStore_ContentBadPush(t *testing.T) {
...
+	}
+
+	tempDir := t.TempDir()
+	s, err := New(tempDir)
+	if err != nil {
+		t.Fatal("New() error =", err)
+	}
+	ctx := context.Background()
+
+	err = s.Push(ctx, desc, strings.NewReader(string(content)))
```
Wwwsylvia, line 472: nit: You may want to use
```go
bytes.NewReader(content)
```

Wwwsylvia (commented):
Do you want to update the same for [`ReadOnlyStore`](https://github.com/oras-project/oras-go/blob/main/content/oci/readonlyoci.go#L87) as well?

content/oci/oci_test.go
oanatmaria: Yes, you are right. I've updated the tests in order for all the checks to be relevant again.

oanatmaria:
> Hi @oanatmaria , thanks for the contribution! Do you want to update the same for [`ReadOnlyStore`](https://github.com/oras-project/oras-go/blob/main/content/oci/readonlyoci.go#L87) as well?

Hi @Wwwsylvia,

Thank you for the review. I've made the change for the readonlystore as well.

content/oci/oci_test.go
```diff
@@ -452,6 +452,86 @@ func TestStore_ContentBadPush(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("error tagging descriptor error = %v, wantErr %v", err, false)
+	}
+
+	resolvedDescr, err := s.Resolve(ctx, string(desc.Digest))
+	if err != nil {
+		t.Errorf("error resolving descriptor error = %v, wantErr %v", err, false)
+	}
+
+	if reflect.DeepEqual(resolvedDescr, desc) {
```
Wwwsylvia, line 530:
```suggestion
	if !reflect.DeepEqual(resolvedDescr, desc) {
```
?

content/oci/oci_test.go
```diff
@@ -452,6 +452,86 @@ func TestStore_ContentBadPush(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("error resolving descriptor error = %v, wantErr %v", err, false)
+	}
+
+	if !reflect.DeepEqual(resolvedDescr, desc) {
+		t.Errorf("Store.Resolve() = %v, want %v", resolvedDescr, desc)
+	}
+
+	// descriptor resolved by tag should have annotations
+	if !reflect.DeepEqual(resolvedDescr.Annotations, annotations) {
```
Wwwsylvia, line 493: Annotations should have been checked by `reflect.DeepEqual()`

content/oci/oci_test.go
oanatmaria: This is the scenario for resolving by digest so a plain descriptor is returned thus they will not be equal

content/oci/oci_test.go
oanatmaria: I removed the check as it's indeed redundant

content/oci/oci_test.go
Wwwsylvia: Oh right... I think it would be better to make a plain descriptor without annotation, and compare the resolved descriptor with the plain descriptor.

oanatmaria: updated

oanatmaria:
> Also wondering if we should call out the behavior difference between resolving by tag and resolving by digest in the doc comment of the `Resolve()` function? @oanatmaria What do you think?

I've updated the comments. Please let me know if they are correctly formulated.

Wwwsylvia (approved):
LGTM

Issue #454, opened 2023-03-04
oci.Store and oci.Store should implement content.Deleter


qweeah:
`oci.Store` and `oci.Store` should implement `content.Deleter` to provide base delete operations on `index.json` without garbage collection.

wangxiaoxuan273:
I'd like to work on this.

PR #470, opened 2023-03-23
feature: Implement deleter

carabasdaniel:
This should close: https://github.com/oras-project/oras-go/issues/454


content/file/file.go
```diff
@@ -349,6 +349,29 @@ func (s *Store) Resolve(ctx context.Context, ref string) (ocispec.Descriptor, er
...
+	}
+	if !exists {
+		return errdef.ErrNotFound
+	}
+	path, _ := s.digestToPath.Load(target.Digest)
+	pathString, ok := path.(string)
+	if !ok {
+		return errdef.ErrInvalidDigest
+	}
+	err = os.RemoveAll(s.absPath(pathString))
```
shizhMSFT, line 368: If `digest` cannot be found in `digestToPath`, we should head to the `fallbackStorage` directly.

content/oci/storage.go
```diff
@@ -106,6 +106,22 @@ func (s *Storage) Push(_ context.Context, expected ocispec.Descriptor, content i
 	return nil
 }
 
+// Delete removes the descriptor blob from the storage
```
shizhMSFT, line 109: The documentation is required to be reworded.
```suggestion
// Delete removes the target blob.
```

content/oci/oci.go
```diff
@@ -192,6 +192,22 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
 	return desc, nil
 }
 
+func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
```
Wwwsylvia, line 195: One tricky thing about deleting is that we need to maintain the predecessors relationship as well. See [Store.Push](https://github.com/oras-project/oras-go/blob/main/content/oci/oci.go#L116-L117) and [Store.Predecessors](https://github.com/oras-project/oras-go/blob/main/content/oci/oci.go#L195-L200).

content/file/file.go
```diff
@@ -349,6 +349,29 @@ func (s *Store) Resolve(ctx context.Context, ref string) (ocispec.Descriptor, er
 	return s.resolver.Resolve(ctx, ref)
 }
 
+func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
```
Wwwsylvia, line 352: Do we have use cases for deleting content in file store?

content/oci/storage.go
```diff
@@ -106,6 +106,22 @@ func (s *Storage) Push(_ context.Context, expected ocispec.Descriptor, content i
...
+	path, err := blobPath(target.Digest)
+	if err != nil {
+		return fmt.Errorf("%s: %s: %w", target.Digest, target.MediaType, errdef.ErrInvalidDigest)
+	}
+	fullpath := filepath.Join(s.root, path)
+
+	// check if the target content does not exists in the blob directory.
+	if _, err := os.Stat(fullpath); os.IsNotExist(err) {
+		return err
+	}
```
shizhMSFT, line 120: There is a pitfall in `os.Stat()`. A bug free implementation would be 
```suggestion
	// check if the target content does not exists in the blob directory.
	if _, err := os.Stat(fullpath); err != nil {
		if os.IsNotExist(err) {
			// return not exist error
		}
		// return generic error
	}
```

content/oci/storage.go
```diff
@@ -106,6 +106,22 @@ func (s *Storage) Push(_ context.Context, expected ocispec.Descriptor, content i
 	return nil
 }
 
+// Delete removes the descriptor blob from the storage
+func (s *Storage) Delete(ctx context.Context, target ocispec.Descriptor) error {
+	path, err := blobPath(target.Digest)
+	if err != nil {
+		return fmt.Errorf("%s: %s: %w", target.Digest, target.MediaType, errdef.ErrInvalidDigest)
+	}
+	fullpath := filepath.Join(s.root, path)
```
shizhMSFT, line 115: nit: you may want to rename `fullpath` to `target` in terms of consistency.

content/oci/storage.go
shizhMSFT: Since `os.Remove()` checks the file existence, we don't need to do `os.Stat()` in advance.

content/oci/oci.go
```diff
@@ -192,6 +192,22 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
...
+	resolvers := s.tagResolver.Map()
+	for reference, desc := range resolvers {
+		if content.Equal(desc, target) {
+			s.tagResolver.Remove(reference)
+		}
+	}
```
shizhMSFT, line 201: What if some is pushing and deleting at the same time? There is a race condition.

shizhMSFT (changes requested):

content/oci/storage.go
carabasdaniel: renamed to targetpath

content/file/file.go
```diff
@@ -349,6 +349,29 @@ func (s *Store) Resolve(ctx context.Context, ref string) (ocispec.Descriptor, er
...
+		return ErrStoreClosed
+	}
+	exists, err := s.Exists(ctx, target)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return errdef.ErrNotFound
+	}
+	path, _ := s.digestToPath.Load(target.Digest)
```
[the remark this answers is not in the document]
carabasdaniel: removed deleter from file store as currently we don't have any use cases for it https://github.com/oras-project/oras-go/pull/470#discussion_r1151709421

content/storage.go
```diff
@@ -43,6 +43,7 @@ type Pusher interface {
 type Storage interface {
 	ReadOnlyStorage
 	Pusher
+	Deleter
```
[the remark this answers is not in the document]
carabasdaniel: Added DeleteStorage interface as extension of Storage

content/file/file.go
carabasdaniel: https://github.com/oras-project/oras-go/pull/470#discussion_r1151976911

content/storage.go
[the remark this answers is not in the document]
carabasdaniel: https://github.com/oras-project/oras-go/pull/470#discussion_r1151978443

content/oci/oci.go
carabasdaniel: The targetResolver.Map() represent an inconsistent snapshot of the storage content as far as I understand, therefore I'm not sure if a delete of a pushing content would be possible as the tagResolver is actually a syncMap.

content/oci/oci.go
```diff
@@ -192,6 +193,44 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
 	return desc, nil
 }
 
+// Untag removes a reference string from index.
+// reference should be a valid tag (e.g. "latest").
+// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/image-layout.md#indexjson-file
+func (s *Store) Untag(ctx context.Context, descr ocispec.Descriptor, reference string) error {
```
shizhMSFT, line 199: What's the purpose and scenario for `Untag`? If it is not related to this PR, could you create a new issue / PR accordingly?

content/oci/oci.go
```diff
@@ -192,6 +193,44 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Delete removed a target descriptor from index and storage.
+func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
```
shizhMSFT, line 218: Supporting a complete scenario of deletion for the OCI layout is complex as we also need to consider garbage collection but this can be in another PR.

content/oci/oci.go
shizhMSFT: @carabasdaniel Since it is possible that someone is trying to push and delete at the same time, we have to trade-off some performance for the deletion.

Similar to what I have mentioned in other comments, the current `oci.Store` is append-only and thus it has the benefits that multiple push and fetch can happen at the same time without race conditions.

However, adding the functionality of `Delete()` makes the store mutable, and we have to change the underlying implementation from sync maps to a `RWMutex` to ensure consistency. The drawback is that we no longer can push and delete simultaneously.

Overall, I suggest

1. Drop `Delete()` for `oci.Store`.
2. Develop `oci.MutableStore` with `Delete()` support.

/cc @Wwwsylvia

internal/cas/memory_test.go
```diff
@@ -136,3 +136,56 @@ func TestMemoryBadPush(t *testing.T) {
...
+	}
+	if got := len(s.Map()); got != 1 {
+		t.Errorf("Memory.Map() = %v, want %v", got, 1)
+	}
+
+	err = s.Delete(ctx, desc)
+	if err != nil {
+		t.Fatal("Memory.Delete() error =", err)
+	}
+	if got := len(s.Map()); got != 0 {
```
Wwwsylvia, line 171: Should we test `Exists()` again after deleting the desc?

internal/cas/memory_test.go
```diff
@@ -136,3 +136,56 @@ func TestMemoryBadPush(t *testing.T) {
...
+		MediaType: "test",
+		Digest:    digest.FromBytes(content),
+		Size:      int64(len(content)),
+	}
+
+	s := NewMemory()
+	ctx := context.Background()
+
+	err := s.Delete(ctx, desc)
+	if err == nil {
```
Wwwsylvia, line 188: If the desc does not exist in the storage, `Delete` should return `errdef.ErrNotFound`. We can test `errdef.ErrNotFound` here.

content/storage.go
```diff
@@ -45,6 +45,12 @@ type Storage interface {
 	Pusher
 }
 
+// DeleteStorage represents an extension of the Storage interface that includes the Deleter.
+type DeleteStorage interface {
```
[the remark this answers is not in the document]
Wwwsylvia: `MutableStorage` sounds like that the existing data can be modified or deleted. But in our case, the data can be deleted but cannot be modified.
How about `StorageDeleter` or `DeletableStorage`?
BTW, do we really need to define this interface?

content/oci/storage_test.go
```diff
@@ -377,3 +377,27 @@ func TestStorage_Fetch_Concurrent(t *testing.T) {
...
+	s, err := NewStorage(tempDir)
+	if err != nil {
+		t.Fatal("New() error =", err)
+	}
+	ctx := context.Background()
+
+	if err := s.Push(ctx, desc, bytes.NewReader(content)); err != nil {
+		t.Fatal("Storage.Push() error =", err)
+	}
+	err = s.Delete(ctx, desc)
```
Wwwsylvia, line 399: How do we verify if the content is deleted or not?

content/oci/oci.go
Wwwsylvia: > Overall, I suggest
Drop Delete() for oci.Store.
Develop oci.MutableStore with Delete() support.

It makes sense.

@carabasdaniel I would also like to learn your use cases of deleting content in OCI layout. Is it for GC purpose? Could you share a little bit?

content/oci/oci.go
carabasdaniel: Hi @shizhMSFT,

The scenario where we want to use Untag, in our use case happens when a user has the same image with two different tags and want to remove one from the local OCI store. 

In essence we want to be able to keep the same manifest, config and layers with different references and be able to remove one of them if desired. 

Do you think this would be a valid use-case scenario for you ? 

I will gladly extract this in a separate PR if it makes sense to be on its own.

content/oci/oci.go
carabasdaniel: Hi @Wwwsylvia, @shizhMSFT,

We have a tool used for building OCI images from rego policies: [policy CLI](https://github.com/opcr-io/policy).

We have been using the previous major version of oras-go to build our containers and interact with upstream OCI compliant registries. In our case we want the user to be able to build and easily remove containers from the local OCI storage, thus the need to have a delete mechanism in place, keeping the same workflow feel as the docker CLI. 

At the moment we use a fork that includes these changes in our v0.2.0 release candidate versions of the policy CLI and it seems to work with our current needs. 

I think having a different store might make sense.

Wwwsylvia: Got it. Thank you for sharing this!

sparr:
I find myself preparing to implement similar functionality as part of https://github.com/awslabs/soci-snapshotter and want to check in on this PR. I would much prefer to contribute upstream and then consume that. Is this still under development? Would additional effort be welcome?

content/oci/oci.go
sparr: I agree that this deserves to be its own PR. It's a valid use case, but not necessary to implement deletion.

content/oci/oci_test.go
```diff
@@ -1993,3 +2042,114 @@ func equalDescriptorSet(actual []ocispec.Descriptor, expected []ocispec.Descript
 	}
 	return true
 }
+
+func TestStore_Delete(t *testing.T) {
```
sparr, line 2046: Can some/most of this be deduplicated with the memory_test TestStoreDelete with similar contents?

content/oci/storage.go
```diff
@@ -106,6 +106,19 @@ func (s *Storage) Push(_ context.Context, expected ocispec.Descriptor, content i
...
 }
 
+// Delete removes the target blob.
+func (s *Storage) Delete(ctx context.Context, target ocispec.Descriptor) error {
+	path, err := blobPath(target.Digest)
+	if err != nil {
+		return fmt.Errorf("%s: %s: %w", target.Digest, target.MediaType, errdef.ErrInvalidDigest)
+	}
+	targetPath := filepath.Join(s.root, path)
+	if _, err := os.Stat(targetPath); errors.Is(err, os.ErrNotExist) {
```
rogpeppe, line 116: This Stat call doesn't appear to be necessary. Can't we just check the result of the os.Remove call instead?

content/oci/oci.go
sparr: @Wwwsylvia for additional use case info... I am working on https://github.com/awslabs/soci-snapshotter which uses https://pkg.go.dev/oras.land/oras-go/content to keep content in the local filesystem. Our artifacts are conceptually similar to image and layer metadata; we store one artifact per image we handle, and one artifact per layer in those images, with one-way references (via a list of digests equivalent to image layers) from artifacts of the first type to artifacts of the second type.

content/storage.go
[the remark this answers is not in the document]
sparr:
```golang
type StorageWithDelete interface {
    Storage
    Deleter
}
```
This is from my current efforts in the same direction as this PR. If you don't define this interface, folks like me will just have to do it themselves, and our naming divergence will make it harder to find usage examples in the wild.

content/oci/storage.go
```diff
@@ -106,6 +106,19 @@ func (s *Storage) Push(_ context.Context, expected ocispec.Descriptor, content i
...
+		return errdef.ErrNotFound
+	}
```
TerryHowe, line 118:
```suggestion
```

content/oci/oci_test.go
```diff
@@ -926,6 +926,55 @@ func TestStore_TagByDigest(t *testing.T) {
...
+	err = s.Untag(ctx, resolvedDescr, ref)
+	if err != nil {
+		t.Errorf("error untagging descriptor error = %v, wantErr %v", err, false)
+	}
+
+	if len(s.tagResolver.Map()) > 0 {
+		t.Error("tagresolver map should be empty")
+	}
+}
+
```
TerryHowe, line 977:
```suggestion
```

content/oci/oci.go
```diff
@@ -192,6 +193,47 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
...
+
+	if s.AutoSaveIndex {
+		err := s.SaveIndex()
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
TerryHowe, line 215:
```suggestion
```

TerryHowe (changes requested):
Sounds like maybe two other changes needed here as well as the deletes

carabasdaniel:
Merged suggested changes and rebased on top of latest upstream. 
If we merge this I can open another PR that adds the untag feature. 

Thanks for the help @TerryHowe.

sparr:
@carabasdaniel There appear to be two outstanding comments referring to needed functionality for updating `graph` and `resolver` when content is deleted.

My limited understanding is that graph is used for predecessor/successor (aka parent/child) lookups and traversal of content, and resolver is used for looking up content by tags.

Wwwsylvia:
> My limited understanding is that graph is used for predecessor/successor (aka parent/child) lookups and traversal of content and is populated in Store.Push where it calls graph.Index and during loadIndex when it loads index.json and then calls graph.IndexAll, and resolver is used for looking up content by tags and is populated by Tag (which may require something like your Untag to undo although I haven't looked far enough to determine if it would meet that need)

Yes it's right.

carabasdaniel:
Hi @sparr @Wwwsylvia,

Currently my bandwidth is very limited so I'm not sure when I'll be able to come back to this. For our current needs we are using a [fork ](https://github.com/opcr-io/oras-go) that has this change and the UnTag change merged. 

In addition to the recent changes I made here I also prepared an [untag branch](https://github.com/opcr-io/oras-go/tree/untag) that would build on top of this PR for the `resolver`. 

I would appreciate any help to get this set implemented/merged.

shizhMSFT (commented):
As it is technically challenging to add `Delete()` to the existing `*oci.Store`, as mentioned in https://github.com/oras-project/oras-go/pull/470#discussion_r1153496806, I would suggest developing a new store named `*oci.DeletableStore` where the functionality of `Predecessors()` can be dropped so that the difficulty of the development should be dramatically reduced.

It should also meet the requirements of https://github.com/opcr-io/policy as it does not leverage the functionality of `Predecessors()`.

sparr:
Speaking only for myself, an implementation of a new Store type with Delete functionality but without Predecessors would not fit my current use case. My current solution is to delete files then recreate index.json then recreate the content store so that the graph will be reindexed from scratch.

shizhMSFT:
The main technical challenge here is that the current `*oci.Store` implementation is optimized for the concurrent use of append-only store in terms of performance. It is non-trivial to modify it to have the delete functionality.

To unblock this PR attempt, @wangxiaoxuan273 could you prototype a `*oci.DeletableStore` with both the functionality of `Delete()` and `Predecessors()`? We can have performance penalty, but we have to make sure that all public methods of `*oci.DeletableStore` MUST be go-routine safe since it is by design for all `Target`s.

wangxiaoxuan273:
> Where is this going. Any progress on the `oci.DeletableStore` ?

No progress so far. I'm still in the stage of investigating. But it's likely that this PR will not be continued, and we will open a new one to fix this issue.

Wwwsylvia:
> Where is this going. Any progress on the `oci.DeletableStore` ?

FYI, here is the PR authored by @wangxiaoxuan273: #614

PR #487, opened 2023-04-19
Make `errSkipDesc` a publicly available error

Noxsios:
With this change, `oras.ErrSkipDesc` can be used during `PreCopy` to do custom filtering during copy operations.

This error is needed to be public as `errors.New` will always create a new error instance, even if the text is the same, so you cannot utilize `errors.New("skip descriptor")`.

ex (psuedo code):

```go

        only := []string{"hello.txt", "world.txt"}
	copyOpts.PreCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
		if len(only) > 0 {
			if !utils.sliceContains(only, desc.Annotations[ocispec.AnnotationTitle]) {
				return oras.ErrSkipDesc
			}
		}
		return nil
	}
```

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -37,8 +37,8 @@ import (
 // defaultConcurrency is the default value of CopyGraphOptions.Concurrency.
 const defaultConcurrency int = 3 // This value is consistent with dockerd and containerd.
 
-// errSkipDesc signals copyNode() to stop processing a descriptor.
-var errSkipDesc = errors.New("skip descriptor")
+// ErrSkipDesc signals copyNode() to stop processing a descriptor.
+var ErrSkipDesc = errors.New("skip descriptor")
 
 // DefaultCopyOptions provides the default CopyOptions.
 var DefaultCopyOptions CopyOptions = CopyOptions{
@@ -281,7 +281,7 @@ func doCopyNode(ctx context.Context, src content.ReadOnlyStorage, dst content.St
 func copyNode(ctx context.Context, src content.ReadOnlyStorage, dst content.Storage, desc ocispec.Descriptor, opts CopyGraphOptions) error {
 	if opts.PreCopy != nil {
 		if err := opts.PreCopy(ctx, desc); err != nil {
-			if err == errSkipDesc {
+			if err == ErrSkipDesc {
 				return nil
 			}
 			return err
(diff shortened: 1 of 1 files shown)


Noxsios:
> Is there any other data that should be in this error like #482

AFAIK, no.

qweeah:
@Noxsios IMHO, it's better to do this work in `FindSuccessors`?

```golang
only := []string{"hello.txt", "world.txt"}
copyOpts.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
  nodes, err := func Successors(ctx, fetcher, desc)
  if err != nil {
    return nil, err
  }
  if len(only) == 0 {
    return nodes
  }
  var ret []ocispec.Descriptor
  for _, n := range nodes {
    if utils.sliceContains(only, n.Annotations[ocispec.AnnotationTitle]) {
      append(ret, n)     
    }
  }
  return ret
})
```

Wwwsylvia:
Hi @Noxsios, can you check if [@qweeah's suggestion](https://github.com/oras-project/oras-go/pull/487#issuecomment-1519255605) works for you? 
My concern is that `ErrSkipDesc` might be easily misused by users.

Noxsios:
@Wwwsylvia Agreed. I was unaware that `FindSuccessors` could mutate the fetched manifest before layers were pulled. I will test and close if it works as intended.

Noxsios:
Looks like that works! Final function looks similar to this in case others want the same behavior.

```go
// where optional is a slice of strings
// estimatedBytes is the result of calculating the total size of a OCI artifact
		alwaysPull := []string{"zarf.yaml", "checksums.txt", "zarf.yaml.sig"}
		toPull = append(optional, alwaysPull...)
		copyOpts.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
			nodes, err := content.Successors(ctx, fetcher, desc)
			if err != nil {
				return nil, err
			}
			var ret []ocispec.Descriptor
			for _, node := range nodes {
				if utils.SliceContains(toPull, node.Annotations[ocispec.AnnotationTitle]) {
					ret = append(ret, node)
				} else {
					estimatedBytes -= node.Size
				}
			}
			return ret, nil
		}
```

PR #515, opened 2023-05-26
feat: add option to skip referrers GC

qweeah:


TerryHowe (approved):
/lgtm

Those tests are hard to look at.

registry/remote/repository.go
```diff
@@ -102,6 +102,9 @@ type Repository struct {
 	// If less than or equal to zero, a default (currently 4MiB) is used.
 	MaxMetadataBytes int64
 
+	// ReferrersGC specifies whether to delete the dangling referrers index.
+	ReferrersGC bool
```
Wwwsylvia, line 106: 1. Wondering if we should call it `SkipReferrersGC` since we perform GC by default🤔. But if we plan to skip GC by default **in the future**, `ReferrersGC` might be a more appropriate name?
2. Whatever option name and default value we opt to, we should explicitly call out the behavior and the default value in the comment documentation.
3. If the default value is `true`, we should make sure it's correctly set in other places like [NewRegistry](https://github.com/oras-project/oras-go/blob/v2.2.0/registry/remote/registry.go#L53-L65) and [newRepositoryWithOptions](https://github.com/oras-project/oras-go/blob/v2.2.0/registry/remote/repository.go#L140-L153). (See the comment at line108😀)

qweeah: >  But if we plan to skip GC by default in the future, ReferrersGC might be a more appropriate name?

Skip GC by default is breaking to oras-go and can be done on CLI side. Since oras CLI is the only user of it and don't require the default change, it's better not to introduce this change.

Wwwsylvia: Yeah we shouldn't introduce any breaking change in this PR. I just wanted to discuss the name.

qweeah: `SkipReferrersGC` (with default value false) and `ReferrersGC` (with default value true) both make sense to me, cc @FeynmanZhou @yizha1 for more advice on the naming.

registry/remote/repository.go
```diff
@@ -102,6 +102,13 @@ type Repository struct {
 	// If less than or equal to zero, a default (currently 4MiB) is used.
 	MaxMetadataBytes int64
 
+	// SkipReferrersGC specifies whether to delete the dangling referrers index.
+	//  - If true, the old referrers index will be deleted after the new one
+	//    is successfully uploaded.
+	//  - If false, the old referrers index is kept.
+	// By default, it is enabled (set to true).
```
Wwwsylvia, line 109: It's better to have a reference link here.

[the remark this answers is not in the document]
qweeah: resolved

qweeah: What link is expected to be put here? There is no spec mentioning how to do referrers index GC.

Wwwsylvia: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#pushing-manifests-with-subject and https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#deleting-manifests？

qweeah: added as you suggested

registry/remote/registry_test.go
```diff
@@ -178,6 +178,7 @@ func TestRegistry_Repository(t *testing.T) {
 		t.Fatalf("NewRegistry() error = %v", err)
 	}
 	reg.PlainHTTP = true
+	reg.SkipReferrersGC = false
```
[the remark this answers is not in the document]
Wwwsylvia: why remove this test case?🤔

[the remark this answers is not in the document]
qweeah: oh, that must be a mistake. updated

Wwwsylvia (approved):
LGTM

registry/remote/repository.go
```diff
@@ -102,6 +102,15 @@ type Repository struct {
 	// If less than or equal to zero, a default (currently 4MiB) is used.
 	MaxMetadataBytes int64
 
+	// SkipReferrersGC specifies whether to delete the dangling referrers index.
+	//  - If false, the old referrers index will be deleted after the new one
+	//    is successfully uploaded.
+	//  - If true, the old referrers index is kept.
+	// By default, it is disabled (set to false). See also:
+	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#pushing-manifests-with-subject
+	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#deleting-manifests
```
shizhMSFT, line 111: Need to mention that this option is only applicable to referrers tag schema

registry/remote/repository.go
```diff
@@ -102,6 +102,18 @@ type Repository struct {
...
+	// SkipReferrersGC specifies whether to delete the dangling referrers index.
+	//  - If false, the old referrers index will be deleted after the new one
+	//    is successfully uploaded.
+	//  - If true, the old referrers index is kept.
+	// By default, it is disabled (set to false).
+	// This option only applicable when referrers API is not available and
+	// tag schema is utilized to manage the referrers. See also:
+	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema
+	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#pushing-manifests-with-subject
+	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#deleting-manifests
```
shizhMSFT, line 114: We can simplify the description.
```suggestion
	// SkipReferrersGC specifies whether to delete the dangling referrers index when referrers tag schema is utilized.
	//  - If false, the old referrers index will be deleted after the new one
	//    is successfully uploaded.
	//  - If true, the old referrers index is kept.
	// By default, it is disabled (set to false). See also:
	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#referrers-tag-schema
	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#pushing-manifests-with-subject
	//  - https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#deleting-manifests
```

qweeah: resolved

shizhMSFT (approved):
LGTM

PR #526, opened 2023-06-20
fix: improve errors for `ReadAll` and `loadIndexfile`

### LINKED ISSUE #432: oras copy error message is not clear enough
I manully modified the `size` field of the index.json of OCI layout file to be an invalid value and got this error message.

```
oras cp --from-oci-layout out:latest --to-oci-layout out3:latest -d
Error: read failed: unexpected EOF
```

index.json file
```jsonc
{
  "schemaVersion": 2,
  "manifests": [
    {
      "digest": "sha256:ff43fa9e6ee876230e679f2ff7d2c61f3c43ca437d3cf46b93c28c2231ec9773",
      "size": 12401, // invalid size
      "annotations": {
        "io.containerd.image.name": "docker.io/library/latest:latest",
        "org.opencontainers.image.created": "2023-01-30T02:26:00Z",
        "org.opencontainers.image.ref.name": "latest"
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    }
  ]
}
```
It looks like the error message is not clear enough.

sparr:
NewWithContext refers to both the `oci-layout` and `index.json` as "OCI Image Layout", this PR fixes the latter.

ReadAll is a deeper change that will improve error reporting for a lot of cases, at the cost of error volume. Some or all of that change could happen higher up to satisfy this particular issue. I think the expected/actual size output belongs here. There seem to be more than a few code paths in this and other repos that end up calling ReadAll multiple times without reporting which call produced an error so I think it's appropriate to also output the digest here, but that's less compelling.

The Verify comment clears up confusion I encountered while working on this issue. "verifies size" implies checking for both too big AND too small, but it only checks if the file is bigger (or reader is longer) than expected.

Fixes #432

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -102,7 +102,7 @@ func NewWithContext(ctx context.Context, root string) (*Store, error) {
 		return nil, fmt.Errorf("invalid OCI Image Layout: %w", err)
 	}
 	if err := store.loadIndexFile(ctx); err != nil {
-		return nil, fmt.Errorf("invalid OCI Image Layout: %w", err)
+		return nil, fmt.Errorf("invalid OCI Image Index: %w", err)
 	}
 
 	return store, nil
--- a/content/reader.go
+++ b/content/reader.go
@@ -70,7 +70,7 @@ func (vr *VerifyReader) Read(p []byte) (n int, err error) {
 	return
 }
 
-// Verify verifies the read content against the size and the digest.
+// Verify checks for remaining unread content and verifies the read content against the digest
 func (vr *VerifyReader) Verify() error {
 	if vr.verified {
 		return nil
@@ -120,7 +120,10 @@ func ReadAll(r io.Reader, desc ocispec.Descriptor) ([]byte, error) {
 	buf := make([]byte, desc.Size)
 
 	vr := NewVerifyReader(r, desc)
-	if _, err := io.ReadFull(vr, buf); err != nil {
+	if n, err := io.ReadFull(vr, buf); err != nil {
+		if errors.Is(err, io.ErrUnexpectedEOF) {
+			return nil, fmt.Errorf("read failed: expected content size of %d, got %d, for digest %s: %w", desc.Size, n, desc.Digest.String(), err)
+		}
 		return nil, fmt.Errorf("read failed: %w", err)
 	}
 	if err := vr.Verify(); err != nil {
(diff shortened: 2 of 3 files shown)


TerryHowe (approved):
/lgtm

content/reader.go
```diff
@@ -120,7 +120,10 @@ func ReadAll(r io.Reader, desc ocispec.Descriptor) ([]byte, error) {
 	buf := make([]byte, desc.Size)
 
 	vr := NewVerifyReader(r, desc)
-	if _, err := io.ReadFull(vr, buf); err != nil {
+	if n, err := io.ReadFull(vr, buf); err != nil {
+		if errors.Is(err, io.ErrUnexpectedEOF) {
+			return nil, fmt.Errorf("read failed, expected content size of %d, got %d, for digest %s: %w", desc.Size, n, desc.Digest.String(), err)
```
shizhMSFT, line 125: nit: just be consistent with the generic error message of `"read failed: %w"` at line 127.
```suggestion
			return nil, fmt.Errorf("read failed: expected content size of %d, got %d, for digest %s: %w", desc.Size, n, desc.Digest.String(), err)
```

content/oci/oci.go
```diff
@@ -97,7 +97,7 @@ func NewWithContext(ctx context.Context, root string) (*Store, error) {
 		return nil, fmt.Errorf("invalid OCI Image Layout: %w", err)
 	}
 	if err := store.loadIndexFile(ctx); err != nil {
-		return nil, fmt.Errorf("invalid OCI Image Layout: %w", err)
+		return nil, fmt.Errorf("invalid OCI Image Index: %w", err)
```
shizhMSFT, line 100: Oh, I just noticed the cause of the confusion and ambiguity here. Thanks for fixing it.

shizhMSFT:
@sparr Could you address the comment and rebase the branch? We are ready to merge 🚀

content/reader.go
sparr: #432 asked for something more specific. The core goal of this PR is to provide more info than line 127 does. Maybe I went overboard with the extra info, but I expect I'd want all three pieces of data if I was trying to fix the problem with my content triggering this behavior.

Wwwsylvia: The error message looks useful!
+1 on changing `read failed,` to `read failed:`

Wwwsylvia:
It looks like the branch is still out-of-date. @sparr Would you mind rebase it?

sparr:
I'm also ok with contributors rebasing the PR.

Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #527, opened 2023-06-28
deps: update docker/docker to v24.0.2 (v1)

dtrudg:
Handle `registry.Service` change from interface to struct.

I've seen https://github.com/oras-project/oras-go/pull/517 and https://github.com/oras-project/oras-go/pull/519 and am not clear why they were closed.

The old docker/docker dependency is blocking updating other dependencies in the containers ecosystem, where oras-go v1 is still in use. Would be grateful if the update could be applied to v1.


shizhMSFT (approved):
LGTM

shizhMSFT:
Note: We should mention this in the release note of the next release. Although there is no API changes in our side, it may still break users if they are using the latest `oras-go v1` along with old versions of `docker/docker`.

Issue #531, opened 2023-07-03
Support Image Index with a `subject`


Wwwsylvia:
The [`subject`](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/image-index.md) field is introduced to Image Index since `image-spec v1.1.0-rc4`.
This is a sub-task of #494.

PR #548, opened 2023-07-10
feat: support Image Index with a subject

Wwwsylvia:
Resolves: #531


internal/manifestutil/parser.go
```diff
@@ -0,0 +1,65 @@
...
+*/
+
+package manifestutil
+
+import (
+	"context"
+	"encoding/json"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+
```
shizhMSFT, line 23: nit: remove empty line

internal/platform/platform.go
```diff
@@ -90,17 +91,20 @@ func SelectManifest(ctx context.Context, src content.ReadOnlyStorage, root ocisp
...
 	case docker.MediaTypeManifest, ocispec.MediaTypeImageManifest:
-		descs, err := content.Successors(ctx, src, root)
+		config, err := manifestutil.Config(ctx, src, root)
 		if err != nil {
 			return ocispec.Descriptor{}, err
 		}
+		if config == nil {
+			// should not happen
+			return ocispec.Descriptor{}, fmt.Errorf("%s: %s: %w", root.Digest, root.MediaType, errdef.ErrUnsupported)
+		}
```
shizhMSFT, line 101: It will never happen. Thus we can remove this block of code. Instead, you should have a comment block for line 94.

Issue #532, opened 2023-07-03
Update `oras.Pack` to comply with `image-spec v1.1.0-rc4`


Wwwsylvia:
We need to update `oras.Pack` to conform to the following requirements introduced by [`image-spec v1.1.0-rc4`](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md#image-manifest):

- Use the [empty descriptor](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md#guidance-for-an-empty-descriptor) for config if none is specified
- Use the [empty descriptor ](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md#guidance-for-an-empty-descriptor)as the single layer if none is specified
- Set the `artifactType` field in the image manifest

https://github.com/oras-project/oras-go/blob/6b5bd4b4372b8ba8f5ec0251ddffba8f360f5cf1/pack.go#L78

PR #550, opened 2023-07-12
feat: add a PackOption to support packing image manifests that conform to image-spec v1.1.0-rc4

Wwwsylvia:
1. Introduce `PackOptions.PackManifestType`
2. Introduce `DefaultPackOptions`

Resolves: #532

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -44,18 +44,18 @@ const (
 type PackManifestType = int
 
 const (
-	// PackManifestTypeImageManifestLegacy represents the OCI Image Manifest
-	// type defined in image-spec v1.1.0-rc2.
+	// PackManifestTypeImageV1_1_0_RC2 represents the OCI Image Manifest type
+	// defined in image-spec v1.1.0-rc2.
 	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
 	//
 	// Deprecated: This type is deprecated and will be removed the next major
 	// version.
-	PackManifestTypeImageManifestLegacy PackManifestType = iota
+	PackManifestTypeImageV1_1_0_RC2 PackManifestType = iota
 
-	// PackManifestTypeImageManifest represents the OCI Image Manifest type
+	// PackManifestTypeImageV1_1_0_RC4 represents the OCI Image Manifest type
 	// defined since image-spec v1.1.0-rc4.
 	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md
-	PackManifestTypeImageManifest
+	PackManifestTypeImageV1_1_0_RC4
 )
 
 var (
@@ -67,7 +67,7 @@ var (
 
 	// ErrMissingArtifactType is returned by [Pack] when artifactType is not
 	// specified and the config media type is set to
-	// "application/vnd.oci.empty.v1+json" (the default config).
+	// "application/vnd.oci.empty.v1+json".
 	ErrMissingArtifactType = errors.New("missing artifact type")
 )
 
@@ -87,7 +87,7 @@ type PackOptions struct {
 
 	// PackManifestType controls which type of manifest to pack.
 	// This option is valid only when PackImageManifest is true.
-	// Default: PackManifestTypeImageManifestLegacy.
+	// Default: PackManifestTypeImageV1_1_0_RC2.
 	PackManifestType PackManifestType
 
 	// ConfigDescriptor is a pointer to the descriptor of the config blob.
@@ -106,19 +106,20 @@ type PackOptions struct {
 // Note that the default options are subject to change in the future.
 var DefaultPackOptions PackOptions = PackOptions{
 	PackImageManifest: true,
-	PackManifestType:  PackManifestTypeImageManifest,
+	PackManifestType:  PackManifestTypeImageV1_1_0_RC4,
 }
 
 // Pack packs the given blobs, generates a manifest for the pack,
 // and pushes it to a content storage.
 //
 //   - If opts.PackImageManifest is true and opts.PackManifestType is
-//     PackManifestTypeImageManifestLegacy,
+//     PackManifestTypeImageV1_1_0_RC2,
 //     artifactType will be used as the the config media type of the image
 //     manifest when opts.ConfigDescriptor is not specified.
 //   - If opts.PackImageManifest is true and opts.PackManifestType is
-//     PackManifestTypeImageManifest, [ErrMissingArtifactType] will be returned
-//     when none of artifactType and opts.ConfigDescriptor is specified.
+//     PackManifestTypeImageV1_1_0_RC4,
+//     [ErrMissingArtifactType] will be returned when none of artifactType and
+//     opts.ConfigDescriptor is specified.
 //
 // If succeeded, returns a descriptor of the manifest.
 func Pack(ctx context.Context, pusher content.Pusher, artifactType string, blobs []ocispec.Descriptor, opts PackOptions) (ocispec.Descriptor, error) {
@@ -127,10 +128,10 @@ func Pack(ctx context.Context, pusher content.Pusher, artifactType string, blobs
 	}
 
 	switch opts.PackManifestType {
-	case PackManifestTypeImageManifestLegacy:
-		return packImageLegacy(ctx, pusher, artifactType, blobs, opts)
-	case PackManifestTypeImageManifest:
-		return packImage(ctx, pusher, artifactType, blobs, opts)
+	case PackManifestTypeImageV1_1_0_RC2:
+		return packImageRC2(ctx, pusher, artifactType, blobs, opts)
+	case PackManifestTypeImageV1_1_0_RC4:
+		return packImageRC4(ctx, pusher, artifactType, blobs, opts)
 	default:
 		return ocispec.Descriptor{}, fmt.Errorf("PackManifestType(%v): %w", opts.PackManifestType, errdef.ErrUnsupported)
 	}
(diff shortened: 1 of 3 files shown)


pack.go
```diff
@@ -178,6 +218,71 @@ func packImage(ctx context.Context, pusher content.Pusher, configMediaType strin
...
+		configData := ocispec.DescriptorEmptyJSON.Data
+		if err := pusher.Push(ctx, configDesc, bytes.NewReader(configData)); err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
+			return ocispec.Descriptor{}, fmt.Errorf("failed to push config: %w", err)
+		}
+	}
+	if artifactType == "" {
+		// artifactType MUST be set when config.mediaType is set to the empty value
+		artifactType = configDesc.MediaType
+		if configDesc.MediaType == ocispec.MediaTypeEmptyJSON {
+			artifactType = MediaTypeUnknownArtifact
```
qweeah, line 242: Why not returning error? Looks like it's not oras-go's job to define a such a default value for artifact type.

Wwwsylvia: You are right! The reason why we used `MediaTypeUnknownArtifact` as the default value for packing Artifact Manifest was because the `artifactType` field was a `SHOULD`. However, it's a MUST for Image Manifest with an empty config.
So we should return an error here.

Wwwsylvia: Updated. Thanks for catching this!

pack.go
```diff
@@ -34,52 +34,102 @@ const (
...
+		return packArtifact(ctx, pusher, artifactType, blobs, opts)
+	}
+
+	switch opts.PackManifestType {
+	case PackManifestTypeImageManifestLegacy:
+		return packImageLegacy(ctx, pusher, artifactType, blobs, opts)
+	case PackManifestTypeImageManifest:
 		return packImage(ctx, pusher, artifactType, blobs, opts)
+	default:
+		return ocispec.Descriptor{}, fmt.Errorf("PackManifestType %v is unsupported: %w", opts.PackManifestType, errdef.ErrUnsupported)
```
shizhMSFT, line 131: nit: the word `unsupported` will be printed twice
```suggestion
		return ocispec.Descriptor{}, fmt.Errorf("PackManifestType(%v): %w", opts.PackManifestType, errdef.ErrUnsupported)
```

pack.go
```diff
@@ -34,52 +34,102 @@ const (
...
 	ConfigDescriptor *ocispec.Descriptor
+
 	// ConfigAnnotations is the annotation map of the config descriptor.
 	// This option is valid only when PackImageManifest is true
 	// and ConfigDescriptor is nil.
 	ConfigAnnotations map[string]string
 }
 
+// DefaultPackOptions provides the default PackOptions.
+var DefaultPackOptions PackOptions = PackOptions{
```
shizhMSFT, line 103: You may document that the default pack options are subject to change in the future.

pack.go
```diff
@@ -178,6 +229,71 @@ func packImage(ctx context.Context, pusher content.Pusher, configMediaType strin
...
+	if opts.ConfigDescriptor != nil {
+		configDesc = *opts.ConfigDescriptor
+	} else {
+		// use the empty descriptor for config
+		configDesc = ocispec.DescriptorEmptyJSON
+		configDesc.Annotations = opts.ConfigAnnotations
+		configData := ocispec.DescriptorEmptyJSON.Data
+		if err := pusher.Push(ctx, configDesc, bytes.NewReader(configData)); err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
+			return ocispec.Descriptor{}, fmt.Errorf("failed to push config: %w", err)
+		}
```
shizhMSFT, line 247: If the `pusher` is a registry/repository/blob store, it will send out two requests to the remote.

Before `push`, we can assert if `pusher` can do `Exists()`. Since the empty blob is likely to exist, it will save one request on average.

pack.go
```diff
@@ -178,6 +229,71 @@ func packImage(ctx context.Context, pusher content.Pusher, configMediaType strin
...
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	if len(layers) == 0 {
+		// use the empty descriptor as the single layer
+		layerDesc := ocispec.DescriptorEmptyJSON
+		layerData := ocispec.DescriptorEmptyJSON.Data
+		if err := pusher.Push(ctx, layerDesc, bytes.NewReader(layerData)); err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
+			return ocispec.Descriptor{}, fmt.Errorf("failed to push empty layer: %w", err)
+		}
```
shizhMSFT, line 267: We can avoid this push if we have already done that for config.

pack.go
```diff
@@ -178,6 +229,71 @@ func packImage(ctx context.Context, pusher content.Pusher, configMediaType strin
...
+	}
+	manifestDesc := content.NewDescriptorFromBytes(ocispec.MediaTypeImageManifest, manifestJSON)
+	// populate ArtifactType and Annotations of the manifest into manifestDesc
+	manifestDesc.ArtifactType = manifest.ArtifactType
+	manifestDesc.Annotations = manifest.Annotations
+	// push manifest
+	if err := pusher.Push(ctx, manifestDesc, bytes.NewReader(manifestJSON)); err != nil && !errors.Is(err, errdef.ErrAlreadyExists) {
+		return ocispec.Descriptor{}, fmt.Errorf("failed to push manifest: %w", err)
+	}
+	return manifestDesc, nil
```
shizhMSFT, line 294: This code block seems duplicated and can be refactored

pack.go
```diff
@@ -34,52 +34,102 @@ const (
...
+	// PackManifestTypeImageManifestLegacy represents the OCI Image Manifest
+	// type defined in image-spec v1.1.0-rc2.
+	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
+	PackManifestTypeImageManifestLegacy PackManifestType = iota
+
+	// PackManifestTypeImageManifest represents the OCI Image Manifest type
+	// defined since image-spec v1.1.0-rc4.
+	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md
+	PackManifestTypeImageManifest
+)
```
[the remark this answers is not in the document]
Wwwsylvia: When the spec reaches GA, do we want to change `PackManifestTypeV11RC4` to `PackManifestTypeV11`? or something like that?

pack.go
```diff
@@ -34,52 +34,102 @@ const (
...
+	// AnnotationArtifactCreated or AnnotationCreated is provided, but its value
+	// is not in RFC 3339 format.
+	// Reference: https://www.rfc-editor.org/rfc/rfc3339#section-5.6
+	ErrInvalidDateTimeFormat = errors.New("invalid date and time format")
+
+	// ErrMissingArtifactType is returned by [Pack] when artifactType is not
+	// specified and the config media type is set to
+	// "application/vnd.oci.empty.v1+json" (the default config).
+	ErrMissingArtifactType = errors.New("missing artifact type")
+)
```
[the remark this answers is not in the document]
Wwwsylvia: I think they are specific to `Pack`.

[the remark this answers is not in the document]
Wwwsylvia: But I don't think `ErrMissingArtifactType` will be used in other places. 🤔

pack.go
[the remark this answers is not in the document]
Wwwsylvia: How about `PackManifestTypeV1_1_0_RC2` and `PackManifestTypeV1_1_0_RC4`?

pack.go
Wwwsylvia: Good point!

pack.go
```diff
@@ -34,52 +34,102 @@ const (
...
 	MediaTypeUnknownArtifact = "application/vnd.unknown.artifact.v1"
 )
 
-// ErrInvalidDateTimeFormat is returned by Pack() when
-// AnnotationArtifactCreated or AnnotationCreated is provided, but its value
-// is not in RFC 3339 format.
-// Reference: https://www.rfc-editor.org/rfc/rfc3339#section-5.6
-var ErrInvalidDateTimeFormat = errors.New("invalid date and time format")
+// PackManifestType represents the manifest type used for [Pack].
+type PackManifestType = int32
```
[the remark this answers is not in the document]
Wwwsylvia: Updated.

pack.go
```diff
@@ -34,52 +34,102 @@ const (
...
+	PackManifestType:  PackManifestTypeImageManifest,
+}
+
 // Pack packs the given blobs, generates a manifest for the pack,
 // and pushes it to a content storage.
 //
-// When opts.PackImageManifest is true, artifactType will be used as the
-// the config descriptor mediaType of the image manifest.
+//   - If opts.PackImageManifest is true and opts.PackManifestType is
+//     PackManifestTypeImageManifestLegacy (the default value),
```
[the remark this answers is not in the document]
Wwwsylvia: Fixed.

pack.go
```diff
@@ -34,52 +34,102 @@ const (
...
-	//   - If false, pack an artifact manifest.
+	//   - If true, pack an OCI Image Manifest
+	//   - If false, pack an OCI Artifact Manifest
 	// Default: false.
 	PackImageManifest bool
+
+	// PackManifestType controls which type of manifest to pack.
+	// This option is valid only when PackImageManifest is true.
+	// Default: PackManifestTypeImageManifestLegacy.
+	PackManifestType
```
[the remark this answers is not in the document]
Wwwsylvia: Good catch!

pack.go
Wwwsylvia: Refactored.

pack.go
```diff
@@ -44,7 +44,7 @@ const (
 )
 
 // PackManifestType represents the manifest type used for [Pack].
```
TerryHowe, line 46: Seems like we wouldn't want to make this change

shizhMSFT: I think Sylvia introduced this is for backward compatibility. That is, we should not introduce any breaking changes in `oras-go v2`.

TerryHowe: I think I was looking at a diff between commits, ignore

pack.go
```diff
@@ -161,20 +205,103 @@ func packImage(ctx context.Context, pusher content.Pusher, configMediaType strin
...
+	if ok {
+		var err error
+		exists, err = ros.Exists(ctx, desc)
+		if err != nil {
+			return fmt.Errorf("failed to check existence: %s: %s: %w", desc.Digest.String(), desc.MediaType, err)
+		}
+	}
+	if exists {
+		return nil
+	}
```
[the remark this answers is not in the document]
Wwwsylvia: Nice catch!

Issue #534, opened 2023-07-03
Update the implementation of pushing manifest


Wwwsylvia:
To comply with [`distribution-spec v1.1.0-rc.2`](https://github.com/opencontainers/distribution-spec/releases/tag/v1.1.0-rc.2) and [the upcoming `distribution-spec v1.1.0-rc3`](https://github.com/opencontainers/distribution-spec/pull/433)

PR #557, opened 2023-07-25
feat: update the implementation of pushing manifest for distribution-spec v1.1.0-rc3

Wwwsylvia:
1. Support pushing image index with a subject
2. Check the "OCI-Subject" header

Resolves: #534


shizhMSFT (approved):
LGTM

Issue #469, opened 2023-03-22
Feature Request: Present Warning header to the end user


shizhMSFT:
The PR https://github.com/opencontainers/distribution-spec/pull/393 has been merged to the `main` branch of the distribution spec.

We need to plan and update the `oras-go` related code accordingly.

We may create a `oras.land/oras-go/v2/registry/remote/warn` package to implement the spec.

PR #560, opened 2023-07-31
feat: report warnings

Wwwsylvia:
Add `Repository.HandleWarning(Warning)` that handles encountered warning headers.

Resolves: #469


registry/remote/warning.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type Warning struct {
+	WarningHeader
+	Reference     registry.Reference
+	RequestMethod string
+	RequestPath   string
+}
+
+func parseWarningHeader(header string) (WarningHeader, error) {
```
shizhMSFT, line 49: We don't care about the error details thus we can just return `(WarningHeader, bool)`

registry/remote/warning.go
```diff
@@ -0,0 +1,72 @@
...
+	Code  int
+	Agent string
+	Text  string
+}
+
+type Warning struct {
+	WarningHeader
+	Reference     registry.Reference
+	RequestMethod string
+	RequestPath   string
```
shizhMSFT, line 46: Are those required? How does `kubectl` handle that?

If it is required, can it be replaced by `Request *http.Request` to provide flexibility?

registry/remote/repository.go
```diff
@@ -234,6 +236,28 @@ func (r *Repository) client() Client {
...
+func (r *Repository) do(req *http.Request) (*http.Response, error) {
+	if r.HandleWarning == nil {
+		return r.client().Do(req)
+	}
+
+	resp, err := r.client().Do(req)
+	if err != nil {
+		return nil, err
+	}
+	warningHeaders := parseWarningHeaders(resp.Header[headerWarning])
```
shizhMSFT, line 248: Headers are case insensitive. You may consider [`Header.Values()`](https://pkg.go.dev/net/http#Header.Values).

registry/remote/repository.go
```diff
@@ -234,6 +236,28 @@ func (r *Repository) client() Client {
...
+	for _, wh := range warningHeaders {
+		warning := Warning{
+			WarningHeader: wh,
+			Reference:     r.Reference,
+			RequestMethod: req.Method,
+			RequestPath:   req.URL.Path,
+		}
+		r.HandleWarning(warning)
+	}
```
shizhMSFT, line 257: This part seems duplicated. Can it be refactored?

registry/remote/warning.go
Wwwsylvia: No these are not required and not handled by `kubectl`. I just thought these information could be useful. 🤔

registry/remote/repository.go
Wwwsylvia: Good catch!

registry/remote/repository.go
Wwwsylvia: Refactored.

registry/remote/warning.go
Wwwsylvia: Removed for now. We can still add it back in the future if needed.

registry/remote/warning.go
Wwwsylvia: I think it's better to return an error although we don't check it for now. It's useful for unit test debugging and can be used in logging (if we have logging in the future).

registry/remote/warning.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func parseWarningHeader(header string) (WarningHeader, error) {
+	matches := warningRegexp.FindStringSubmatch(header)
```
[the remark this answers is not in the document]
Wwwsylvia: Tried another way.

Wwwsylvia:
Next step: #562

registry/remote/warning.go
```diff
@@ -0,0 +1,101 @@
...
+// handleWarningHeaders parses the warning headers and handles the parsed
+// warnings using handleWarning.
+func handleWarningHeaders(headers []string, handleWarning func(Warning)) {
+	for _, h := range headers {
+		if value, err := parseWarningHeader(h); err == nil {
+			// ignore warnings in unexpected formats
+			warning := Warning{
+				WarningValue: value,
+			}
+			handleWarning(warning)
```
shizhMSFT, line 98: nit:
```suggestion
			handleWarning(Warning{
				WarningValue: value,
			})
```

shizhMSFT (approved):
LGTM with suggestions

registry/remote/warning.go
Wwwsylvia: Updated😀

shizhMSFT (approved):
LGTM

PR #565, opened 2023-08-07
fix: pack should not set `artifactType` when config is specified

Wwwsylvia:
Related issue: #532 
Related discussion: https://github.com/oras-project/oras/issues/1011#issuecomment-1664614098


shizhMSFT (approved):
LGTM

Issue #568, opened 2023-08-09
Re-visit the deprecation design of oras.Pack


Wwwsylvia:
Recently we introduced a new `PackOption` called [PackManifestTypeImageV1_1_0_RC2 ](https://github.com/oras-project/oras-go/blob/main/pack.go#L56) for [`oras.Pack`](https://github.com/oras-project/oras-go/blob/main/pack.go#L132) but marked it as deprecated (See #532).
This might not be appropriate as it will cause linter error on the caller side.
We need to re-visit the design for `oras.Pack`.

PR #570, opened 2023-08-21
refactor: add `PackManifest` and deprecate `Pack`  

Wwwsylvia:
This PR refactors `oras.Pack` that was updated by #532.

1. Move the support of Image Manifest `v1.1.0-rc4` to `PackManifest`
2. Deprecate `Pack`

Resolves: #568

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -31,10 +31,12 @@ import (
 )
 
 const (
-	// MediaTypeUnknownConfig is the default mediaType used for [Pack] when
-	// PackOptions.PackImageManifest is true and PackOptions.PackManifestType
-	// is PackManifestTypeImageV1_1_0_RC2 and PackOptions.ConfigDescriptor
-	// is not specified.
+	// MediaTypeUnknownConfig is the default config mediaType used
+	//   - for [Pack] when PackOptions.PackImageManifest is true and
+	//     PackOptions.ConfigDescriptor is not specified.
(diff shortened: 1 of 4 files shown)


shizhMSFT (approved):
LGTM

shizhMSFT (approved):
LGTM

pack.go
```diff
@@ -43,38 +44,85 @@ const (
...
+	PackManifestVersion1_0 PackManifestVersion = 1
+
+	// PackManifestVersion1_1_RC4 represents the OCI Image Manifest defined
+	// in image-spec v1.1.0-rc4.
+	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md
+	PackManifestVersion1_1_RC4 PackManifestVersion = 2
+)
+
+// PackManifestOptions contains parameters for [PackManifest].
+type PackManifestOptions struct {
```
qweeah, line 77: Why `artifactType` is not put in this struct?

Wwwsylvia: Two reasons:
1. We encourage users to set this value
2. This value must be set when config is empty

Wwwsylvia: `PackManifestOptions` contains optional parameters only.

Wwwsylvia: Maybe we should mention `optional` in the doc.

shizhMSFT (approved):
LGTM

PR #572, opened 2023-08-23
feat: validate input media type for `oras.PackManifest`

Wwwsylvia:
Resolves: #571

PR #573, opened 2023-08-23
refactor: refactor unexported pack methods

Wwwsylvia:
I'm splitting PR #572 into 2 parts, this is the first part.

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -100,12 +100,12 @@ type PackManifestOptions struct {
 // of the manifest to be packed is determined by packManifestVersion
 // (Recommended value: PackManifestVersion1_1_RC4).
 //
-//   - If packManifestVersion is [PackManifestVersion1_1_RC4],
+//   - If packManifestVersion is [PackManifestVersion1_1_RC4]:
 //     artifactType MUST NOT be empty unless opts.ConfigDescriptor is specified.
-//   - If packManifestVersion is [PackManifestVersion1_0],
-//     artifactType will be used as the the config media type unless
-//     opts.ConfigDescriptor is specified. If artifactType is empty,
-//     "application/vnd.unknown.config.v1+json" will be used.
+//   - If packManifestVersion is [PackManifestVersion1_0]:
+//     when opts.ConfigDescriptor is nil, artifactType will be used as the
+//     config media type; when opts.ConfigDescriptor is NOT nil,
+//     artifactType will be ignored.
 //
 // If succeeded, returns a descriptor of the packed manifest.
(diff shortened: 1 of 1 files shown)


pack.go
```diff
@@ -100,12 +100,12 @@ type PackManifestOptions struct {
...
+//   - If packManifestVersion is [PackManifestVersion1_1_RC4]:
 //     artifactType MUST NOT be empty unless opts.ConfigDescriptor is specified.
-//   - If packManifestVersion is [PackManifestVersion1_0],
-//     artifactType will be used as the the config media type unless
-//     opts.ConfigDescriptor is specified. If artifactType is empty,
-//     "application/vnd.unknown.config.v1+json" will be used.
+//   - If packManifestVersion is [PackManifestVersion1_0]:
+//     when opts.ConfigDescriptor is nil, artifactType will be used as the
+//     config media type; when opts.ConfigDescriptor is NOT nil,
+//     artifactType will be ignored.
```
shizhMSFT, line 108: The new documentation is not accurate. If `opts.ConfigDescriptor` is `nil` and `artifactType` is empty, "application/vnd.unknown.config.v1+json" is used.

pack.go
```diff
@@ -233,6 +228,50 @@ func packManifestV1_1_RC2(ctx context.Context, pusher content.Pusher, configMedi
...
 		Layers:      opts.Layers,
+		Annotations: annotations,
+	}
+	return pushManifest(ctx, pusher, manifest, manifest.MediaType, manifest.Config.MediaType, manifest.Annotations)
+}
+
+// packManifestV1_1_RC2 packs an image manifest as defined in image-spec
+// v1.1.0-rc2.
+// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
+func packManifestV1_1_RC2(ctx context.Context, pusher content.Pusher, configMediaType string, layers []ocispec.Descriptor, opts PackOptions) (ocispec.Descriptor, error) {
```
shizhMSFT, line 239: Let's see if we can do some refactoring in the next PR.

pack.go
Wwwsylvia: I was wondering if it's worth mention the empty artifactType behavior. We don't encourage users to leave `artifactType` empty, right?

shizhMSFT: Yes, we don’t encourage users to have empty `artifactType`. However, we need to be honest to documenting the behavior if the users insist to leave it empty.

shizhMSFT (approved):
LGTM

Issue #571, opened 2023-08-23
Should validate input artifactType and config mediaType for `oras.PackManifest`


Wwwsylvia:
We should validate the input `artifactType` and `ConfigDescriptor.MediaType` for [`oras.PackManifest`](https://github.com/oras-project/oras-go/blob/main/pack.go#L111)as per [image-spec](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/descriptor.md) .

PR #574, opened 2023-08-24
feat: validate input media type for `oras.PackManifest `

Wwwsylvia:
Resolves: #571

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -208,19 +208,17 @@ func packManifestV1_0(ctx context.Context, pusher content.Pusher, artifactType s
 	var configDesc ocispec.Descriptor
 	if opts.ConfigDescriptor != nil {
 		if err := validateMediaType(opts.ConfigDescriptor.MediaType); err != nil {
-			return ocispec.Descriptor{}, fmt.Errorf("invalid config mediaType format: %s: %w", opts.ConfigDescriptor.MediaType, err)
+			return ocispec.Descriptor{}, fmt.Errorf("invalid config mediaType format: %w", err)
 		}
 		configDesc = *opts.ConfigDescriptor
 	} else {
-		if artifactType != "" {
-			if err := validateMediaType(artifactType); err != nil {
-				return ocispec.Descriptor{}, fmt.Errorf("invalid artifactType format: %s: %w", artifactType, err)
-			}
-		} else {
+		if artifactType == "" {
 			artifactType = MediaTypeUnknownConfig
+		} else if err := validateMediaType(artifactType); err != nil {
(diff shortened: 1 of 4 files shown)


pack.go
```diff
@@ -369,6 +384,22 @@ func pushManifest(ctx context.Context, pusher content.Pusher, manifest any, medi
 	return manifestDesc, nil
 }
 
+// pushCustomConfig generates and pushes a custom config blob.
+func pushCustomConfig(ctx context.Context, pusher content.Pusher, mediaType string, annotations map[string]string) (ocispec.Descriptor, error) {
```
shizhMSFT, line 388: It is not custom config but legacy empty config.

pack.go
```diff
@@ -190,28 +199,30 @@ func packArtifact(ctx context.Context, pusher content.Pusher, artifactType strin
...
-		// push config
-		if err := pushIfNotExist(ctx, pusher, configDesc, configBytes); err != nil {
-			return ocispec.Descriptor{}, fmt.Errorf("failed to push config: %w", err)
+		if artifactType != "" {
+			if err := validateMediaType(artifactType); err != nil {
+				return ocispec.Descriptor{}, fmt.Errorf("invalid artifactType format: %s: %w", artifactType, err)
+			}
+		} else {
+			artifactType = MediaTypeUnknownConfig
+		}
```
shizhMSFT, line 221:
```suggestion
		if artifactType == "" {
			artifactType = MediaTypeUnknownConfig
		} else if err := validateMediaType(artifactType); err != nil {
			return ocispec.Descriptor{}, fmt.Errorf("invalid artifactType format: %s: %w", artifactType, err)
		}
```

pack.go
```diff
@@ -392,3 +423,11 @@ func ensureAnnotationCreated(annotations map[string]string, annotationCreatedKey
 	copied[annotationCreatedKey] = now.Format(time.RFC3339)
 	return copied, nil
 }
+
+// validateMediaType validates the format of mediaType.
+func validateMediaType(mediaType string) error {
+	if !mediaTypeRegexp.MatchString(mediaType) {
+		return errdef.ErrInvalidMediaType
```
shizhMSFT, line 430: It seems can be refactored as 
```suggestion
		return fmt.Errorf("%s: %w", mediaType, errdef.ErrInvalidMediaType)
```

pack.go
Wwwsylvia: maybe `customEmptyConfig` 😀

shizhMSFT (approved):
LGTM

PR #582, opened 2023-09-01
[WIP] Implement Delete for OCI Layout

### LINKED ISSUE #454: oci.Store and oci.Store should implement content.Deleter

wangxiaoxuan273:
Attempt to resolve #454 

Based on the implementation of #470, but with `RWMutex`.

`DeletableStore` is basically copied from `oci.Store` with an added `RWMutex`.
`DeletableMemory` is basically copied from `graph.Memory` with an added `RWMutex`.


internal/graph/memory.go
```diff
@@ -116,6 +116,25 @@ func (m *Memory) Predecessors(_ context.Context, node ocispec.Descriptor) ([]oci
...
+			}
+			m.predecessors.Delete(predecessor)
+		}
+	}
+	return nil
+}
+
 // index indexes predecessors for each direct successor of the given node.
 // There is no data consistency issue as long as deletion is not implemented
 // for the underlying storage.
```
shizhMSFT, line 140: Here's the note to `RemoveFromIndex`.

wangxiaoxuan273: Moved the lock to the `DeletableStore` level.

content/oci/deletableOci.go
```diff
@@ -0,0 +1,348 @@
...
+	//     this method on each Tag() call.
+	//   - If AutoSaveIndex is set to false, it's the caller's responsibility
+	//     to manually call SaveIndex() when needed.
+	//   - Default value: true.
+	AutoSaveIndex bool
+	root          string
+	indexPath     string
+	index         *ocispec.Index
+	indexLock     sync.Mutex
+	lock          sync.RWMutex
```
Wwwsylvia, line 57: Looks like we can merge `lock` and `indexLock`.

wangxiaoxuan273: This is a good point, I need to think more about the lockings of `Tag()` and `SaveIndex()`. It looks like two locks are unnecessary, but currently merging the locks would cause a deadlock scenario.

wangxiaoxuan273: Current solution: merged `indexLock` into `lock`, and refactored `SaveIndex()` to be a private `saveIndex()` without lock and a public `SaveIndex()` with the lock. `Tag()` and `Delete()` will directly call the private `saveIndex()`.

internal/graph/memory.go
```diff
@@ -31,6 +31,7 @@ import (
 // Memory is a memory based PredecessorFinder.
 type Memory struct {
 	predecessors sync.Map // map[descriptor.Descriptor]map[descriptor.Descriptor]ocispec.Descriptor
+	successors   sync.Map // map[descriptor.Descriptor]map[descriptor.Descriptor]ocispec.Descriptor
```
Wwwsylvia, line 34: Since the successor list of a node is immutable and we don't need random access on the successors, the mapping can be from a descriptor to a successor list (`map[descriptor.Descriptor][]ocispec.Descriptor`).

internal/graph/memory.go
```diff
@@ -116,19 +117,54 @@ func (m *Memory) Predecessors(_ context.Context, node ocispec.Descriptor) ([]oci
...
+		successorsMap := succ.(*sync.Map)
+		successorsMap.Store(successorKey, successor)
 	}
 }
+
+func (m *Memory) indexIntoMemory(ctx context.Context, node ocispec.Descriptor) {
+	key := descriptor.FromOCI(node)
+	m.predecessors.LoadOrStore(key, &sync.Map{})
+	m.successors.LoadOrStore(key, &sync.Map{})
+	m.indexed.LoadOrStore(key, &sync.Map{})
```
Wwwsylvia, line 162: We should mark a node as indexed **after** it gets indexed.

internal/graph/memory.go
```diff
@@ -116,19 +117,54 @@ func (m *Memory) Predecessors(_ context.Context, node ocispec.Descriptor) ([]oci
...
+		predecessorsMap := pred.(*sync.Map)
+		predecessorsMap.Store(predecessorKey, node)
+		// store in m.successors, memory.successors[predecessorKey].Store(successor)
+		succ, _ := m.successors.Load(predecessorKey)
+		successorsMap := succ.(*sync.Map)
+		successorsMap.Store(successorKey, successor)
 	}
 }
+
+func (m *Memory) indexIntoMemory(ctx context.Context, node ocispec.Descriptor) {
```
Wwwsylvia, line 158: Why is it called `indexIntoMemory`? Everything is in memory in this file. 🤔

internal/graph/memory.go
```diff
@@ -116,19 +117,54 @@ func (m *Memory) Predecessors(_ context.Context, node ocispec.Descriptor) ([]oci
...
+	key := descriptor.FromOCI(node)
+	m.predecessors.LoadOrStore(key, &sync.Map{})
+	m.successors.LoadOrStore(key, &sync.Map{})
+	m.indexed.LoadOrStore(key, &sync.Map{})
+}
+
+func (m *Memory) removeFromMemory(ctx context.Context, node ocispec.Descriptor) {
+	key := descriptor.FromOCI(node)
+	m.predecessors.Delete(key)
```
Wwwsylvia, line 167: We might not want to delete the key from `predecessors` when a node is deleted.

content/oci/deletableOci.go
```diff
@@ -0,0 +1,350 @@
...
+	err = json.NewDecoder(layoutFile).Decode(&layout)
+	if err != nil {
+		return fmt.Errorf("failed to decode OCI layout file: %w", err)
+	}
+	return validateOCILayout(&layout)
+}
+
+// loadIndexFile reads index.json from the file system.
+// Create index.json if it does not exist.
+func (ds *DeletableStore) loadIndexFile(ctx context.Context) error {
```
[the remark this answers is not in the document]
Wwwsylvia: Oh it was because `loadIndexFile` is only called on initialization. We need a comment for this then. (I wrote this code bug I already forget it😅)

internal/graph/memory.go
wangxiaoxuan273: Do you have a better name? Maybe `createEntriesInMaps`?

internal/graph/memory.go
```diff
@@ -116,19 +117,54 @@ func (m *Memory) Predecessors(_ context.Context, node ocispec.Descriptor) ([]oci
 	return res, nil
 }
 
+// RemoveFromIndex removes the node from its predecessors and successors.
+func (m *Memory) RemoveFromIndex(ctx context.Context, node ocispec.Descriptor) error {
```
shizhMSFT, line 121: There is a race condition where the same node is indexed and removed at the same time.

content/oci/deletableOci.go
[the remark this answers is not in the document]
wangxiaoxuan273: I don't think we need to. This function is only used in `New()` and called only once?

internal/graph/memory.go
shizhMSFT: For example, what if another goroutine calls `Index()` when the current goroutine just about to hit the line 132?

internal/graph/memory.go
wangxiaoxuan273: Why not? Do we have a scenario of still needing it after the node is deleted? The information is stored in the node's predecessors' `successors` anyway.

internal/graph/memory.go
Wwwsylvia: The current implementation marks the node as indexed **before** it gets indexed.

internal/graph/memory.go
Wwwsylvia: To get the predecessors of a deleted node.

content/oci/deletableOci.go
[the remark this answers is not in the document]
Wwwsylvia: Yeah but people (like me) might not realize that😂

internal/graph/memoryWithDelete_test.go
```diff
@@ -0,0 +1,259 @@
...
+	B    = ocispec.Descriptor{MediaType: "B"}
+	BKey = descriptor.FromOCI(B)
+	C    = ocispec.Descriptor{MediaType: "C"}
+	CKey = descriptor.FromOCI(C)
+	D    = ocispec.Descriptor{MediaType: "D"}
+	DKey = descriptor.FromOCI(D)
+)
+
+// 条件：
+// 1. 只要node被index（作为函数的第二个输入），node就在predecessors，successors，indexed中有entry
```
northtyphoon, line 39: Can you please translate the comments to English? =)

wangxiaoxuan273: I will remove it or put it formally in English in later versions. Currently this is a draft PR, and I wrote this comment to remind myself the designs.

internal/graph/memory.go
wangxiaoxuan273: Why do we need to mark a node after it gets indexed?

internal/graph/memory.go
wangxiaoxuan273: In what scenarios do we need to get the predecessors of a deleted node? It sounds nice to have though.

internal/graph/memory.go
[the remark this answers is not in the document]
wangxiaoxuan273: Renamed as `Remove`.

internal/graph/memory.go
wangxiaoxuan273: Good point. I did not notice that I changed the behavior of `indexed` from the original behavior. Changed it back now. Currently `indexed` is updated **after** the node gets indexed.

internal/graph/memory.go
wangxiaoxuan273: Added the entry back. Currently we don't delete the entry in `predecessors` when doing `Delete`.

internal/graph/memory.go
wangxiaoxuan273: changed both `successors` and `predecessors` to the regular map, and added locks in `index` and `Remove`.

wangxiaoxuan273:
Created `graph.MemoryWithDelete`, which is basically `graph.Memory` with a lock. I think there are lock problems with `IndexAll`, I need help/discussion on that.

internal/graph/memoryWithDelete.go
```diff
@@ -0,0 +1,173 @@
...
+	"oras.land/oras-go/v2/internal/descriptor"
+	"oras.land/oras-go/v2/internal/status"
+	"oras.land/oras-go/v2/internal/syncutil"
+)
+
+// MemoryWithDelete is a MemoryWithDelete based PredecessorFinder.
+type MemoryWithDelete struct {
+	indexed      sync.Map // map[descriptor.Descriptor]any, this variable is only used by IndexAll
+	predecessors map[descriptor.Descriptor]map[descriptor.Descriptor]ocispec.Descriptor
+	successors   map[descriptor.Descriptor]map[descriptor.Descriptor]ocispec.Descriptor
```
Wwwsylvia, line 35: Can this be of type `map[descriptor.Descriptor][]ocispec.Descriptor`?
See https://github.com/oras-project/oras-go/pull/582#discussion_r1316895910

internal/graph/memoryWithDelete.go
```diff
@@ -0,0 +1,173 @@
...
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/internal/descriptor"
+	"oras.land/oras-go/v2/internal/status"
+	"oras.land/oras-go/v2/internal/syncutil"
+)
+
+// MemoryWithDelete is a MemoryWithDelete based PredecessorFinder.
+type MemoryWithDelete struct {
+	indexed      sync.Map // map[descriptor.Descriptor]any, this variable is only used by IndexAll
```
Wwwsylvia, line 33: Looks like this can be moved into `IndexAll`.

internal/graph/memoryWithDelete.go
```diff
@@ -0,0 +1,173 @@
...
+func (m *MemoryWithDelete) Predecessors(_ context.Context, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+	key := descriptor.FromOCI(node)
+	_, exists := m.predecessors[key]
+	if !exists {
+		return nil, nil
+	}
+	var res []ocispec.Descriptor
+	for _, v := range m.predecessors[key] {
+		res = append(res, v)
+	}
```
Wwwsylvia, line 121: nit:
```suggestion
	predecessors, exists := m.predecessors[key]
	if !exists {
		return nil, nil
	}
	var res []ocispec.Descriptor
	for _, v := range predecessors {
		res = append(res, v)
	}
```
?

internal/graph/memoryWithDelete.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+		m.predecessors[successorKey][predecessorKey] = node
+		// store in m.successors, MemoryWithDelete.successors[predecessorKey].Store(successor)
+		m.successors[predecessorKey][successorKey] = successor
+	}
+}
+
+func (m *MemoryWithDelete) createEntriesInMaps(ctx context.Context, node ocispec.Descriptor) {
+	key := descriptor.FromOCI(node)
+	if _, hasEntry := m.predecessors[key]; !hasEntry {
```
Wwwsylvia, line 161: nit: a simple `ok` should be good enough.

internal/graph/memoryWithDelete.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	m.removeEntriesFromMaps(ctx, node)
+	return nil
+}
+
+// index indexes predecessors for each direct successor of the given node.
+// There is no data consistency issue as long as deletion is not implemented
+// for the underlying storage.
+func (m *MemoryWithDelete) index(ctx context.Context, node ocispec.Descriptor, successors []ocispec.Descriptor) {
+	m.createEntriesInMaps(ctx, node)
```
Wwwsylvia, line 142: I don't think we need `createEntriesInMaps` and `removeEntriesFromMaps` as they are used in only one place. Just expanding them may make the caller function more readable?

content/oci/readonlyoci.go
```diff
@@ -186,6 +186,25 @@ func loadIndex(ctx context.Context, index *ocispec.Index, fetcher content.Fetche
 	return nil
 }
 
+// loadIndex loads index into memory.
+func loadIndexWithMemoryWithDelete(ctx context.Context, index *ocispec.Index, fetcher content.Fetcher, tagger content.Tagger, graph *graph.MemoryWithDelete) error {
```
Wwwsylvia, line 190: This should be defined in `deletableOci.go` as it has nothing to do with `readonlyoci.go`.

internal/graph/memoryWithDelete.go
wangxiaoxuan273: I think it's better to have these two functions for maintainability and modularity. The exact behaviors of `predecessors`, `successors` and other variables may change later as we continue to improve our design, the two functions reduce the chance of errors.

I originally did not have these two functions, but as I keep working on this feature, I've found the functions useful.

internal/graph/memoryWithDelete_test.go
```diff
@@ -0,0 +1,242 @@
...
+// 当前条件（后续可能改动）：
+// * 变量indexed的作用和行为和原始版本一样，没有加以变动
+// * 只要node被index（作为函数的第二个输入），node就在predecessors，successors中都有entry
+// * 只有node被index（作为函数的第二个输入），successors中才有其entry
+// * successors中的内容与node中的内容一致，只要entry还在，其内容就不会改变
+// * 删除一个node的时候，indexed和successors里面的entry也会删除。predecessors中的entry不会被删除
+
+// GC情况：
+// predecessors中可能有空的map，如下图中B被删除后，C还在predecessors中有entry但内容为空
+/*
```
shizhMSFT, line 50: All comments should be in English.

internal/resolver/memory.go
```diff
@@ -48,6 +48,11 @@ func (m *Memory) Tag(_ context.Context, desc ocispec.Descriptor, reference strin
 	return nil
 }
 
+// Delete removes a reference from index map.
+func (m *Memory) Delete(reference string) {
```
shizhMSFT, line 52: Missing unit tests for this func

internal/graph/memoryWithDelete_test.go
```diff
@@ -0,0 +1,242 @@
...
+var (
+	A    = ocispec.Descriptor{MediaType: "A"}
+	AKey = descriptor.FromOCI(A)
+	B    = ocispec.Descriptor{MediaType: "B"}
+	BKey = descriptor.FromOCI(B)
+	C    = ocispec.Descriptor{MediaType: "C"}
+	CKey = descriptor.FromOCI(C)
+	D    = ocispec.Descriptor{MediaType: "D"}
+	DKey = descriptor.FromOCI(D)
+)
```
shizhMSFT, line 35: Those variables should be local instead of global.

Also, name should be changed to something like `nodeA`, `nodeB`, `keyA`, `keyB`.

internal/graph/memory.go
wangxiaoxuan273: The variable `indexed` has been removed now.

content/oci/readonlyoci.go
wangxiaoxuan273: Good point, thanks.

internal/graph/DeletableMemory.go
shizhMSFT, line 1: golang file names use underscores like `deletable_memory.go`

wangxiaoxuan273: Updated the file names, sorry.

internal/container/set/set.go
```diff
@@ -33,3 +33,8 @@ func (s Set[T]) Contains(item T) bool {
 	_, ok := s[item]
 	return ok
 }
+
+// Delete deletes an item from the set
```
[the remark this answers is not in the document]
wangxiaoxuan273: Added the period.

internal/graph/deletablememory.go
```diff
@@ -0,0 +1,155 @@
...
+func NewDeletableMemory() *DeletableMemory {
+	return &DeletableMemory{
+		nodes:        make(map[descriptor.Descriptor]ocispec.Descriptor),
+		predecessors: make(map[descriptor.Descriptor]set.Set[descriptor.Descriptor]),
+		successors:   make(map[descriptor.Descriptor]set.Set[descriptor.Descriptor]),
+	}
+}
+
+// Index indexes predecessors for each direct successor of the given node.
+// There is no data consistency issue as long as deletion is not implemented
```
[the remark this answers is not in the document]
wangxiaoxuan273: Removed the comment.

content/oci/deletableoci.go
```diff
@@ -0,0 +1,357 @@
...
+func (ds *DeletableStore) writeIndexFile() error {
+	indexJSON, err := json.Marshal(ds.index)
+	if err != nil {
+		return fmt.Errorf("failed to marshal index file: %w", err)
+	}
+	return os.WriteFile(ds.indexPath, indexJSON, 0666)
+}
+
+// loadIndexInDeletableMemory loads index into the memory.
+func loadIndexInDeletableMemory(ctx context.Context, index *ocispec.Index, fetcher content.Fetcher, tagger content.Tagger, graph *graph.DeletableMemory) error {
```
Wwwsylvia, line 341: If we want to maintain `DeletableStore` separately, this function can be renamed to
`func (ds *DeletableStore) loadIndex` since it is only used by `DeletableStore`.

Issue #592, opened 2023-09-07
Referrers tag index is not updated when all referrers are removed and GC is skipped


shizhMSFT:
> [!NOTE]
> Credits to @qweeah for finding this bug.

Problem statement: Under the mode of referrers tag schema, when all referrers of a manifest are removed the referrers index is not updated if `SkipReferrersGC` is toggled to `true`.

Related code:
https://github.com/oras-project/oras-go/blob/4a52cfdc8e2e84430654dba83ce3fa7abf398b28/registry/remote/repository.go#L1455-L1476

If there are no referrers (i.e. `len(updatedReferrers) == 0`), we should remove the referrers index if GC is enabled as the above code does.

However, when GC is disabled / skipped, we need to keep the previous index. In this case, a new index with no manifest should be generated and pushed. Note that the `image-spec` allows the index to have zero `manifests`.

PR #593, opened 2023-09-08
fix: push an empty index when deleting the last referrer with SkipReferrersGC

Wwwsylvia:
Fix: #592

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -1425,17 +1425,17 @@ func (s *manifestStore) indexReferrersForPush(ctx context.Context, desc ocispec.
 func (s *manifestStore) updateReferrersIndex(ctx context.Context, subject ocispec.Descriptor, change referrerChange) (err error) {
 	referrersTag := buildReferrersTag(subject)
 
-	skipDelete := s.repo.SkipReferrersGC
+	var noOldIndex bool
 	var oldIndexDesc ocispec.Descriptor
-	var referrers []ocispec.Descriptor
+	var oldReferrers []ocispec.Descriptor
 	prepare := func() error {
 		// 1. pull the original referrers list using the referrers tag schema
 		var err error
-		oldIndexDesc, referrers, err = s.repo.referrersFromIndex(ctx, referrersTag)
+		oldIndexDesc, oldReferrers, err = s.repo.referrersFromIndex(ctx, referrersTag)
 		if err != nil {
 			if errors.Is(err, errdef.ErrNotFound) {
 				// no old index found, skip delete
-				skipDelete = true
+				noOldIndex = true
 				return nil
 			}
 			return err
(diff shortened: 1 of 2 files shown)


registry/remote/repository.go
```diff
@@ -1453,7 +1453,11 @@ func (s *manifestStore) updateReferrersIndex(ctx context.Context, subject ocispe
 		}
 
 		// 3. push the updated referrers list using referrers tag schema
-		if len(updatedReferrers) > 0 {
+		if len(updatedReferrers) > 0 || s.repo.SkipReferrersGC {
+			// push a new index when:
+			// 1. updatedReferrers is not empty
+			// 2. OR referrers GC is skipped, in this case the old index
+			//    won't get deleted and an empty index should be pushed
```
shizhMSFT, line 1460: This block of comment is a bit confusing. You may reword it to be more natural.

Wwwsylvia: Reworded

shizhMSFT (approved):
LGTM

TerryHowe (approved):
/lgtm

registry/remote/repository.go
```diff
@@ -1463,14 +1468,15 @@ func (s *manifestStore) updateReferrersIndex(ctx context.Context, subject ocispe
...
-		// 4. delete the dangling original referrers index
-		if !skipDelete {
-			if err := s.repo.delete(ctx, oldIndexDesc, true); err != nil {
-				return &ReferrersError{
-					Op:      opDeleteReferrersIndex,
-					Err:     fmt.Errorf("failed to delete dangling referrers index %s for referrers tag %s: %w", oldIndexDesc.Digest.String(), referrersTag, err),
-					Subject: subject,
-				}
+		// 4. delete the dangling original referrers index, if applicable
+		if s.repo.SkipReferrersGC || noOldIndex {
```
qweeah, line 1472: Shouldn't `noOldIndex` already be implied by `len(oldReferrers) == 0`?

Wwwsylvia: No. Consider the following cases:
1. Old referrers index exists and contains at least one referrers: `noOldIndex == false`, `len(oldReferrers) > 0`
2. Old referrers index exists but contains no referrer: `noOldIndex == false`, `len(oldReferrers) == 0`
3. Old referrers index does not exist (server returns 404):  `noOldIndex == true`, `len(oldReferrers) == 0`

We skip deleting the old index when `skipReferrersGC == true` or the old index is not present (case 3).

qweeah: How about `s.repo.SkipReferrersGC || content.Equal(oldIndexDesc, ocispec.Descriptor{})`

qweeah: The main purpose of this comment is: whether an old index exists is already implied by returning value of function `referrersFromIndex`.

Wwwsylvia: The suggested way does have the same effect as the current code. But I think having `noOldIndex` is more readable. Otherwise we will need extra comment to explain what does `content.Equal(oldIndexDesc, ocispec.Descriptor{})` mean.

Wwwsylvia: Refactored a little bit.

shizhMSFT (approved):
LGTM

Issue #581, opened 2023-08-31
Bug: `auth.AppendScopes` does not distinguish between registries


Wwwsylvia:
When copying an artifact from a remote repository to another, we may code something like below:

```go
// add scope hints
scope := auth.ScopeRepository(dst.Reference.Repository, auth.ActionPull, auth.ActionPush)
ctx = auth.AppendScopes(ctx, scope)
// copy
desc, err := oras.Copy(ctx, src, srcRef, dst, dstRef, opts)
```

This code snippet adds both `pull` and `push` scope hints to `ctx` for the `dst` repository in order to decrease the number of auth requests on `dst` in the later copy operation.
However, `auth.AppendScopes` does not distinguish scopes between registries, resulting in `pull` and `push` scopes being requested for both `src` and `dst` repository, which is not safe.

PR #604, opened 2023-09-21
feat: support per-host scope hints

Wwwsylvia:
The purpose of this PR is to fix the bug, but new APIs are needed to avoid breaking changes.
1. Introduce `auth.WithScopesForHost`
2. Introduce `auth.AppendScopesForHost`
3. Introduce `auth.GetScopesForHost` and `auth.GetAllScopesForHost`
4. Introduce `auth.AppendRepositoryScope`

Resolves: #581


registry/remote/auth/scope.go
```diff
@@ -54,6 +58,28 @@ func ScopeRepository(repository string, actions ...string) string {
...
+// For example, uploading blob to the repository "hello-world" does HEAD request
+// first then POST and PUT. The HEAD request will return a challenge for scope
+// `repository:hello-world:pull`, and the auth client will fetch a token for
+// that challenge. Later, the POST request will return a challenge for scope
+// `repository:hello-world:push`, and the auth client will fetch a token for
+// that challenge again. By invoking AppendScopeHints with the actions
+// [ActionPull] and [ActionPush] for the repository `hello-world`,
+// the auth client with cache is hinted to fetch a token via a single token
+// fetch request for all the HEAD, POST, PUT requests.
+func AppendScopeHints(ctx context.Context, ref registry.Reference, actions ...string) context.Context {
```
Wwwsylvia, line 75: It's hard to name this method...

registry/remote/auth/scope.go
```diff
@@ -93,7 +119,62 @@ func AppendScopes(ctx context.Context, scopes ...string) context.Context {
...
+// that challenge again. By invoking WithScopesPerHost with the scope
+// `repository:hello-world:pull,push`, the auth client with cache is hinted to
+// fetch a token via a single token fetch request for all the HEAD, POST, PUT
+// requests.
+//
+// Passing an empty list of scopes will virtually remove the scope hints in the
+// context for the given host.
+//
+// Reference: https://docs.docker.com/registry/spec/auth/scope/
+func WithScopesPerHost(ctx context.Context, host string, scopes ...string) context.Context {
```
Wwwsylvia, line 149: I was struggling with naming... The options were:
1. `WithScopesPerHost`
2. `WithPerHostScopes`
3. `WithScopesForHost`

registry/remote/auth/scope.go
```diff
@@ -93,7 +119,62 @@ func AppendScopes(ctx context.Context, scopes ...string) context.Context {
 // GetScopes returns the scopes in the context.
 func GetScopes(ctx context.Context) []string {
```
Wwwsylvia, line 120: Should we deprecate the old `GetScopes`, WithScopes` and `AppendScopes`? They are not safe enough but they a little bit convenient than the new ones in scenarios where there is only one host or two hosts share the same scope.

registry/remote/auth/scope.go
qweeah: 1 sounds good to me.

registry/remote/auth/scope.go
```diff
@@ -93,7 +119,62 @@ func AppendScopes(ctx context.Context, scopes ...string) context.Context {
...
 	if scopes, ok := ctx.Value(scopesContextKey{}).([]string); ok {
-		return append([]string(nil), scopes...)
+		return slices.Clone(scopes)
+	}
+	return nil
+}
+
+// scopesPerHostContextKey is the context key for per-host scopes.
+type scopesPerHostContextKey struct{}
```
shizhMSFT, line 128: Why not make
```suggestion
type scopesPerHostContextKey string
```
where the `string` stores `host`?

shizhMSFT: so that we don't require a map

registry/remote/auth/scope.go
shizhMSFT: Nope. We can consider `GetScopes()` as a global version of `GetScopesForHost()`.

registry/remote/auth/client.go
```diff
@@ -223,17 +228,25 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
...
-		// merge hinted scopes with challenged scopes
-		scopes := GetScopes(ctx)
-		if scope := params["scope"]; scope != "" {
-			scopes = append(scopes, strings.Split(scope, " ")...)
+		scopes := GetScopesPerHost(ctx, host)
+		cleanScopeLen := len(scopes)
+		if moreScopes := GetScopes(ctx); len(moreScopes) > 0 {
+			// merge per-host scopes with generic scopes
+			scopes = append(scopes, moreScopes...)
+		}
```
shizhMSFT, line 236: The code related to `GetScopes()` should be done inside of `GetScopesPerHost()`.

registry/remote/auth/scope.go
shizhMSFT: `WithScopesForHost` sounds good to me.

wangxiaoxuan273: I suggest `WithHostScopes`, lol.

registry/remote/auth/scope.go
wangxiaoxuan273: > Why not make
> 
> where the `string` stores `host`?

According to the [doc](https://pkg.go.dev/context#WithValue:~:text=The%20provided%20key%20must%20be%20comparable%20and%20should%20not%20be%20of%20type%20string%20or%20any%20other%20built%2Din%20type%20to%20avoid%20collisions%20between%20packages%20using%20context.), we should not use a string as context key.

registry/remote/auth/scope.go
```diff
@@ -93,7 +119,62 @@ func AppendScopes(ctx context.Context, scopes ...string) context.Context {
...
+func WithScopesPerHost(ctx context.Context, host string, scopes ...string) context.Context {
+	var scopesByHost map[string][]string
+	if old, ok := ctx.Value(scopesPerHostContextKey{}).(map[string][]string); ok {
+		scopesByHost = make(map[string][]string, len(old))
+		maps.Copy(scopesByHost, old)
+	} else {
+		scopesByHost = make(map[string][]string, 1)
+	}
+
+	scopesByHost[host] = CleanScopes(scopes)
```
wangxiaoxuan273, line 158: Is there going to be a data race condition, if more than one threads simultaneously reach L153 and L158?

registry/remote/auth/scope.go
```diff
@@ -93,7 +119,62 @@ func AppendScopes(ctx context.Context, scopes ...string) context.Context {
...
+// `repository:hello-world:pull,push`, the auth client with cache is hinted to
+// fetch a token via a single token fetch request for all the HEAD, POST, PUT
+// requests.
+//
+// Passing an empty list of scopes will virtually remove the scope hints in the
+// context for the given host.
+//
+// Reference: https://docs.docker.com/registry/spec/auth/scope/
+func WithScopesPerHost(ctx context.Context, host string, scopes ...string) context.Context {
+	var scopesByHost map[string][]string
```
wangxiaoxuan273, line 150: nit: maybe we can have a `if len(scopes) == 0 { return ctx }` here, just like in `AppendScopesPerHost()` and `AppendScopes()`?

Wwwsylvia: Allowing an empty list of scopes actually enables users to clean up the scopes in the context.

registry/remote/auth/scope.go
Wwwsylvia: The votes are tied!

registry/remote/auth/client.go
Wwwsylvia: But users may not want to get a mixed result of scopes and per-host scopes when invoking ` GetScopesPerHost()`.

registry/remote/auth/scope.go
Wwwsylvia: Good catch! There is a Time-of-Check to Time-of-Use race condition... I'm still finding a way to fix it.

registry/remote/auth/scope.go
wangxiaoxuan273: This conversation can be resolved.

registry/remote/auth/scope.go
Wwwsylvia: Just realized that context key can actually be a `string` as long as it's a custom type so that there won't be collisions between packages, as shown in the [example](https://pkg.go.dev/context#example-WithValue).

internal/maps/maps.go
```diff
@@ -13,17 +13,16 @@ See the License for the specific language governing permissions and
...
-func WithScopeHint(ctx context.Context, ref registry.Reference, actions ...string) context.Context {
-	scope := auth.ScopeRepository(ref.Repository, actions...)
-	return auth.AppendScopes(ctx, scope)
+// Copy copies all key/value pairs in src adding them to dst.
+// When a key in src is already present in dst,
+// the value in dst will be overwritten by the value associated
+// with the key in src.
+//
+// Reference: https://pkg.go.dev/maps@go1.21.1#Copy
+func Copy[M1 ~map[K]V, M2 ~map[K]V, K comparable, V any](dst M1, src M2) {
```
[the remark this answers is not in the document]
Wwwsylvia: Ended up removing this function.

registry/remote/auth/scope.go
Wwwsylvia: Updated.

registry/remote/auth/scope.go
Wwwsylvia: Updated the code and this is not valid anymore.

registry/remote/auth/client.go
Wwwsylvia: Added a private helper function for this.

registry/remote/auth/scope.go
```diff
@@ -93,11 +118,77 @@ func AppendScopes(ctx context.Context, scopes ...string) context.Context {
...
+func GetScopesForHost(ctx context.Context, host string) []string {
+	if scopes, ok := ctx.Value(scopesForHostContextKey(host)).([]string); ok {
+		return slices.Clone(scopes)
 	}
 	return nil
 }
 
+// getAllScopesForHost returns the scopes in the context for the given host,
+// including global scopes added by [WithScopes] and [AppendScopes].
+func getAllScopesForHost(ctx context.Context, host string) []string {
```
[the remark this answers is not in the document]
Wwwsylvia: Do you have scenarios in mind where people would use it? My thinking was we can expose it when someone needs it.

registry/remote/auth/scope.go
Wwwsylvia: Yeah I agree. Let's call it `AppendRepositoryScope`.
// Just that I liked the idea of "hint"😄.

Wwwsylvia: I even thought of "HintRepositoryScope"...

registry/remote/auth/scope.go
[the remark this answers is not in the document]
shizhMSFT: Actually, you've just demonstrated it in this PR.

shizhMSFT (approved):
LGTM

PR #605, opened 2023-09-22
refactor: make credentials.NewMemoryStore return an interface

Wwwsylvia:
1. Un-expose `credentials.MemoryStore` as it is unecessary to be public
2. Make `credentials.NewMemoryStore` return an interface instead of a struct

### DIFF
--- a/registry/remote/credentials/memory_store.go
+++ b/registry/remote/credentials/memory_store.go
@@ -22,18 +22,18 @@ import (
 	"oras.land/oras-go/v2/registry/remote/auth"
 )
 
-// MemoryStore is a store that keeps credentials in memory.
-type MemoryStore struct {
+// memoryStore is a store that keeps credentials in memory.
+type memoryStore struct {
 	store sync.Map
 }
 
 // NewMemoryStore creates a new in-memory credentials store.
-func NewMemoryStore() *MemoryStore {
-	return &MemoryStore{}
+func NewMemoryStore() Store {
+	return &memoryStore{}
 }
 
 // Get retrieves credentials from the store for the given server address.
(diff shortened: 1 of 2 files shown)


Wwwsylvia:
@Uanid Please take a look

shizhMSFT (approved):
LGTM

Uanid:
> @Uanid Please take a look at this change

How about changing FileStore to a private too?
If we add the DisablePut property to NewFileStore? Then, we could make it private.
Changing the public API could be an issue, but since you moved the code to oras-go and package path is changed, some changes now seem okay.

TerryHowe:
> > @Uanid Please take a look at this change
> 
> How about changing FileStore to a private too? If we add the DisablePut property to NewFileStore? Then, we could make it private. Changing the public API could be an issue, but since you moved the code to oras-go and package path is changed, some changes now seem okay.

Sounds like a good idea, but subject of another PR

TerryHowe (approved):
/lgtm

Uanid:
> @Uanid Please take a look at this change

LGTM too

Wwwsylvia:
> How about changing FileStore to a private too?
If we add the DisablePut property to NewFileStore? Then, we could make it private.
Changing the public API could be an issue, but since you moved the code to oras-go and package path is changed, some changes now seem okay.

But that would require users to specify an extra parameter on initialization. I think it's ok to keep `FileStore` exposed for convenience. @shizhMSFT What do you think?

shizhMSFT:
> But that would require users to specify an extra parameter on initialization. I think it's ok to keep FileStore exposed for convenience.

The point of having `FileStore` exposed is to make the property changeable after initialization.

PR #608, opened 2023-09-25
feat: support deletion for OCI storage

wangxiaoxuan273:
Signed-off-by: Xiaoxuan Wang <wangxiaoxuan119@gmail.com>

Part 3/4 of #454

Based on draft PR #582

PR #607, opened 2023-09-25
feat: support deletion for memory resolver

wangxiaoxuan273:
Part 2/4 of #454 

Based on draft PR #582

PR #606, opened 2023-09-25
feat: support deletion for memory graph

wangxiaoxuan273:
Signed-off-by: Xiaoxuan Wang <wangxiaoxuan119@gmail.com>

Part 1/4 of #454

Based on draft PR #582 

Current behavior regarding `graph.Memory.Remove(node)`:
* `node` entry in `m.successors` and `m.nodes` is removed.
* `node` is removed from its successors predecessors list.
* `node` entry in `m.predecessors` is NOT removed, **unless all its predecessors no longer exist**. 
* `node` is NOT removed from its predecessors' `m.successors` list. The `m.successors` is always in accordance with the actual content.

PR #620, opened 2023-10-18
[RFC] Add Delete() to oci.Store.

eiffel-fl:
Hi.

First, I would like to thank you for the work done in #470.
I tried to use this function but had troubles as the `content.Storage` in `oci.Store` does not implement the `Delete` function.
I thought to add `Deleter` to `oci.Storage` but then I got troubles during compilation with `cas.Memory`.
So, I casted the `content.Storage` to `oci.Storage` which is not really good but at least it works.

As a consequence, I would like to get your feedback on how to permit `oci.Store` to `Delete()` descriptor in a proper way regarding the layers of abstraction you have.

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -103,6 +103,11 @@ func (s *Store) Fetch(ctx context.Context, target ocispec.Descriptor) (io.ReadCl
 	return s.storage.Fetch(ctx, target)
 }
 
+// Delete deletes the content identified by the descriptor.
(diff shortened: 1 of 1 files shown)


Wwwsylvia:
Hi @eiffel-fl , thanks for the PR. We have merged PR #606, #607 and #608 that enhanced the `oci` package, and we have another PR #614 in progress, feel free to review it and let us know if it helps.

PR #627, opened 2023-10-24
fix: error handling when delete non-existent target

### LINKED ISSUE #626: `oci.Storage.Delete()` should return `ErrNotFound` if target is not present in the storage
In the current implementation, a `*PathError` will be return if `target` is not present in the storage, which is not user-friendly.

https://github.com/oras-project/oras-go/blob/0f1dc30a0fd713f3bafcf3bd52693b2e8a8a6fa5/content/oci/storage.go#L109-L117

wangxiaoxuan273:
resolves #626

### DIFF
--- a/content/oci/storage.go
+++ b/content/oci/storage.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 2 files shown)


content/oci/storage.go
```diff
@@ -108,6 +108,13 @@ func (s *Storage) Push(_ context.Context, expected ocispec.Descriptor, content i
 
 // Delete removes the target from the system.
 func (s *Storage) Delete(ctx context.Context, target ocispec.Descriptor) error {
+	exists, err := s.Exists(ctx, target)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return fmt.Errorf("target to delete is not present in the storage: %w", errdef.ErrNotFound)
+	}
```
shizhMSFT, line 117: It introduces a race condition of [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use).

Can you check the error returned by `os.Remove()` and return `ErrNotFound` accordingly?

content/oci/storage.go
```diff
@@ -113,7 +114,14 @@ func (s *Storage) Delete(ctx context.Context, target ocispec.Descriptor) error {
 		return fmt.Errorf("%s: %s: %w", target.Digest, target.MediaType, errdef.ErrInvalidDigest)
 	}
 	targetPath := filepath.Join(s.root, path)
-	return os.Remove(targetPath)
+	err = os.Remove(targetPath)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return fmt.Errorf("target to delete is not present in the storage: %w", errdef.ErrNotFound)
```
shizhMSFT, line 120: Can you follow the same error message pattern as line 114?

wangxiaoxuan273: changed accordingly,

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

Wwwsylvia:
Sorry, accidentally clicked "close", thought it was "merge"...
image

PR #631, opened 2023-10-31
feat: minimal change to support cross-repo blob mounting

ktarplee:
Export `ErrSkipDesc` as `ErrSkipDesc` allows users of the library to implement `CopyGraphOptions.PreCopy` to call `Mount()` and then return `ErrSkipDesc` from `PreCopy` which bypasses the downstream copy operation.

Closes #580

PR #632, opened 2023-10-31
feat: add `WithMount` method to support cross repository blob mounting

### LINKED ISSUE #580: Support cross-repository mounting in Copy
We have already have [Mounter](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.0/registry#Mounter) implemented for [Repository](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.0/registry/remote#Repository.Mount), we can consider leveraging it in [Copy](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.0#Copy) and [CopyGraph](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.0#CopyGraph) for better performance.

ktarplee:
This includes everything from #631 but adds more (debatable) changes.

Closes #580

I realize I still need to add tests but I wanted to see if this approach is acceptable first.

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -114,6 +114,66 @@ type CopyGraphOptions struct {
 	FindSuccessors func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error)
 }
 
+// WithMount enabled cross repository blob mounting.
+// sourceReference is the repository to use for mounting (the mount point).
+// mounter is the destination for the mount (a well-known implementation of this is *registry.Repository representing the target).
+// onMounted is called (if provided) when the blob is mounted.
+// The original PreCopy hook is called only on copy, and therefore not when the blob is mounted.
+func (opts *CopyGraphOptions) WithMount(sourceRepository string, mounter registry.Mounter, onMounted func(context.Context, ocispec.Descriptor) error) {
+	preCopy := opts.PreCopy
+	opts.PreCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
+		// Only care to mount blobs
+		if descriptor.IsManifest(desc) {
+			// still want to call PreCopy if it is a manifest
+			if preCopy != nil {
+				return preCopy(ctx, desc)
+			}
+			return nil
+		}
+
+		var mountFailed bool
+		getContent := func() (io.ReadCloser, error) {
+			// call the original PreCopy function if it exists
+			if preCopy != nil {
+				if err := preCopy(ctx, desc); err != nil {
+					return nil, err
+				}
+			}
+			// the invocation of getContent indicates that mounting has failed
+			mountFailed = true
+
+			// To avoid needing a content.Fetcher as an input argument we simply fall back to the default behavior
+			// as if getContent was nil
+			return nil, errdef.ErrUnsupported
+		}
+
+		// Mount or copy
+		if err := mounter.Mount(ctx, desc, sourceRepository, getContent); err != nil {
+			return err
+		}
+
+		if !mountFailed {
+			// mounted
+			if onMounted != nil {
+				if err := onMounted(ctx, desc); err != nil {
+					return err
+				}
+			}
+			// signal that the descriptor now exists
+			return SkipNode
+		}
+
+		// we copied it
+		if opts.PostCopy != nil {
+			if err := opts.PostCopy(ctx, desc); err != nil {
+				return err
+			}
+		}
+		// signal that the descriptor now exists
+		return SkipNode
+	}
+}
+
 // Copy copies a rooted directed acyclic graph (DAG) with the tagged root node
 // in the source Target to the destination Target.
 // The destination reference will be the same as the source reference if the
--- a/copy.go
+++ b/copy.go
@@ -144,7 +144,7 @@ func (opts *CopyGraphOptions) WithMount(sourceRepository string, mounter registr
 
 			// To avoid needing a content.Fetcher as an input argument we simply fall back to the default behavior
 			// as if getContent was nil
-			return nil, errdef.ErrUnsupported
+			return nil, registry.UseSourceRepository
 		}
 
 		// Mount or copy
--- a/example_copy_test.go
+++ b/example_copy_test.go
@@ -35,6 +35,7 @@ import (
 	"oras.land/oras-go/v2/content/memory"
 	"oras.land/oras-go/v2/content/oci"
 	"oras.land/oras-go/v2/internal/spec"
+	"oras.land/oras-go/v2/registry"
 	"oras.land/oras-go/v2/registry/remote"
 )
 
@@ -213,6 +214,37 @@ func ExampleCopy_remoteToRemote() {
 
 	// Output:
 	// sha256:7cbb44b44e8ede5a89cf193db3f5f2fd019d89697e6b87e8ed2589e60649b0d1
+}
+
+func ExampleCopy_remoteToRemoteWithMount() {
+	reg, err := remote.NewRegistry(remoteHost)
+	if err != nil {
+		panic(err) // Handle error
+	}
+	ctx := context.Background()
+	src, err := reg.Repository(ctx, "source")
+	if err != nil {
+		panic(err) // Handle error
+	}
+	dst, err := reg.Repository(ctx, "target")
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	tagName := "latest"
+
+	opts := oras.CopyOptions{}
+	// Enable cross-repository blob mounting
+	opts.WithMount("source", dst.(registry.Mounter), nil)
+
+	desc, err := oras.Copy(ctx, src, tagName, dst, tagName, opts)
+	if err != nil {
+		panic(err) // Handle error
+	}
+	fmt.Println("Final", desc.Digest)
+
+	// Output:
+	// Final sha256:7cbb44b44e8ede5a89cf193db3f5f2fd019d89697e6b87e8ed2589e60649b0d1
 }
 
 func ExampleCopy_remoteToLocal() {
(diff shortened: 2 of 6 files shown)


ktarplee:
@Wwwsylvia Any feedback on this approach?

Wwwsylvia:
@ktarplee I'm busy with other stuffs these days, might take a closer look into this a bit later. Sorry for the inconvenience. 😟

copy.go
```diff
@@ -112,6 +112,54 @@ type CopyGraphOptions struct {
...
+			return err
+		}
+
+		if !mountFailed && onMounted != nil {
+			onMounted(ctx, desc)
+		}
+
+		// Mount succeeded or we copied it
+		// either way we return ErrSkipDesc to signal that the descriptor now exists
+		return ErrSkipDesc
```
Wwwsylvia, line 159: We also need to call `PostCopy` if `desc` is copied.

Wwwsylvia (commented):
I think this works.

copy.go
```diff
@@ -112,6 +112,64 @@ type CopyGraphOptions struct {
 	FindSuccessors func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error)
 }
 
+// WithMount enabled cross repository blob mounting.
+// sourceReference is the repository to use for mounting (the mount point).
+// mounter is the destination for the mount (a well-known implementation of this is *registry.Repository representing the target).
+// onMounted is called (if provided) when the blob is mounted.
+// The original PreCopy hook is called only on copy, and therefore not when the blob is mounted.
+func (opts *CopyGraphOptions) WithMount(sourceRepository string, mounter registry.Mounter, onMounted func(context.Context, ocispec.Descriptor)) {
```
ktarplee, line 120: @Wwwsylvia Should `onMounted` return an `error` like PreCopy and PostCopy?

ktarplee:
Why does `registry.Repository` not include `registry.Mounter`?

copy.go
Wwwsylvia: I think it should for extendibility. But I don't have a good use case in mind though.

Wwwsylvia:
> Why does registry.Repository not include registry.Mounter?

[`registry.Mounter`](https://pkg.go.dev/oras.land/oras-go/v2/registry#Mounter) was introduced in [`v2.2.0`](https://github.com/oras-project/oras-go/releases/tag/v2.2.0) where breaking changes were not allowed.

https://github.com/oras-project/oras-go/blob/f29607202d5a72ee5325171bf7fc6698ccb4e45f/registry/repository.go#L113-L124

Wwwsylvia:
I need some more time to review this and do some integration tests.

copy.go
ktarplee: `onMounted` now returns an error.

Wwwsylvia (commented):
I've tested this against ACR, zot and distribution.
The blobs can be mounted as expected in ACR and distribution, while they gets skipped in zot. Looks like zot deduplicates blobs across repositories.

LGTM with a few nit suggestions.
BTW can we have the PR title start with "feat:"?

registry/remote/repository_test.go
```diff
@@ -421,16 +421,37 @@ func TestRepository_Mount_Fallback(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("Repository.Push() error = %v", err)
-	}
-	if !bytes.Equal(gotBlob, blob) {
-		t.Errorf("Repository.Mount() = %v, want %v", gotBlob, blob)
-	}
-	if got, want := sequence, "post get put "; got != want {
-		t.Errorf("unexpected request sequence; got %q want %q", got, want)
-	}
+	t.Run("nil", func(t *testing.T) {
```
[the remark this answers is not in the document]
Wwwsylvia: Same for the `ErrUnsupported` one?

Wwwsylvia (approved):
LGTM

Wwwsylvia:
@shizhMSFT Please take a look when you have a chance.

shizhMSFT:
> @shizhMSFT Please take a look when you have a chance.

I’m currently OOF and will review when back.

qweeah:
LGTM although IANAM

(Tried to implement an [index creation command](https://github.com/qweeah/oras/releases/tag/v1.3.0-index) based on d5cc4c20693ee5b31237fdf626dd8f6ee942a5c9 and it looks good, see [asciicast](https://asciinema.org/a/BrBhvqGkQAvC9mZutPnNFyqst))

registry/remote/repository.go
```diff
@@ -806,6 +806,10 @@ func (s *blobStore) Mount(ctx context.Context, desc ocispec.Descriptor, fromRepo
 	var r io.ReadCloser
 	if getContent != nil {
 		r, err = getContent()
+		if errors.Is(err, errdef.ErrUnsupported) {
+			// getContent can return a ErrUnsupported to fallback to the default copy operation
+			r, err = s.sibling(fromRepo).Fetch(ctx, desc)
+		}
```
shizhMSFT, line 812: This behavior of `errdef.ErrUnsupported` should also be documented in https://pkg.go.dev/oras.land/oras-go/v2@v2.3.1/registry#Mounter

shizhMSFT: To avoid breaking change and confusion, I would like to introduce an internal error instance to replace the well known  `errdef.ErrUnsupported` so that users won't get surprised when they do
```go
store.Mount(ctx, desc, fromRepo, func() (io.ReadCloser, error) {
    return nil, errdef.ErrUnsupported
})

File: example_copy_test.go [hunk 5]
```diff
@@ -215,6 +216,37 @@ func ExampleCopy_remoteToRemote() {
...
+	dst, err := reg.Repository(ctx, "target")
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	tagName := "latest"
+
+	opts := oras.CopyOptions{}
+	// Enable cross-repository blob mounting
+	opts.WithMount("source", dst.(registry.Mounter), nil)
```
shizhMSFT, line 238: What will happen if other target instead of `dst` is passed to `WithMount`?

shizhMSFT, line 238: What will happen if someone call `WithMount` twice?

example_copy_test.go
```diff
@@ -215,6 +216,37 @@ func ExampleCopy_remoteToRemote() {
...
+		panic(err) // Handle error
+	}
+	dst, err := reg.Repository(ctx, "target")
+	if err != nil {
+		panic(err) // Handle error
+	}
+
+	tagName := "latest"
+
+	opts := oras.CopyOptions{}
```
shizhMSFT, line 236: What if someone reuse `oras.CopyOptions`?

shizhMSFT (commented):
@ktarplee Sorry for the late review as I was OOF.

example_copy_test.go
ktarplee: Let's say the true destination target is `dst` but instead they pass in `dst2`.  The blobs would be mounted or copied to `dst2` but the manifest would be copied to `dst` and fail due to missing blobs.

ktarplee: This is a fun one.  I think it would work and not produce an error (but I have not tried it).  Let's call the `preCopy` functions `preCopy1`, `preCopy2`, `preCopy3`.  The last one is the one created by the last call to `WithMount`.   The first thing that would happen is the call to `preCopy3` which would attempt a mount.  Let's assume the mount failed, then `getContent` would get called (defined by `preCopy3`) and it would call `preCopy2`.  `preCopy2` was created by the first call to `WithMount` so it will try to mount the blob with that source repo.  Let's assume it also fails, then `getContent` would call `preCopy1`.  Then `getContent` (from `preCopy2`) would return `errdef.ErrUnsupported` (to be renamed) and the blob would be copied.  Then `getContent` (from `preCopy3`) would return `errdef.ErrUnsupported` (to be renamed) and the blob would be copied **again**.  So it seems like the blob would be copied twice with this design.  Not fatal but not ideal either.

example_copy_test.go
ktarplee: The CopyOptions after a WithMount would only be usable when the same source and destination are not changed.  That is not ideal.  A shallow copy of the `oras.CopyOptions` before calling `WithMount()` solves this problem.  That is how I use this in my code.

example_copy_test.go
shizhMSFT: Can we avoid that situation in the first place with a better API design?

example_copy_test.go
shizhMSFT: It would be ideal if we would reuse the CopyOptions. Currently, only PackOptions is not re-usable.

registry/remote/repository.go
ktarplee: Do you really think the error should be non-exported (internal)?  I think it might be better to be exported so consumers of the library can use it.  `clone` and `sibling` are not exported so user's defining your own Fetch based on the existing repo is slightly error prone.  But we can decide this later, if we can implement the more complicated use case (trying multiple mounts within a copy for a single blob) then we can easily justify keeping this sentinel error internal.

So what should the name be?  Maybe `unimplemented` or `useSourceRepository`.  Also where would we want this defined.  For now I exported it and defined it right next to the definition of the `type Mounter interface`.  If we make it internal (but exported from the package because it is needed in both the top level package and `registry/remote` we probably need to define a new internal package at the top level.

example_copy_test.go
ktarplee: We could change the signature of the `WithMount()` to return a modified copy of the receiver.  It would not modify the receiver so `opts` below could be reused.  It would then be used like so:
```go
desc, err := oras.Copy(ctx, src, tagName, dst, tagName, opts.WithMount("source", dst.(registry.Mounter), nil))
```
I think that would reduce the misuse and make it more obvious what it does.

example_copy_test.go
ktarplee: We could add a field to `CopyOptions`, something like `tryingMount bool`.  In `WithMount()` we can check that `tryingMount` is false (else panic) and then set it to true.  Then if the programmer makes the error to call `WithMount` twice they get a reasonable error.

ktarplee:
I have an idea to improve the API design of this and allow for trying to mount from multiple source locations.  I will work on that very soon.

ktarplee:
Closing this in favor of #665

PR #635, opened 2023-11-06
fix: check content-type when using referrers API

### LINKED ISSUE #633: Repository.Referrers does not properly detect the absence of the referrers API for registry.k8s.io
To reproduce

```shell
$ oras discover registry.k8s.io/sig-storage/csi-resizer:v1.8.0
Error: GET "https://accounts.google.com/v3/signin/identifier REDACTED": failed to decode response: invalid character '<' looking for beginning of value
```

The issue does not seem to be with in the `oras` CLI tool but rather the detection logic within the Repository's Referrers method to determine when a registry supports the referrers API or not.  In the above example, `registry.k8s.io` was erroneously determined to implement the referrers API.  In reality the response was a 307 that was followed to a login page for a user that was HTML and page returned a 200 so `oras-go` thinks that it supports the referrers API.  Then when the referrers API is used, it is obvious that the JSON parser is trying to parse HTML.  Not good.

So we need better detection logic for the referrers API.  Having the request (made in pingReferrers) respond with a 200 is necessary but not sufficient in most cases.   **We could check that the content type in the response is actually an OCI index.**  It looks like there are lots of places withing `oras-go` where the Content-Type of the response is not checked to be a set of acceptable values before parsing the body.  This is generally considered bad practice.

ktarplee:
When accessing the referrer's API, this adds a check to make sure the content-type is an OCI index.

Also fixes the tests to properly set the content-type header in the response.

Closes #633

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -483,18 +483,12 @@ func (r *Repository) Referrers(ctx context.Context, desc ocispec.Descriptor, art
 
 	// The referrers state is unknown.
 	if err != nil {
-		var errResp *errcode.ErrorResponse
-		if !errors.As(err, &errResp) || errResp.StatusCode != http.StatusNotFound {
-			return err
+		if errors.Is(err, errdef.ErrUnsupported) {
+			// Referrers API is not supported, fallback to referrers tag schema.
+			r.SetReferrersCapability(false)
+			return r.referrersByTagSchema(ctx, desc, artifactType, fn)
 		}
-		if errutil.IsErrorCode(errResp, errcode.ErrorCodeNameUnknown) {
-			// The repository is not found, no fallback.
-			return err
-		}
-		// A 404 returned by Referrers API indicates that Referrers API is
-		// not supported. Fallback to referrers tag schema.
-		r.SetReferrersCapability(false)
-		return r.referrersByTagSchema(ctx, desc, artifactType, fn)
+		return err
 	}
 
 	r.SetReferrersCapability(true)
@@ -544,8 +538,22 @@ func (r *Repository) referrersPageByAPI(ctx context.Context, artifactType string
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK {
+	switch resp.StatusCode {
+	case http.StatusOK:
+	case http.StatusNotFound:
+		if errResp := errutil.ParseErrorResponse(resp); errutil.IsErrorCode(errResp, errcode.ErrorCodeNameUnknown) {
+			// The repository is not found, Referrers API status is unknown
+			return "", errResp
+		}
+		// Referrers API is not supported.
+		return "", fmt.Errorf("failed to query referrers API: %w", errdef.ErrUnsupported)
+	default:
 		return "", errutil.ParseErrorResponse(resp)
+	}
+
+	// also check the content type
+	if ct := resp.Header.Get("Content-Type"); ct != ocispec.MediaTypeImageIndex {
+		return "", fmt.Errorf("unknown content returned (%s), expecting image index: %w", ct, errdef.ErrUnsupported)
 	}
 
 	var index ocispec.Index
@@ -657,8 +665,9 @@ func (r *Repository) pingReferrers(ctx context.Context) (bool, error) {
 
 	switch resp.StatusCode {
 	case http.StatusOK:
-		r.SetReferrersCapability(true)
-		return true, nil
+		supported := resp.Header.Get("Content-Type") == ocispec.MediaTypeImageIndex
+		r.SetReferrersCapability(supported)
+		return supported, nil
 	case http.StatusNotFound:
 		if err := errutil.ParseErrorResponse(resp); errutil.IsErrorCode(err, errcode.ErrorCodeNameUnknown) {
 			// repository not found
--- a/registry/remote/repository_test.go
+++ b/registry/remote/repository_test.go
@@ -1272,6 +1272,7 @@ func TestRepository_Predecessors(t *testing.T) {
 			MediaType: ocispec.MediaTypeImageIndex,
 			Manifests: referrers,
 		}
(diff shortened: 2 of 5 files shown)


registry/remote/repository.go
```diff
@@ -548,6 +548,11 @@ func (r *Repository) referrersPageByAPI(ctx context.Context, artifactType string
 		return "", errutil.ParseErrorResponse(resp)
 	}
 
+	if resp.Header.Get("Content-Type") != ocispec.MediaTypeImageIndex {
+		resp.StatusCode = http.StatusNotFound
```
ktarplee, line 552: I'm not sure this is the correct way to return this error.

registry/remote/repository.go
```diff
@@ -548,6 +548,11 @@ func (r *Repository) referrersPageByAPI(ctx context.Context, artifactType string
...
+		return "", errutil.ParseErrorResponse(resp)
```
shizhMSFT, line 553: We should modify the response status code. Instead, we can do something like
```go
err := errutil.ParseErrorResponse(resp)
return "", fmt.Errorf("unknown content returned, expecting image index: %w", err)
```
You may improve the error message :P

registry/remote/repository_test.go
```diff
@@ -7082,6 +7091,7 @@ func TestRepository_pingReferrers(t *testing.T) {
 		switch {
 		case r.Method == http.MethodGet && r.URL.Path == "/v2/test/referrers/"+zeroDigest:
 			count++
+			w.Header().Set("Content-Type", ocispec.MediaTypeImageIndex)
```
Wwwsylvia, line 7094: Can we have another test case for `pingReferrers` and `Referrers`, where `200` is returned with an unexpected content type?

ktarplee: I added two tests since there are two places in the code that have to check the referrer's content type.

registry/remote/repository.go
ktarplee: I am using `errors.Join` to return multiple errors from the function.

registry/remote/repository.go
ktarplee: This was handled by the discussion below.

ktarplee:
@Wwwsylvia @shizhMSFT This is ready from my perspective.  Once you review it I can update the branch to it is a single fast-forward commit if desired.

ktarplee:
@shizhMSFT Can we get this merged?

Wwwsylvia:
Just back from some other work, sorry for the delay! I'm reviewing it.

registry/remote/repository.go
```diff
@@ -544,10 +542,21 @@ func (r *Repository) referrersPageByAPI(ctx context.Context, artifactType string
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK {
+	switch resp.StatusCode {
+	case http.StatusOK:
+	case http.StatusNotFound:
+		// A 404 returned by Referrers API indicates that Referrers API is not supported.
+		return "", errors.Join(errdef.ErrUnsupported, errutil.ParseErrorResponse(resp))
```
Wwwsylvia, line 549: If the status code is 404 and the error code is `NAME_UNKNOWN`, it means that the repository does not exist, and the referrers API status is still unknown. For such case, we shouldn't return `ErrUnsupported`.

Wwwsylvia: How about:

```suggestion
		if errResp := errutil.ParseErrorResponse(resp); errutil.IsErrorCode(errResp, errcode.ErrorCodeNameUnknown) {
			// The repository is not found, Referrers API status is unknown
			return "", errResp
		}
		// Referrers API is not supported.
		return "", fmt.Errorf("failed to query referrers API: %w", errdef.ErrUnsupported)
```

registry/remote/repository.go
```diff
@@ -483,16 +483,14 @@ func (r *Repository) Referrers(ctx context.Context, desc ocispec.Descriptor, art
...
-		var errResp *errcode.ErrorResponse
-		if !errors.As(err, &errResp) || errResp.StatusCode != http.StatusNotFound {
+		if !errors.Is(err, errdef.ErrUnsupported) {
 			return err
 		}
-		if errutil.IsErrorCode(errResp, errcode.ErrorCodeNameUnknown) {
+		if errutil.IsErrorCode(err, errcode.ErrorCodeNameUnknown) {
 			// The repository is not found, no fallback.
 			return err
 		}
```
Wwwsylvia, line 492: How about:
If we move the check for the repository not found case to `referrersByAPI`, we can do:
```go
	if err != nil {
		if errors.Is(err, errdef.ErrUnsupported) {
			// Referrers API is not supported, fallback to referrers tag schema.
			r.SetReferrersCapability(false)
			return r.referrersByTagSchema(ctx, desc, artifactType, fn)
		}
		return err
	}
```

ktarplee:
@Wwwsylvia I liked you changes.  I tested them and it still resolves the original issue.   Let me know if there are any other changes or if you want me to squash the commits.

registry/remote/repository.go
```diff
@@ -483,16 +483,13 @@ func (r *Repository) Referrers(ctx context.Context, desc ocispec.Descriptor, art
...
-			return err
+		if errors.Is(err, errdef.ErrUnsupported) {
+			// Referrers API is not supported, fallback to referrers tag schema.
+			r.SetReferrersCapability(false)
+			return r.referrersByTagSchema(ctx, desc, artifactType, fn)
 		}
-		// A 404 returned by Referrers API indicates that Referrers API is
-		// not supported. Fallback to referrers tag schema.
+		return err
+		// Fallback to referrers tag schema.
```
Wwwsylvia, line 492: We need to remove line 492-494

shizhMSFT: Agree to remove the code in a dead path.

ktarplee:
@Wwwsylvia @shizhMSFT This is rebased and should be ready to be merged.

Wwwsylvia (approved):
LGTM

Issue #472, opened 2023-03-24
Feature Request: Consider supporting size management for OCI Store


akashsinghal:
Currently, there isn't built in functionality for deleting and managing OCI Store size. It would be nice if there was a garbage collection-like functionality to evict LRU blobs from the index and delete the corresponding blobs from the OCI store

shizhMSFT:
Duplicate of #454

akashsinghal:
Is there a plan to implement higher level function beyond Delete? Such as a way to add a time last accessed annotation in the index.json and an accompanying lastused descriptor method? Basically an API to make it easier to implement an LRU eviction strategy.

shizhMSFT:
> add a time last accessed annotation in the index.json and an accompanying last used descriptor method?

That seems an auto purge functionality and is currently out of the scope since `index.json` might be immutable or supposed to be read-only.

sparr:
While considering similar concerns in our code base, we considered using containerd's content store package, which offers [garbage collection](https://github.com/containerd/containerd/blob/main/docs/garbage-collection.md) controlled by labels.

Having support for some similar functionality in oras would be great, but at the moment I'm more interested in getting #454 so we can delete at all. Once it supports delete and updating the predecessor/successor graph, then we can at least identify unreferenced/orphaned resources.

shizhMSFT:
Removing the `duplicate` label as this issue request Garbage Collection in addition to #454

shizhMSFT:
cc: @wangxiaoxuan273 @eiffel-fl

wangxiaoxuan273:
Let's have some discussions on the following general questions, which choices work better with actual needs?

1. Do we want to implement auto-garbage collection in `Delete()`, or provide a public GC function and let user call it manually?
2. Do we need to clean garbage recursively? For example given the graph `A -> B -> C -> D`, when `A` is deleted, should we delete `B, C, D` or only `B`?

Wwwsylvia:
> 1. Do we want to implement auto-garbage collection in `Delete()`, or provide a public GC function and let user call it manually?

I think we can start with providing a public GC function. And then we can consider having a store option for enabling auto-GC. It could be somewhat similar to `SaveIndex()` and `AutoSaveIndex`.


> 2. Do we need to clean garbage recursively? For example given the graph `A -> B -> C -> D`, when `A` is deleted, should we delete `B, C, D` or only `B`?

[...]

PR #645, opened 2023-11-20
feat: extend oci.Store with DeleteTree().

eiffel-fl:
Hi!

Following the merge of #614, I opened this PR to propose `DeleteTree()`.
This function will delete the content matching the given descriptor but also its children if they do not have any other parent.
We use this code in Inspektor Gadget to delete OCI image:
https://github.com/inspektor-gadget/inspektor-gadget/pull/2162
https://github.com/inspektor-gadget/inspektor-gadget/pull/2162/commits/248811856471fb36f4befe36af77a27ee57249b2#diff-570aea5baf5f0e0b7f46b570a48f173e8858df3436f091fb61d2e3c745db6f0eR298
While it is still not merged at the moment, it was nonetheless validated.

If you see any way to improve this PR, feel free to share your feedback.

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -168,6 +168,43 @@ func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
 	return s.storage.Delete(ctx, target)
 }
 
+// DeleteTree deletes the content matching the descriptor as well as its
+// children from the store.
+// Children will be deleted only if they do not any other predecessors than the
+// given root.
+// DeleteTree may fail on certain systems (i.e. NTFS), if there is a process
+// (i.e. an unclosed Reader) using any of root children.
+func (s *Store) DeleteTree(ctx context.Context, root ocispec.Descriptor) error {
+	predecessors, err := s.Predecessors(ctx, root)
+	if err != nil {
+		return fmt.Errorf("getting predecessors: %w", err)
+	}
+
+	// We need to check if there are more than 1 predecessor as the current tree
+	// is counted in.
+	if len(predecessors) > 1 {
+		return nil
+	}
+
+	descriptors, err := content.Successors(ctx, s, root)
+	if err != nil {
+		return fmt.Errorf("getting successors: %w", err)
+	}
+
+	for _, descriptor := range descriptors {
+		if content.Equal(descriptor, ocispec.DescriptorEmptyJSON) {
+			continue
+		}
+
+		err := s.DeleteTree(ctx, descriptor)
+		if err != nil {
+			return err
+		}
+	}
+
+	return s.Delete(ctx, root)
+}
+
 // Tag tags a descriptor with a reference string.
 // reference should be a valid tag (e.g. "latest").
 // Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc5/image-layout.md#indexjson-file
--- a/content/oci/oci_test.go
+++ b/content/oci/oci_test.go
@@ -2237,6 +2237,103 @@ func TestStore_PredecessorsAndDelete(t *testing.T) {
 	}
 }
 
+func TestStore_PredecessorsAndDeleteTree(t *testing.T) {
+	tempDir := t.TempDir()
+	s, err := New(tempDir)
+	if err != nil {
+		t.Fatal("New() error =", err)
+	}
+	ctx := context.Background()
+
+	// generate test content
+	var blobs [][]byte
+	var descs []ocispec.Descriptor
+	appendBlob := func(mediaType string, blob []byte) {
+		blobs = append(blobs, blob)
+		descs = append(descs, ocispec.Descriptor{
+			MediaType: mediaType,
+			Digest:    digest.FromBytes(blob),
+			Size:      int64(len(blob)),
+		})
+	}
+	generateManifest := func(config ocispec.Descriptor, layers ...ocispec.Descriptor) {
+		manifest := ocispec.Manifest{
+			Config: config,
+			Layers: layers,
+		}
+		manifestJSON, err := json.Marshal(manifest)
+		if err != nil {
+			t.Fatal(err)
+		}
+		appendBlob(ocispec.MediaTypeImageManifest, manifestJSON)
+	}
+	generateIndex := func(manifests ...ocispec.Descriptor) {
+		index := ocispec.Index{
+			Manifests: manifests,
+		}
+		indexJSON, err := json.Marshal(index)
+		if err != nil {
+			t.Fatal(err)
+		}
+		appendBlob(ocispec.MediaTypeImageIndex, indexJSON)
+	}
+
+	appendBlob(ocispec.MediaTypeImageLayer, []byte("foo"))     // Blob 0
+	appendBlob(ocispec.MediaTypeImageLayer, []byte("bar"))     // Blob 1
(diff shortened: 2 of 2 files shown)


content/oci/oci.go
```diff
@@ -168,6 +168,34 @@ func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
...
+		return fmt.Errorf("getting image predecessors: %w", err)
+	}
+
+	if len(predecessors) > 1 {
+		return nil
+	}
+
+	descriptors, err := content.Successors(ctx, s, root)
+	if err != nil {
+		return fmt.Errorf("getting image successors: %w", err)
```
TerryHowe, line 189:
```suggestion
		return fmt.Errorf("getting successors: %w", err)
```

content/oci/oci.go
```diff
@@ -168,6 +168,34 @@ func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
...
+// DeleteTree deletes the content matching the descriptor as well as its
+// children from the store.
+// Children will be deleted only if they do not any other predecessors than the
+// given root.
+// DeleteTree may fail on certain systems (i.e. NTFS), if there is a process
+// (i.e. an unclosed Reader) using any of root children.
+func (s *Store) DeleteTree(ctx context.Context, root ocispec.Descriptor) error {
+	predecessors, err := s.Predecessors(ctx, root)
+	if err != nil {
+		return fmt.Errorf("getting image predecessors: %w", err)
```
TerryHowe, line 180:
```suggestion
		return fmt.Errorf("getting predecessors: %w", err)
```
Or maybe something other than "image" I would think

shizhMSFT:
@eiffel-fl Thank you for proposing the `DeleteTree()`.

Instead of deleting a tree, I'm thinking about garbage collection (GC).

There are many benefits of GC over simply deleting a tree.
- GC can be performed after several combination of `Push()` and `Delete()` operations.
- GC can clean up any dangling nodes.

The basic idea is to mimic the garbage collection behavior of a real registry.

Related:
- #472

/cc @wangxiaoxuan273 @Wwwsylvia

wangxiaoxuan273:
@eiffel-fl Thanks for opening this PR. Let's make future discussions on the #472 issue page, so that more stakeholders can participate.

Wwwsylvia:
@eiffel-fl Given that #653 has been merged, shall we close this PR? Thanks for bringing this up!

Wwwsylvia:
Hi @eiffel-fl , can you help us to understand the relationship between those blobs you listed?

```sh
$sudo ls /var/lib/ig/oci-store/blobs/sha256  
225a8e697e260684d9451ccf8a622cd5548b0ae5fac6e13c416cd1d85247b7b7  b4be31f07444fb9be6dee8cd1349aab3c780d333b63406be10c5c14c1b39dd37
662ca58536d5a03864781eb42870f6712927981383f1dab1449d9adf80067313  c06a86dad7e582686b53bec9c729b92fc56c838b59cc11d0105c2c20f3158dcd
b0eff7f0ec2a27b8d71fe904cf373fa2002c6d669db54c25a9542cf2cb5f46d1  f61eef5ff800b2e4366c6d2e7311e95ce8dc3cff5c020a3a810429656fd93efb
```

A [mermaid diagram](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/) would be very helpful.

Wwwsylvia:
@eiffel-fl I have some further questions:

- What are their media types?
- Is any of them a `subject` of another one?
- Which of them are recorded in `index.json`?

Wwwsylvia:
Hi @eiffel-fl , thanks for the clarification. I now understand what's happening.

**TL;DR: the behavior you observed is by design. We don't automatically delete the successor manifests along with the deleted index, if they are present in `index.json`.**

Let me put the diagram in another way:

```mermaid
graph TD;
b0eff7f0ec2a27b8d71fe904cf373fa2002c6d669db54c25a9542cf2cb5f46d1[Index:latest] --> 662ca58536d5a03864781eb42870f6712927981383f1dab1449d9adf80067313
b0eff7f0ec2a27b8d71fe904cf373fa2002c6d669db54c25a9542cf2cb5f46d1 --> f61eef5ff800b2e4366c6d2e7311e95ce8dc3cff5c020a3a810429656fd93efb
662ca58536d5a03864781eb42870f6712927981383f1dab1449d9adf80067313[Manifest:amd64] --> b4be31f07444fb9be6dee8cd1349aab3c780d333b63406be10c5c14c1b39dd37
662ca58536d5a03864781eb42870f6712927981383f1dab1449d9adf80067313 --> 225a8e697e260684d9451ccf8a622cd5548b0ae5fac6e13c416cd1d85247b7b7
f61eef5ff800b2e4366c6d2e7311e95ce8dc3cff5c020a3a810429656fd93efb[Manifest:arm64] --> b4be31f07444fb9be6dee8cd1349aab3c780d333b63406be10c5c14c1b39dd37
f61eef5ff800b2e4366c6d2e7311e95ce8dc3cff5c020a3a810429656fd93efb --> c06a86dad7e582686b53bec9c729b92fc56c838b59cc11d0105c2c20f3158dcd 

```

We know that:
-  `Index:latest` references two manifests: `Manifest:amd64` and `Manifest:arm64`
- `Index:latest`, `Manifest:amd64`, and `Manifest:arm64` are all recorded in `index.json`

When one is deleting `Index:latest`, they may expect:
- Manifest:amd64` and `Manifest:arm64` also get deleted
- Manifest:amd64` and `Manifest:arm64` are not deleted (they wants to manually delete `Manifest:amd64` later and keep `Manifest:arm64`)

`oras-go` could not tell the user's purpose, so it chooses not to automatically delete `Manifest:amd64`  and `Manifest:arm64`.
To delete the whole graph, you need to manually delete `Manifest:amd64` and `Manifest:arm64` after deleting `Index:latest`.

The idea is that `oras-go` considers everything in `index.json` is meaningful.
That is to say, if `Manifest:amd64` and `Manifest:arm64` were not present in `index.json`, they would get automatically deleted together with `Index:latest`.

Does this make sense to you?

Wwwsylvia:
BTW, may I know what tool did you use to generate this OCI layout?

Wwwsylvia:
> > 2. Is any of them a `subject` of another one?
> 
> Can you please precise how can I get this information 😅?

Oh you can just look into the manifests and see if they have a `subject` field, like:
```sh
cat /var/lib/ig/oci-store/blobs/sha256/662ca58536d5a03864781eb42870f6712927981383f1dab1449d9adf80067313 | jq .
```

But in this particular case, the answer is that: none of them is a [`subject`](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc5/manifest.md#image-manifest-property-descriptions) of another. 😀

Wwwsylvia:
> Now, I am really wondering if we do not do something wrong as I find it odd to have the architecture manifests in index.json, as, from my understanding, the index.json should only contain index which do the abstraction for the architectures.

Actually, it is `oras-go` to blame!
Currently `oras-go` will automatically add manifests into `index.json`, no matter it has a tag or not. This was to make sure referrers manifests without tags were still trackable.
We might need to improve this design.

Wwwsylvia:
This is not a bug, and it is not straightforward to change it🤔.
We need to work out a more proper design for this. Would you like to open an issue for discussion?

Wwwsylvia:
@eiffel-fl Thank you!

PR #647, opened 2023-11-21
feat: add `Untag()` to `*oci.Store`

eiffel-fl:
Hi!

`oci.Store` offers the possibility to tag images but not to untag them.
This is a problem when you want to remove images because you only want to untag a tag when several point to the same image (_e.g._ `docker rmi` and see https://github.com/inspektor-gadget/inspektor-gadget/pull/2162#pullrequestreview-1740423646).

I will add proper testing if we agree that this addition is useful to the package.

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -238,6 +238,29 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
 	return desc, nil
 }
 
+func (s *Store) Untag(ctx context.Context, reference string) error {
+	if reference == "" {
+		return errdef.ErrMissingReference
+	}
+
+	s.sync.RLock()
+	defer s.sync.RUnlock()
+
+	desc, err := s.tagResolver.Resolve(ctx, reference)
+	if err != nil {
+		return fmt.Errorf("resolving reference %q: %w", reference, err)
+	}
+	if reference == desc.Digest.String() {
+		return fmt.Errorf("reference %q is a digest and not a tag: %w", reference, errdef.ErrInvalidReference)
+	}
+
+	s.tagResolver.Untag(reference)
+	if s.AutoSaveIndex {
+		return s.saveIndex()
+	}
+	return nil
+}
+
 // Predecessors returns the nodes directly pointing to the current node.
 // Predecessors returns nil without error if the node does not exists in the
 // store.
--- a/content/oci/oci_test.go
+++ b/content/oci/oci_test.go
@@ -575,7 +575,8 @@ func TestStore_DisableAutoSaveIndex(t *testing.T) {
 		Digest:    digest.FromBytes(content),
 		Size:      int64(len(content)),
 	}
-	ref := "foobar"
+	ref0 := "foobar"
+	ref1 := "barfoo"
 
 	tempDir := t.TempDir()
 	s, err := New(tempDir)
@@ -623,16 +624,20 @@ func TestStore_DisableAutoSaveIndex(t *testing.T) {
 	}
 
 	// test tag
-	err = s.Tag(ctx, desc, ref)
+	err = s.Tag(ctx, desc, ref0)
 	if err != nil {
 		t.Fatal("Store.Tag() error =", err)
 	}
-	if got, want := len(internalResolver.Map()), 2; got != want {
+	err = s.Tag(ctx, desc, ref1)
+	if err != nil {
+		t.Fatal("Store.Tag() error =", err)
+	}
+	if got, want := len(internalResolver.Map()), 3; got != want {
 		t.Errorf("resolver.Map() = %v, want %v", got, want)
 	}
 
 	// test resolving by digest
-	gotDesc, err = s.Resolve(ctx, ref)
+	gotDesc, err = s.Resolve(ctx, ref0)
 	if err != nil {
 		t.Fatal("Store.Resolve() error =", err)
 	}
@@ -648,11 +653,30 @@ func TestStore_DisableAutoSaveIndex(t *testing.T) {
 		t.Fatal("Store.SaveIndex() error =", err)
 	}
 	// test index file again
-	if got, want := len(s.index.Manifests), 1; got != want {
+	if got, want := len(s.index.Manifests), 2; got != want {
 		t.Errorf("len(index.Manifests) = %v, want %v", got, want)
 	}
 	if _, err := os.Stat(s.indexPath); err != nil {
 		t.Errorf("error: %s does not exist", s.indexPath)
+	}
+
+	// test untag
+	err = s.Untag(ctx, ref0)
+	if err != nil {
+		t.Fatal("Store.Untag() error =", err)
+	}
+	if got, want := len(internalResolver.Map()), 2; got != want {
+		t.Errorf("resolver.Map() = %v, want %v", got, want)
+	}
+	if got, want := len(s.index.Manifests), 2; got != want {
+		t.Errorf("len(index.Manifests) = %v, want %v", got, want)
+	}
+	if err := s.saveIndex(); err != nil {
+		t.Fatal("Store.SaveIndex() error =", err)
+	}
+	// test index file again
+	if got, want := len(s.index.Manifests), 1; got != want {
+		t.Errorf("len(index.Manifests) = %v, want %v", got, want)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


content/oci/oci.go
```diff
@@ -238,6 +238,38 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
 	return desc, nil
 }
 
+func (s *Store) Untag(ctx context.Context, reference string) error {
+	s.sync.RLock()
+	defer s.sync.RUnlock()
+
+	if reference == "" {
```
qweeah, line 245: Do the check before locking?

wangxiaoxuan273:
Hi @eiffel-fl, for a feature like this, we need more discussion and design. We encourage you to open an issue, so that everyone can discuss ideas and scenarios on the issue page.

Best regards.

content/oci/oci.go
wangxiaoxuan273: I prefer doing the check after locking, since this ensures style consistency with `Tag()` and other functions.

wangxiaoxuan273:
Hi, I think your original implementation was good (before you change the input from `reference` to `digest`), how about changing it back and continue on that one?

content/oci/oci.go
```diff
@@ -238,6 +238,30 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
...
+		return errdef.ErrMissingReference
+	}
+
+	_, err := s.tagResolver.Resolve(ctx, reference)
+	if err != nil {
+		return fmt.Errorf("reference %q does not point to a descriptor: %w", reference, err)
+	}
```
wangxiaoxuan273, line 252:
```suggestion
	desc, err := s.tagResolver.Resolve(ctx, reference)

	// if the reference does not exist or is a digest, do nothing
	if err != nil || reference == desc.Digest.String() {
		return fmt.Errorf("reference not found")
	}
```

content/resolver.go
```diff
@@ -39,3 +39,9 @@ type TagResolver interface {
 	Tagger
 	Resolver
 }
+
+// Untagger untags reference tags.
+type Untagger interface {
+	// Untag untags the given reference string.
+	Untag(ctx context.Context, reference string) error
+}
```
Wwwsylvia, line 47: nit: need an empty line at the end of file.

content/oci/oci.go
```diff
@@ -238,6 +238,33 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
...
+	if reference == "" {
+		return errdef.ErrMissingReference
+	}
+
+	desc, err := s.tagResolver.Resolve(ctx, reference)
+	if err != nil {
+		return fmt.Errorf("reference %q does not point to a descriptor: %w", reference, err)
+	}
+	if reference == desc.Digest.String() {
+		return fmt.Errorf("reference %q is a digest and not a tag", reference)
```
Wwwsylvia, line 254: Should we return `errdef.ErrInvalidReference` here?

content/oci/oci.go
```diff
@@ -238,6 +238,33 @@ func (s *Store) Resolve(ctx context.Context, reference string) (ocispec.Descript
...
+	}
+
+	return s.untag(reference)
+}
+
+func (s *Store) untag(reference string) error {
```
Wwwsylvia, line 260: Wondering if we really need a separate function for this logic?

Wwwsylvia:
Can we follow the [Conventional Commits ](https://www.conventionalcommits.org/en/v1.0.0/#specification) to have the PR title start with "feat:" ?

wangxiaoxuan273:
LGTM but I'm not a maintainer. For style consistency, can you change the PR title to `feat: add Untag() to oci.Store`?

content/oci/oci_test.go
```diff
@@ -654,6 +672,15 @@ func TestStore_DisableAutoSaveIndex(t *testing.T) {
 	if _, err := os.Stat(s.indexPath); err != nil {
 		t.Errorf("error: %s does not exist", s.indexPath)
 	}
+
+	// test untag
+	err = s.Untag(ctx, ref)
+	if err != nil {
+		t.Fatal("Store.Untag() error =", err)
+	}
+	if got, want := len(internalResolver.Map()), 1; got != want {
```
Wwwsylvia, line 681: We should test like line 662-671

content/oci/oci_test.go
```diff
@@ -333,6 +342,15 @@ func TestStore_RelativeRoot_Success(t *testing.T) {
...
+		t.Errorf("resolver.Map() = %v, want %v", got, want)
+	}
```
Wwwsylvia, line 353: We should test `Resolve` again after untagging the reference.

content/oci/oci_test.go
```diff
@@ -204,6 +204,15 @@ func TestStore_Success(t *testing.T) {
...
```
Wwwsylvia, line 215: We should test `Resolve` again after untagging the reference.

content/oci/oci_test.go
Wwwsylvia: I think we can tag the descriptor with 2 different references at the beginning, so we have 3 entries in the internal resolver:
```
"foo" -> desc
"bar" -> desc
desc.Digest -> desc
```
After manually calling `saveIndex`, we have the 2 entries in `index.Manifests`:
```
desc ["org.opencontainers.image.title": "foo"]
desc ["org.opencontainers.image.title": "bar"]
```

And then we untag "bar", the internal resolver map becomes:
```
"foo" -> desc
desc.Digest -> desc
```
While `index.Manifests` has no change because we disabled auto save.

After manually calling `saveIndex`, we have only 1 entry left in `index.Manifests`:
```
desc ["org.opencontainers.image.title": "foo"]
```

content/oci/oci_test.go
```diff
@@ -333,6 +346,19 @@ func TestStore_RelativeRoot_Success(t *testing.T) {
 	if !bytes.Equal(got, manifest) {
 		t.Errorf("Store.Fetch() = %v, want %v", got, manifest)
 	}
+
+	// test untag
```
Wwwsylvia, line 350: We don't need this anymore if we have a dedicated method for testing `Untag`.

Wwwsylvia (commented):
Sorry for the delayed review! Left a few suggestions regarding the unit tests. Let me know if anything is unclear.

content/oci/oci_test.go
```diff
@@ -2237,6 +2274,123 @@ func TestStore_PredecessorsAndDelete(t *testing.T) {
...
+	tempDir := t.TempDir()
+	s, err := New(tempDir)
+	if err != nil {
+		t.Fatal("NewDeletableStore() error =", err)
+	}
+	ctx := context.Background()
+
+	err = s.Untag(ctx, "")
+	if err == nil {
+		t.Errorf("Store.Untag() error = %v, wantErr %v", err, errdef.ErrMissingReference)
```
Wwwsylvia, line 2357: The error can potentially be of other type which is not nil and not `ErrMissingReference`, so it is better to test if the error type matches.

```suggestion
	if !errors.Is(err, errdef.ErrMissingReference) {
		t.Errorf("Store.Untag() error = %v, wantErr %v", err, errdef.ErrMissingReference)
```

This also applies to other similar places.

content/oci/oci_test.go
```diff
@@ -2237,6 +2274,123 @@ func TestStore_PredecessorsAndDelete(t *testing.T) {
...
+	if !reflect.DeepEqual(resolvedDescr, desc) {
+		t.Errorf("Store.Resolve() = %v, want %v", resolvedDescr, desc)
+	}
+
+	err = s.Untag(ctx, ref)
+	if err != nil {
+		t.Errorf("Store.Untag() = %v, wantErr %v", err, nil)
+	}
+
+	_, err = s.Resolve(ctx, ref)
```
Wwwsylvia, line 2325: Same here, we should test the error type.

content/oci/oci_test.go
```diff
@@ -204,6 +204,19 @@ func TestStore_Success(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("Store.Untag() error =", err)
+	}
+	gotDesc, err = s.Resolve(ctx, ref)
+	if err == nil {
+		t.Fatalf("error expected as reference was untagged, got %v", gotDesc)
+	}
+	if got, want := len(internalResolver.Map()), 1; got != want {
+		t.Errorf("resolver.Map() = %v, want %v", got, want)
+	}
```
Wwwsylvia, line 219: Given that we now have `TestStore_Untag`, I think this can be removed.

Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

Wwwsylvia:
@eiffel-fl Merged! Thank you for your contribution!

PR #651, opened 2023-11-28
feat: add a host-based auth cache as a fallback

### LINKED ISSUE #650: Prevent extra authentication round trips in non-compliant registries
I have been testing with JFrog Artifactory and `oras cp` will receive a 401 for each and every access to the repo that it has to go resolve by following the auth challenge and then use the received token (actually it is the same as stored in the auth cache already but under a different scoped key) to make the request and succeed.  While the current behavior works, it is obviously inefficient.  `crane` and `docker` use a simplified host-based caching (not scope based caching like oras) so they do not have an issue in this specific case.

The discussion over on [slack](https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1699538327616879) has lots of details.

ktarplee:
This PR adds a new exported method called `NewRobustCache()` and changes the `DefaultCache` to use the robust cache.  The robust cache uses scoped based auth but falls back to host based auth to better handle the situations described in #650 but retain the benefits of scoped based auth (where the token for a repo might be different than another repo in the same registry).

Closes #650

### DIFF
--- a/registry/remote/auth/cache.go
+++ b/registry/remote/auth/cache.go
@@ -157,3 +157,76 @@ func (noCache) GetToken(ctx context.Context, registry string, scheme Scheme, key
 func (noCache) Set(ctx context.Context, registry string, scheme Scheme, key string, fetch func(context.Context) (string, error)) (string, error) {
 	return fetch(ctx)
 }
+
+// hostCache is an auth cache that ignores scopes.  Uses only the registry's hostname to find a token.
+type hostCache struct {
+	Cache
+}
+
+// GetToken implements Cache.
+func (c *hostCache) GetToken(ctx context.Context, registry string, scheme Scheme, key string) (string, error) {
+	return c.Cache.GetToken(ctx, registry, scheme, "")
+}
+
+// Set implements Cache.
+func (c *hostCache) Set(ctx context.Context, registry string, scheme Scheme, key string, fetch func(context.Context) (string, error)) (string, error) {
+	return c.Cache.Set(ctx, registry, scheme, "", fetch)
+}
+
+// fallbackCache tries the primary cache then falls back to the secondary cache.
+type fallbackCache struct {
+	primary   Cache
+	secondary Cache
+}
+
+// GetScheme implements Cache.
+func (fc *fallbackCache) GetScheme(ctx context.Context, registry string) (Scheme, error) {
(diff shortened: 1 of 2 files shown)


ktarplee:
This PR obviously still needs 

- [ ] Unit tests
- [ ] Integration testing
- [ ] Documentation

registry/remote/auth/cache.go
```diff
@@ -157,3 +157,67 @@ func (noCache) GetToken(ctx context.Context, registry string, scheme Scheme, key
...
+	}
+
+	// fallback
+	return fc.secondary.GetScheme(ctx, registry)
+}
+
+// GetToken implements Cache.
+func (fc *fallbackCache) GetToken(ctx context.Context, registry string, scheme Scheme, key string) (string, error) {
+	token, err := fc.primary.GetToken(ctx, registry, scheme, key)
+	if err == nil {
```
ktarplee, line 196: Not sure if we want to check for `errdef.ErrNotFound` and return other errors.

shizhMSFT:
@ktarplee Would you like to update this PR to reflect the latest discussion in #650?

ktarplee:
> @ktarplee Would you like to update this PR to reflect the latest discussion in #650?

I made the name change.

ktarplee:
I also added unit tests for the hostCache and exported `NewSingleContextCache()` function. @shizhMSFT

registry/remote/auth/cache.go
```diff
@@ -157,3 +157,76 @@ func (noCache) GetToken(ctx context.Context, registry string, scheme Scheme, key
...
+
+	// fallback
+	return fc.secondary.GetToken(ctx, registry, scheme, key)
+}
+
+// Set implements Cache.
+func (fc *fallbackCache) Set(ctx context.Context, registry string, scheme Scheme, key string, fetch func(context.Context) (string, error)) (string, error) {
+	token, err := fc.primary.Set(ctx, registry, scheme, key, fetch)
+	if err != nil {
+		return token, err
```
shizhMSFT, line 208: nit:
```suggestion
		return "", err
```

shizhMSFT (approved):
LGTM and thanks for the update.

shizhMSFT:
@ktarplee LoL. The DCO check failed. Could you fix that?

shizhMSFT (approved):
LGTM

PR #654, opened 2023-12-06
Update example_test.go: added :443

guettli:
OCI-Registries can deny all authenticated requests done via http, since it is unencrypted.

It took some time for us to discover that oras does http by default.


shizhMSFT:
@guettli The following statement is not true.

> It took some time for us to discover that oras does http by default.

Instead, `oras-go` always do **HTTPS** request by default as the following code snippet:

https://github.com/oras-project/oras-go/blob/15ee0be60ddc4a54b285044ad693fd35a9899549/registry/remote/url.go#L28-L33

The `oras-go` library do HTTP request only if [`PlainHTTP` option](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.1/registry/remote#Repository) is toggled.

Additionally, it is worth noting that the HTTP/HTTPS protocol is not determined by the port but the protocol scheme. For instance,
- `https://myregistry.example.com:80/` goes HTTPS although the port is `80`.
- `http://myregistry.example.com:443/` goes HTTP although the port is `443`.

I'm also curious on how you discover that oras does HTTP by default. Could you share some code snippet of your project so that we can investigate to improve the overall experience of `oras-go`?

PR #656, opened 2023-12-08
feat: Public GC function of `oci.Store`

wangxiaoxuan273:
Part of #472

### DIFF
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"sync"
 
@@ -480,31 +481,42 @@ func (s *Store) GC(ctx context.Context) error {
 
 	// clean up garbage blobs in the storage
 	rootpath := filepath.Join(s.root, "blobs")
-	return filepath.Walk(rootpath, func(path string, info fs.FileInfo, err error) error {
+	algDirs, err := os.ReadDir(rootpath)
+	if err != nil {
+		return err
+	}
+	for _, algDir := range algDirs {
+		alg := algDir.Name()
+		// skip unsupported directories
+		if !isValidAlgorithm(alg) {
+			continue
+		}
+		algPath := path.Join(rootpath, alg)
+		dgstDirs, err := os.ReadDir(algPath)
 		if err != nil {
 			return err
 		}
-		// skip the directories
-		if info.IsDir() {
-			return nil
-		}
-		alg := filepath.Base(filepath.Dir(path))
-		blobDigest, err := digest.Parse(fmt.Sprintf("%s:%s", alg, info.Name()))
-		if err != nil {
-			return err
-		}
-		if exists := s.graph.Exists(blobDigest); !exists {
-			// remove the blob from storage if it does not exist in Store
-			err = os.Remove(path)
+		for _, dgstDir := range dgstDirs {
+			dgst := dgstDir.Name()
+			blobDigest := digest.NewDigestFromEncoded(digest.Algorithm(alg), dgst)
+			err := blobDigest.Validate()
+			// skip unsupported directories
 			if err != nil {
-				if errors.Is(err, fs.ErrNotExist) {
-					return fmt.Errorf("%s: %w", blobDigest, errdef.ErrNotFound)
+				continue
+			}
+			if exists := s.graph.Exists(blobDigest); !exists {
+				// remove the blob from storage if it does not exist in Store
+				err = os.Remove(path.Join(algPath, dgst))
+				if err != nil {
+					if errors.Is(err, fs.ErrNotExist) {
+						return fmt.Errorf("%s: %w", blobDigest, errdef.ErrNotFound)
+					}
+					return err
 				}
-				return err
 			}
 		}
-		return nil
-	})
+	}
+	return nil
 }
 
 // unsafeStore is used to bypass lock restrictions in Delete.
@@ -528,4 +540,9 @@ func validateReference(ref string) error {
 
 	// TODO: may enforce more strict validation if needed.
 	return nil
+}
+
+// isValidAlgorithm checks is a string is a supported hash algorithm
+func isValidAlgorithm(alg string) bool {
+	return alg == string(digest.SHA256) || alg == string(digest.SHA512) || alg == string(digest.SHA384)
 }
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -465,6 +465,9 @@ func (s *Store) traverseIndex(ctx context.Context) (set.Set[digest.Digest], erro
 	queue := []ocispec.Descriptor{}
 	queue = append(queue, manifests...)
 	for len(queue) > 0 {
+		if err := isContextDone(ctx); err != nil {
+			return nil, err
+		}
 		head := queue[0]
 		queue = queue[1:]
 		if visited.Contains(head.Digest) {
@@ -521,6 +524,9 @@ func (s *Store) GC(ctx context.Context) error {
 			return err
 		}
 		for _, dgstDir := range dgstDirs {
+			if err := isContextDone(ctx); err != nil {
+				return err
+			}
 			dgst := dgstDir.Name()
 			blobDigest := digest.NewDigestFromEncoded(digest.Algorithm(alg), dgst)
 			err := blobDigest.Validate()
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -170,17 +170,13 @@ func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
 	s.sync.Lock()
 	defer s.sync.Unlock()
 
-	return s.delete(ctx, target, s.AutoGC, s.AutoDeleteReferrers)
-}
-
-func (s *Store) delete(ctx context.Context, target ocispec.Descriptor, autoGC bool, autoDeleteReferrers bool) error {
 	deleteQueue := []ocispec.Descriptor{target}
 	for len(deleteQueue) > 0 {
 		head := deleteQueue[0]
 		deleteQueue = deleteQueue[1:]
 
 		// get referrers if applicable
-		if autoDeleteReferrers && descriptor.IsManifest(head) {
+		if s.AutoDeleteReferrers && descriptor.IsManifest(head) {
 			referrers, err := registry.Referrers(ctx, &unsafeStore{s}, head, "")
 			if err != nil {
 				return err
@@ -189,11 +185,11 @@ func (s *Store) delete(ctx context.Context, target ocispec.Descriptor, autoGC bo
 		}
 
 		// delete the head of queue
-		danglings, err := s.deleteNode(ctx, head)
+		danglings, err := s.delete(ctx, head)
 		if err != nil {
 			return err
 		}
-		if autoGC {
+		if s.AutoGC {
 			for _, d := range danglings {
 				// do not delete existing manifests in tagResolver
 				_, err = s.tagResolver.Resolve(ctx, string(d.Digest))
@@ -207,8 +203,8 @@ func (s *Store) delete(ctx context.Context, target ocispec.Descriptor, autoGC bo
 	return nil
 }
 
-// deleteNode deletes one node and returns the dangling nodes caused by the delete.
-func (s *Store) deleteNode(ctx context.Context, target ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+// delete deletes one node and returns the dangling nodes caused by the delete.
+func (s *Store) delete(ctx context.Context, target ocispec.Descriptor) ([]ocispec.Descriptor, error) {
 	resolvers := s.tagResolver.Map()
 	untagged := false
 	for reference, desc := range resolvers {
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -465,9 +465,6 @@ func (s *Store) traverseIndex(ctx context.Context) (set.Set[digest.Digest], erro
 	queue := []ocispec.Descriptor{}
 	queue = append(queue, manifests...)
 	for len(queue) > 0 {
-		if err := isContextDone(ctx); err != nil {
-			return nil, err
-		}
 		head := queue[0]
 		queue = queue[1:]
 		if visited.Contains(head.Digest) {
@@ -524,10 +521,6 @@ func (s *Store) GC(ctx context.Context) error {
 			return err
 		}
 		for _, dgstDir := range dgstDirs {
-			if err := isContextDone(ctx); err != nil {
-				return err
-			}
-
 			dgst := dgstDir.Name()
 			blobDigest := digest.NewDigestFromEncoded(digest.Algorithm(alg), dgst)
 			err := blobDigest.Validate()
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -489,10 +489,9 @@ func (s *Store) GC(ctx context.Context) error {
 		return err
 	}
 	for _, algDir := range algDirs {
-		// if !algDir.IsDir() {
-		// 	fmt.Println("got here")
-		// 	continue
-		// }
+		if !algDir.IsDir() {
+			continue
+		}
 		alg := algDir.Name()
 		// skip unsupported directories
 		if !isKnownAlgorithm(alg) {
@@ -501,7 +500,7 @@ func (s *Store) GC(ctx context.Context) error {
 		algPath := path.Join(rootpath, alg)
 		dgstDirs, err := os.ReadDir(algPath)
 		if err != nil {
-			continue
+			return err
 		}
 		for _, dgstDir := range dgstDirs {
 			if err := isContextDone(ctx); err != nil {
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -459,7 +459,6 @@ func (s *Store) writeIndexFile() error {
 // successor and referrer relations. It returns a set of digest of visited nodes.
 func (s *Store) traverseIndex(ctx context.Context) (set.Set[digest.Digest], error) {
 	manifests := s.index.Manifests
-	visited := set.New[digest.Digest]()
 	results := set.New[digest.Digest]()
 	queue := []ocispec.Descriptor{}
 	queue = append(queue, manifests...)
@@ -469,11 +468,10 @@ func (s *Store) traverseIndex(ctx context.Context) (set.Set[digest.Digest], erro
 		}
 		head := queue[0]
 		queue = queue[1:]
-		if visited.Contains(head.Digest) {
+		if results.Contains(head.Digest) {
 			continue
 		}
 		results.Add(head.Digest)
-		visited.Add(head.Digest)
 		// find successors
 		succ, err := content.Successors(ctx, &unsafeStore{s}, head)
 		if err != nil {
--- a/content/oci/oci.go
+++ b/content/oci/oci.go
@@ -485,17 +485,22 @@ func (s *Store) GC(ctx context.Context) error {
 			return err
 		}
 		// skip the directories
-		if !info.IsDir() {
-			alg := filepath.Base(filepath.Dir(path))
-			blobDigest, err := digest.Parse(fmt.Sprintf("%s:%s", alg, info.Name()))
+		if info.IsDir() {
+			return nil
+		}
+		alg := filepath.Base(filepath.Dir(path))
+		blobDigest, err := digest.Parse(fmt.Sprintf("%s:%s", alg, info.Name()))
+		if err != nil {
+			return err
+		}
+		if exists := s.graph.Exists(blobDigest); !exists {
+			// remove the blob from storage if it does not exist in Store
+			err = os.Remove(path)
 			if err != nil {
-				return err
-			}
-			if exists := s.graph.Exists(blobDigest); !exists {
-				// remove the blob from storage if it does not exist in Store
-				if err := s.storage.deleteByDigest(ctx, blobDigest); err != nil {
-					return err
+				if errors.Is(err, fs.ErrNotExist) {
+					return fmt.Errorf("%s: %w", blobDigest, errdef.ErrNotFound)
 				}
+				return err
 			}
 		}
 		return nil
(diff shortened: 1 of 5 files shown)


content/oci/oci.go
```diff
@@ -52,12 +52,20 @@ type Store struct {
...
 	//   - Default value: true.
 	AutoSaveIndex bool
-	root          string
-	indexPath     string
-	index         *ocispec.Index
-	storage       *Storage
-	tagResolver   *resolver.Memory
-	graph         *graph.Memory
+
+	// AutoGC controls if the OCI store will automatically clean newly produced
```
qweeah, line 56: Just curious, what does **newly produced** mean?

wangxiaoxuan273: It means the new dangling nodes produced by this operation of `Delete()`.

internal/graph/memory.go
```diff
@@ -20,29 +20,45 @@ import (
...
+	//    the other fields.
+	// - Memory.predecessors:
+	//  1. a node exists in Memory.predecessors if it has at least one predecessor
+	//    in the memory, regardless of whether or not the node itself exists in
+	//    the memory.
+	//  2. a node does not exist in Memory.nodes, if it doesn't have any predecessors
+	//    in the memory.
+	// - Memory.successors:
+	//  1. a node exists in Memory.successors if and only if it exists in the memory.
+	//  2. A node's entry in Memory.successors is always consistent with the actual
```
qweeah, line 47: nit
```suggestion
	//  2. a node's entry in Memory.successors is always consistent with the actual
```

wangxiaoxuan273: changed accordingly

content/oci/oci.go
```diff
@@ -398,6 +402,42 @@ func (s *Store) writeIndexFile() error {
...
+			if !exists {
+				// remove the blob from storage if it does not exist in Store
+				err = os.Remove(path)
+				if err != nil {
+					if errors.Is(err, fs.ErrNotExist) {
+						return fmt.Errorf("%s: %w", blobDigest, errdef.ErrNotFound)
+					}
+					return err
+				}
+			} else if !s.graph.HasPredecessors(blobDigest) {
```
Wwwsylvia, line 430: Looks like `HasPredecessors` is not necessary?
We can get a descriptor from the memory graph via the digest, and then we can call `Predecessors` and check the size of the result array.

Wwwsylvia (changes requested):

content/oci/oci.go
```diff
@@ -398,6 +402,42 @@ func (s *Store) writeIndexFile() error {
...
 }
 
+// GC removes garbage from Store. The garbage to be cleaned are:
+//   - garbage blobs in the storage whose metadata is not stored in Store.
+//   - unreferenced (dangling) blobs in Store which have no predecessors
+func (s *Store) GC(ctx context.Context) error {
+	s.sync.RLock()
+	defer s.sync.RUnlock()
+
+	rootpath := filepath.Join(s.root, "blobs", "sha256")
```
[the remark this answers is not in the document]
wangxiaoxuan273: changed to `rootpath := filepath.Join(s.root, "blobs")`

content/oci/oci.go
```diff
@@ -398,6 +402,42 @@ func (s *Store) writeIndexFile() error {
...
+	return filepath.Walk(rootpath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		// skip the working directory
+		if !info.IsDir() {
+			blobDigest := digest.NewDigestFromEncoded("sha256", info.Name())
```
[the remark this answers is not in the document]
wangxiaoxuan273: Used `digest.Parse` instead.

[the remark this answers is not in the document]
wangxiaoxuan273: Used `digest.Parse`.

content/oci/oci.go
```diff
@@ -398,6 +402,42 @@ func (s *Store) writeIndexFile() error {
...
+					if errors.Is(err, fs.ErrNotExist) {
+						return fmt.Errorf("%s: %w", blobDigest, errdef.ErrNotFound)
+					}
+					return err
+				}
+			} else if !s.graph.HasPredecessors(blobDigest) {
+				if _, err := s.tagResolver.Resolve(ctx, string(blobDigest)); err != nil {
+					// remove the blob and its metadata from the Store
+					s.deleteNode(ctx, desc)
+				}
```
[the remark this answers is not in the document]
wangxiaoxuan273: Refined the GC logic. Now dangling layer clean happens before the memory scan.

content/oci/oci.go
```diff
@@ -398,6 +402,42 @@ func (s *Store) writeIndexFile() error {
...
+	rootpath := filepath.Join(s.root, "blobs", "sha256")
+	return filepath.Walk(rootpath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		// skip the working directory
+		if !info.IsDir() {
+			blobDigest := digest.NewDigestFromEncoded("sha256", info.Name())
+			exists, desc := s.graph.ExistsInStore(blobDigest)
+			if !exists {
```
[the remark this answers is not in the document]
wangxiaoxuan273: Refined the GC logic. Now dangling layer clean happens before the memory scan, and readability has improved.

internal/graph/memory.go
```diff
@@ -133,22 +155,35 @@ func (m *Memory) index(ctx context.Context, fetcher content.Fetcher, node ocispe
...
 			m.predecessors[successorKey] = predecessorSet
 		}
 		predecessorSet.Add(nodeKey)
 	}
 	return successors, nil
 }
+
+// ExistsInStore returns if a blob denoted by its digest exists in the memory.
+// The descriptor associated with the digest is also returned.
+func (m *Memory) ExistsInStore(dgst digest.Digest) (bool, ocispec.Descriptor) {
```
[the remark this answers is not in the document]
wangxiaoxuan273: changed this function to a plain `Exists`, after refining the GC logic.

content/oci/oci.go
wangxiaoxuan273: Removed this function after refining the GC logic.

content/oci/oci.go
```diff
@@ -398,6 +402,42 @@ func (s *Store) writeIndexFile() error {
...
+				// remove the blob from storage if it does not exist in Store
+				err = os.Remove(path)
```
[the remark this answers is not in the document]
wangxiaoxuan273: I like this idea, since `oci.Storage.Delete` would be just a reuse of it.

content/oci/oci.go
```diff
@@ -398,6 +402,49 @@ func (s *Store) writeIndexFile() error {
...
+	danglings := s.graph.GetDanglingLayers()
+	for _, desc := range danglings {
+		// do not remove existing manifests in the index
+		if _, err := s.tagResolver.Resolve(ctx, string(desc.Digest)); err == errdef.ErrNotFound {
+			// remove the blob and its metadata from the Store
+			if err := s.deleteNode(ctx, desc); err != nil {
+				return err
+			}
+		}
+	}
```
shizhMSFT, line 422: It seems it only covers the root nodes of the dangling graph.

Think about this senario:
```mermaid
graph TD
  A --> B
  A --> C
  C --> D
  D --> E
  D --> F
```
After deleting `C`, it becomes
```mermaid
graph TD
  A --> B

  D --> E
  D --> F
```
In this case, `D` is tracked in `danglings` and will be removed but `E` and `F` are still there and not removed.

content/oci/oci.go
```diff
@@ -398,6 +402,49 @@ func (s *Store) writeIndexFile() error {
...
+				return err
+			}
+			if exists := s.graph.Exists(blobDigest); !exists {
+				// remove the blob from storage if it does not exist in Store
+				if err := s.storage.deleteByDigest(ctx, blobDigest); err != nil {
+					return err
+				}
+			}
+		}
+		return nil
```
shizhMSFT, line 444: nit: code style: consider
```go
if info.IsDir() {
    return nil
}

// do other stuffs
return nil
```

content/oci/oci.go
```diff
@@ -398,6 +402,49 @@ func (s *Store) writeIndexFile() error {
...
+	return filepath.Walk(rootpath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		// skip the directories
+		if !info.IsDir() {
+			alg := filepath.Base(filepath.Dir(path))
+			blobDigest, err := digest.Parse(fmt.Sprintf("%s:%s", alg, info.Name()))
+			if err != nil {
+				return err
```
shizhMSFT, line 435: We probably need to return a more meaningful error here. For instance, there are unrecognized files / folders.

content/oci/oci.go
```diff
@@ -398,6 +402,49 @@ func (s *Store) writeIndexFile() error {
...
+			}
+			if exists := s.graph.Exists(blobDigest); !exists {
+				// remove the blob from storage if it does not exist in Store
+				if err := s.storage.deleteByDigest(ctx, blobDigest); err != nil {
+					return err
+				}
```
shizhMSFT, line 441: Since we are operating in a low level, could we call `os.Remove` directly to delete the file?

Wwwsylvia: I think it makes sense to delete a blob through the underlying storage. Maybe we can implement a `Storage.listBlobs`?

content/oci/oci.go
Wwwsylvia: I just realized `deleteByDigest` also validates the digest, so the digest gets validated twice. I think we can save one.

internal/graph/memory.go
```diff
@@ -133,22 +155,39 @@ func (m *Memory) index(ctx context.Context, fetcher content.Fetcher, node ocispe
...
 }
+
+// Exists returns if a blob denoted by its digest exists in the memory.
+func (m *Memory) Exists(dgst digest.Digest) bool {
+	_, exists := m.nodes[dgst]
+	return exists
+}
+
+// GetDanglingLayers returns the dangling (unreferenced) layer nodes in the memory.
+func (m *Memory) GetDanglingLayers() []ocispec.Descriptor {
```
Wwwsylvia, line 185: Wondering if this should be called `GetUnreferencedNodes`? The returned nodes can be either a manifest or a layer blob, right?

content/oci/oci.go
```diff
@@ -398,6 +402,49 @@ func (s *Store) writeIndexFile() error {
...
+	rootpath := filepath.Join(s.root, "blobs")
+	return filepath.Walk(rootpath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		// skip the directories
+		if !info.IsDir() {
+			alg := filepath.Base(filepath.Dir(path))
+			blobDigest, err := digest.Parse(fmt.Sprintf("%s:%s", alg, info.Name()))
+			if err != nil {
```
Wwwsylvia, line 434: There can potentially be invalid content (e.g. path that is not a encoded digest, path is a dir, etc.)  in the directory. I think we should just silently skip them. That is, if we fail to parse the digest, we should just ignore the error.

content/oci/oci.go
Wwwsylvia: We should scan every `alg` directory and every `encoded` in it, rather than walking into every nested sub-dir under "blobs" directory. For example, given the below structure:

```
blobs
  |- sha256
      |- <encoded_1>
      |- random_string
 |- sha512
     |- <encoded_2>
     |- random_dir_1
 |- random_dir_2
```

We should not walk into `random_dir_1` and `random_dir_2`.

content/oci/oci.go
Wwwsylvia: Or maybe we can do `os.Remove` in this PR first and leave the implementation and the use of `Storage.walkBlobs` (or `ReadonlyStorage.Range`) in a separate PR? @shizhMSFT What do you think?

shizhMSFT: SGTM

content/oci/oci.go
```diff
@@ -398,6 +402,49 @@ func (s *Store) writeIndexFile() error {
 	return os.WriteFile(s.indexPath, indexJSON, 0666)
 }
 
+// GC removes garbage from Store. The garbage to be cleaned are:
+//   - unreferenced (dangling) blobs in Store which have no predecessors
+//   - garbage blobs in the storage whose metadata is not stored in Store
+func (s *Store) GC(ctx context.Context) error {
```
shizhMSFT, line 408: Since `GC` is a time-consuming operation, could you add some checkpoint for context cancelling? Also, we need to make sure that `GC` is an idempotent method.

content/oci/oci.go
wangxiaoxuan273: Refined the delete function to delete a dangling tree recursively.

content/oci/oci.go
```diff
@@ -454,6 +455,50 @@ func (s *Store) writeIndexFile() error {
...
+	// clean up dangling layers in Store
+	danglings := s.graph.GetDanglingLayers()
+	for _, desc := range danglings {
+		// do not remove existing manifests in the index
+		if _, err := s.tagResolver.Resolve(ctx, string(desc.Digest)); err == errdef.ErrNotFound {
+			// remove the blob and its metadata from the Store
+			// TODO: s.delete returns 2 variables, a dangling tree is not deleted properly
+			if _, err := s.delete(ctx, desc); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
wangxiaoxuan273: Now I have implemented the recursive delete to remove the entire dangling tree. I put TODO there because it was still in progress.

content/oci/oci.go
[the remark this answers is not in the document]
wangxiaoxuan273: After offline discussion, we decided to discard the idea of `deleteByDigest`.

content/oci/oci.go
wangxiaoxuan273: > nit: code style: consider
> 
> ```go
> if info.IsDir() {
>     return nil
> }
> 
> // do other stuffs
> return nil
> ```

changed accordingly.

wangxiaoxuan273: > We should scan every `alg` directory and every `encoded` in it, rather than walking into every nested sub-dir under "blobs" directory. For example, given the below structure:
> 
> ```
> blobs/
>   |- sha256/
>      |- <encoded_1>
>      |- random_string
>   |- sha512/
>      |- <encoded_2>
>      |- random_dir_1/
>   |- random_dir_2/
> ```
> 
> We should not walk into `random_dir_1/` and `random_dir_2/`.

I think this can be achieved by the below? If the parse fails the invalid content will be silently skipped.

```
			blobDigest, err := digest.Parse(fmt.Sprintf("%s:%s", alg, info.Name()))
			if err != nil {
				return err
			}
```

content/oci/oci.go
wangxiaoxuan273: > I just realized `deleteByDigest` also validates the digest, so the digest gets validated twice. I think we can save one.

Removed `deleteByDigest`.

wangxiaoxuan273: > We probably need to return a more meaningful error here. For instance, there are unrecognized files / folders.

Per offline discussion, we decided to silently skip unrecognized contents.

content/oci/oci.go
wangxiaoxuan273: Put `os.Remove` back.

internal/graph/memory.go
wangxiaoxuan273: Renamed the function to `GetUnreferencedRootNodes`.

content/oci/oci.go
wangxiaoxuan273: Replaced `Walk` with a loop. This comment is no longer relevant.

content/oci/oci.go
wangxiaoxuan273: Changed the way GC works to address this scenario. Now we traverse `index.json` directly.

content/oci/oci.go
wangxiaoxuan273: Silently skip irrelevant stuff for now.

internal/graph/memory.go
```diff
@@ -183,8 +183,8 @@ func (m *Memory) Exists(dgst digest.Digest) bool {
 	return exists
 }
 
-// GetDanglingLayers returns the dangling (unreferenced) layer nodes in the memory.
-func (m *Memory) GetDanglingLayers() []ocispec.Descriptor {
+// GetUnreferencedRootNodes returns the dangling (unreferenced) layer nodes in the memory.
```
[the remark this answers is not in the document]
wangxiaoxuan273: Thanks for the comment. Now I have reverted all changes in `memory.go` after using a better approach (traverse index.json) to scan for garbage.

content/oci/oci.go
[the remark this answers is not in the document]
wangxiaoxuan273: Today I have replaced this entire block of code with a better "scan index.json" approach.

content/oci/oci.go
wangxiaoxuan273: Added context cancelling checkpoints in the loops.

content/oci/oci.go
```diff
@@ -454,6 +455,96 @@ func (s *Store) writeIndexFile() error {
...
+	for len(queue) > 0 {
+		if err := isContextDone(ctx); err != nil {
+			return nil, err
+		}
+		head := queue[0]
+		queue = queue[1:]
+		if visited.Contains(head.Digest) {
+			continue
+		}
+		results.Add(head.Digest)
```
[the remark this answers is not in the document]
wangxiaoxuan273: Thanks @eiffel-fl , that's a good suggestion!

content/oci/oci.go
```diff
@@ -476,3 +576,8 @@ func validateReference(ref string) error {
 	// TODO: may enforce more strict validation if needed.
 	return nil
 }
+
+// isValidAlgorithm checks is a string is a supported hash algorithm
+func isValidAlgorithm(alg string) bool {
+	return alg == string(digest.SHA256) || alg == string(digest.SHA512) || alg == string(digest.SHA384)
```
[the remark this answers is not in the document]
wangxiaoxuan273: replaced with `switch`

content/oci/oci.go
```diff
@@ -476,3 +576,8 @@ func validateReference(ref string) error {
 	// TODO: may enforce more strict validation if needed.
 	return nil
 }
+
+// isValidAlgorithm checks is a string is a supported hash algorithm
+func isValidAlgorithm(alg string) bool {
```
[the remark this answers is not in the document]
wangxiaoxuan273: renamed the function

content/oci/oci.go
```diff
@@ -454,6 +455,94 @@ func (s *Store) writeIndexFile() error {
...
+		alg := algDir.Name()
+		// skip unsupported directories
+		if !isValidAlgorithm(alg) {
+			continue
+		}
+		algPath := path.Join(rootpath, alg)
+		dgstDirs, err := os.ReadDir(algPath)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
wangxiaoxuan273: ignored the error.

content/oci/oci.go
```diff
@@ -454,6 +455,94 @@ func (s *Store) writeIndexFile() error {
...
+//   - garbage blobs in the storage whose metadata is not stored in Store
+func (s *Store) GC(ctx context.Context) error {
+	s.sync.Lock()
+	defer s.sync.Unlock()
+
+	// traverse index.json to find all reachable nodes
+	reachableNodes, err := s.traverseIndex(ctx)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
wangxiaoxuan273: This function doesn't create or change index, it only returns a `Set` for quick check of existence.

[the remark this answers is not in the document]
wangxiaoxuan273: Good point. Removed this function and replaced it with reloading the index (creating a new instance).

wangxiaoxuan273:
Current behavior regarding referrers: when a referrer manifest is not in `index.json`, it will be cleaned by GC. Change may be made in the future when OCI layout has better support for referrers.

content/oci/oci.go
```diff
@@ -454,6 +455,76 @@ func (s *Store) writeIndexFile() error {
...
+	s.index = newStore.index
+	s.storage = newStore.storage
+	s.tagResolver = newStore.tagResolver
+	s.graph = newStore.graph
+	return nil
+}
+
+// GC removes garbage from Store. The garbage to be cleaned are:
```
shizhMSFT, line 471: Need to mention that unsaved index will be lost.

content/oci/oci.go
```diff
@@ -467,6 +538,17 @@ func (s *unsafeStore) Predecessors(ctx context.Context, node ocispec.Descriptor)
 	return s.graph.Predecessors(ctx, node)
 }
 
+// isContextDone returns an error if the context is done.
+// Reference: https://pkg.go.dev/context#Context
+func isContextDone(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
```
shizhMSFT, line 546: This path can be tested by having a context cancelled immediately.

Basically, you need unit test cases for `isContextDone()`.

content/oci/oci.go
```diff
@@ -454,6 +455,76 @@ func (s *Store) writeIndexFile() error {
...
+		if !algDir.IsDir() {
+			continue
+		}
+		alg := algDir.Name()
+		// skip unsupported directories
+		if !isKnownAlgorithm(alg) {
+			continue
+		}
+		algPath := path.Join(rootpath, alg)
+		dgstDirs, err := os.ReadDir(algPath)
```
shizhMSFT, line 501: nit: rename `dgstDirs` to `digestEntries` for more meaningful names.

wangxiaoxuan273: renamed.

content/oci/oci.go
wangxiaoxuan273: Added documentation to mention this.

content/oci/oci.go
wangxiaoxuan273: Added unit tests.

shizhMSFT (approved):
LGTM

Wwwsylvia (commented):
LGTM with a few nit suggestions. Please rebase and resolve comments to unblock merge.

Wwwsylvia (approved):
LGTM

PR #659, opened 2023-12-15
feat: Referrers support for content.ReadOnlyGraphStorage

### LINKED ISSUE #392: Referrers API support for OCI Layout
The OCI layout can be viewed as a simplified local repository, which should also support the equivalence of [Referrers API](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc1/spec.md#listing-referrers).

Proposal: Let [content/oci.Store](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0-rc.6/content/oci#Store) and [content/oci.ReadOnlyStore](https://pkg.go.dev/oras.land/oras-go/v2@v2.0.0-rc.6/content/oci#ReadOnlyStore) implement [registry.ReferrerFinder](https://pkg.go.dev/oras.land/oras-go/v2/registry#ReferrerFinder).

wangxiaoxuan273:
This PR provides `Referrers` support for the `content.ReadOnlyGraphStorage` interface, which includes `oci.Store`.

Resolves #392


content/oci/oci.go
```diff
@@ -375,6 +376,71 @@ func (s *Store) writeIndexFile() error {
...
+//
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#listing-referrers
+func (s *Store) Referrers(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error {
+	var results []ocispec.Descriptor
+	predecessors, err := s.Predecessors(ctx, desc)
+	if err != nil {
+		return err
+	}
+	for _, node := range predecessors {
+		var nodeArtifactType string
```
qweeah, line 393: Annotations are missing, is it on-purpose?

content/oci/oci.go
```diff
@@ -375,6 +376,71 @@ func (s *Store) writeIndexFile() error {
...
+				return err
+			}
+			var index ocispec.Index
+			if err := json.Unmarshal(fetched, &index); err != nil {
+				return err
+			}
+			if index.Subject == nil || !content.Equal(*index.Subject, desc) {
+				continue
+			}
+			nodeArtifactType = index.MediaType
```
qweeah, line 420: Why it's not below?
```suggestion
			nodeArtifactType = index.ArtifactType
```

content/oci/oci.go
```diff
@@ -375,6 +376,71 @@ func (s *Store) writeIndexFile() error {
...
+				return err
+			}
+			var manifest ocispec.Manifest
+			if err := json.Unmarshal(fetched, &manifest); err != nil {
+				return err
+			}
+			if manifest.Subject == nil || !content.Equal(*manifest.Subject, desc) {
+				continue
+			}
+			nodeArtifactType = manifest.MediaType
```
qweeah, line 407: why it's not
```suggestion
			nodeArtifactType = manifest.ArtifactType
```

qweeah: Also what if the artifact type is [denoted by the config descriptor's media type](https://oras.land/docs/concepts/artifact#prior-art)?

content/oci/oci.go
```diff
@@ -375,6 +376,71 @@ func (s *Store) writeIndexFile() error {
...
 
+// Referrers lists the descriptors of image or artifact manifests directly
+// referencing the given manifest descriptor.
+//
+// fn is called on the referrer results. If artifactType is not empty, only
+// referrers of the same artifact type are fed to fn.
+//
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#listing-referrers
+func (s *Store) Referrers(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error {
+	var results []ocispec.Descriptor
```
qweeah, line 387: Can we fail fast like returning directly if the passed-in `desc` is not a manifest at all?

content/oci/oci.go
wangxiaoxuan273: added back annotations.

content/oci/oci.go
wangxiaoxuan273: fixed.

content/oci/oci.go
wangxiaoxuan273: > Also what if the artifact type is [denoted by the config descriptor's media type](https://oras.land/docs/concepts/artifact#prior-art)?

changed L406 to `node.ArtifactType = manifest.Config.MediaType`

content/oci/oci.go
wangxiaoxuan273: Do we need to fail and return an error, or just return `nil`?

content/oci/oci.go
wangxiaoxuan273: Changed L407 to 

```
			if manifest.ArtifactType != "" {
				node.ArtifactType = manifest.ArtifactType
			} else {
				node.ArtifactType = manifest.Config.MediaType
			}
			node.Annotations = manifest.Annotations
```

content/oci/oci_test.go
shizhMSFT, line 1: Can we add a test that make sure the OCI store is indeed a [ReferrerLister](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.1/registry#ReferrerLister)?

content/oci/oci.go
```diff
@@ -375,6 +376,80 @@ func (s *Store) writeIndexFile() error {
...
 }
 
+// Referrers lists the descriptors of image or artifact manifests directly
+// referencing the given manifest descriptor.
+//
+// fn is called on the referrer results. If artifactType is not empty, only
+// referrers of the same artifact type are fed to fn.
+//
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#listing-referrers
+func (s *Store) Referrers(ctx context.Context, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error {
```
shizhMSFT, line 386: `Referrers` is a read-only operation. Should it also be applied to `ReadOnlyStorage`?

shizhMSFT, line 386: This looks like a generic implementation that can work on any `ReadOnlyStorage` which support Predecessors. Can we have a generic implementation like [Tags](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.1/registry#Tags)?

content/oci/oci.go
```diff
@@ -375,6 +376,80 @@ func (s *Store) writeIndexFile() error {
...
+	if !descriptor.IsManifest(desc) {
+		return fmt.Errorf("the descriptor %v is not a manifest", desc)
```
Wwwsylvia, line 388: nit: Should we wrap `errdef.ErrUnsupported`? 🤔

content/oci/oci.go
```diff
@@ -375,6 +376,80 @@ func (s *Store) writeIndexFile() error {
...
+	}
+	var results []ocispec.Descriptor
+	predecessors, err := s.Predecessors(ctx, desc)
+	if err != nil {
+		return err
+	}
+	for _, node := range predecessors {
+		switch node.MediaType {
```
nathan-joslin, line 396: In my implementation, I included a check for the ArtifactType in the descriptor prior to this switch statement as an optimization. Thoughts?

My reasoning was from reading "listing-referrers" from [1.1.0-rc3](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#listing-referrers), "Upon success, the response MUST be a JSON body with an image index containing a list of descriptors. Each descriptor is of an image manifest in the same <name> namespace with a subject field that specifies the value of <digest>. **The descriptors MUST include an artifactType field that is set to the value of the artifactType in the image manifest, if present.** If the artifactType is empty or missing in the image manifest, the value of artifactType MUST be set to the config descriptor mediaType value."

content/oci/oci_test.go
wangxiaoxuan273: Added the interface check test.

content/oci/oci.go
wangxiaoxuan273: Done a generic implementation, with an input parameter of `ReadOnlyGraphStorage`.

content/oci/oci.go
wangxiaoxuan273: changed accordingly.

content/oci/oci.go
wangxiaoxuan273: In the current implementation, we check the actual manifest content (the json) for the value of `artifactType`. This gives the most accurate value, as the descriptor value of `artifactType` may be missing. Given that the actual manifest is checked, I don't think the descriptor value check is necessary.

nathan-joslin: Correct it is not necessary, but it is aimed to be an optimization that I believe is worthwhile. 

The quote I included from the distribution spec outlines exactly what you have said, "the `artifactType` *may* be missing". As such I believe it is best to only fetch the manifest if it is necessary to do so, i.e. in the case that the `artifactType` *is not missing* it is unnecessary to fetch the manifest as the `artifactType` can be pulled straight from the descriptor. As you already have, in the case that the `artifactType` *is missing* I agree the manifest must be fetched and the `artifactType` be pulled from there.

I believe the only case where it is *necessary* to fetch the manifest each and every time is if the distribution spec allows for the return of a mismatched, non-empty, `artifactType` in the descriptor; which I do not believe this is the case as this would be a major flaw.

Taking another look at my proposed addition, I believe it should be closer to the `artifactType` check at the bottom of the function. In the case that the `artifactType` we use as a filter is empty then no manifests will be fetched and all will be included in the results.

```go
		if artifactType == "" || node.ArtifactType == artifactType {
			results = append(results, node)
			continue
		}
```

registry/repository.go
```diff
@@ -134,3 +139,81 @@ func Tags(ctx context.Context, repo TagLister) ([]string, error) {
...
 }
+
+// Referrers lists the descriptors of image or artifact manifests directly
+// referencing the given manifest descriptor.
+//
+// fn is called on the referrer results. If artifactType is not empty, only
+// referrers of the same artifact type are fed to fn.
+//
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#listing-referrers
+func Referrers(ctx context.Context, store content.ReadOnlyGraphStorage, desc ocispec.Descriptor, artifactType string, fn func(referrers []ocispec.Descriptor) error) error {
```
shizhMSFT, line 150: Consider
```suggestion
func Referrers(ctx context.Context, store content.ReadOnlyGraphStorage, desc ocispec.Descriptor, artifactType string) ([]ocispec.Descriptor) error) {
```

registry/repository_test.go
shizhMSFT, line 1: Generally, you need at least the following test cases
1. Test against a mock instance which is a `ReadOnlyGraphStorage` but not a `ReferrersLister`
2. Test against a mock instance which is a `ReadOnlyGraphStorage` and a `ReferrersLister`

registry/repository_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+func (s *testStorage) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+	return s.store.Exists(ctx, target)
+}
+
+func (s *testStorage) Predecessors(ctx context.Context, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+	return s.store.Predecessors(ctx, node)
+}
+
+func TestReferrers(t *testing.T) {
```
Wwwsylvia, line 55: We should also cover test cases where:
1. The input descripor is not a manifest
2. The storage is a `ReferrersLister`
3. The predecessor is of other media type

See https://app.codecov.io/gh/oras-project/oras-go/pull/659/blob/registry/repository.go for the current coverage.

content/oci/oci.go
wangxiaoxuan273: We have to fetch the manifest each and every time, as we are using `Predecessors()` to implement `Referrers()`. `Predecessors()` returns not only the referrers of the node but also other predecessors. We have to check the content of each manifest, and mark it as a referrer only when the `Subject` field matches. So even if the `ArtifactType` field of the descriptor is present and correct, we have to check the manifest content for its `Subject`.

registry/repository.go
wangxiaoxuan273: changed accordingly.

registry/repository_test.go
```diff
@@ -0,0 +1,238 @@
...
+	return s.store.Fetch(ctx, target)
+}
+
+func (s *testStorage) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+	return s.store.Exists(ctx, target)
+}
+
+func (s *testStorage) Predecessors(ctx context.Context, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+	return s.store.Predecessors(ctx, node)
+}
```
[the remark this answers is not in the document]
wangxiaoxuan273: It's better to test with a more general instance of `ReadOnlyGraphStorage` that has only the required interface.

registry/repository_test.go
wangxiaoxuan273: Added these two test cases.

registry/repository_test.go
wangxiaoxuan273: Added these three test cases.

registry/repository.go
```diff
@@ -134,3 +139,89 @@ func Tags(ctx context.Context, repo TagLister) ([]string, error) {
 	}
 	return res, nil
 }
+
+// Referrers lists the descriptors of image or artifact manifests directly
+// referencing the given manifest descriptor.
+//
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#listing-referrers
+
```
shizhMSFT, line 147: remove empty line

shizhMSFT (changes requested):

shizhMSFT (commented):
@wangxiaoxuan273 The code LGTM. Can you increase the unit test code coverage at least to the current level of 75%?

Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

Issue #641, opened 2023-11-07
Add issue template for oras-go


FeynmanZhou:
Hi maintainers,

I am thinking it would be better to replicate the issue template from [ORAS repo](https://github.com/oras-project/oras/tree/main/.github/ISSUE_TEMPLATE) to oras-go since it helps issue creators to shape the issue content to a standard structure and added related labels automatically. It would be convenient for issue creators to provide enough context by following a content structure so that maintainer are easy to triage issues.

dextrot:
Hi @FeynmanZhou 
Can i submit a PR for this issue?

PR #661, opened 2023-12-21
add support for optional progress updates channel

deitch:
If `CopyOptions.UpdateChannel` is non-nil, then with each block copied during a `Copy*()`, it sends an update down the channel. The update includes the amount copied in this update, and the descriptor of what is being copied. This descriptor lets the consumer determine what blob it applies to, and what the expected total size is (which can be used, e.g. to calculate percentages or create a progress bar).

It does not include relative sizes of the whole root desc or tag in the `Copy()`, because we have no way of knowing that; you would have to walk the entire tree first. Granted, you can get it from manifests first, and only then the more expensive other elements, but that would require a complete restructure. This is more than good enough for know.

Discussed in Slack with @Wwwsylvia


deitch:
Sylvia did raise the following questions. Adding here for follow-on discussion:

1. It is not clear to me that how does the caller read the channel. Is the channel generated internally or is it passed in by the caller? Is it per-blob?
1. How do we report progress for the skipped (existing) blobs? What if the caller (e.g. ORAS CLI) wants to show 100% for the skipped blobs (as shown in the screenshot)?
1. What if the caller wants to track the progress of [PushBytes](https://pkg.go.dev/oras.land/oras-go/v2#PushBytes)? Is there a way to reuse this mechanism?

Basically, the idea is that we want to make sure the APIs of oras-go are generic and extensible.

deitch:
Some responses:

> It is not clear to me that how does the caller read the channel. Is the channel generated internally or is it passed in by the caller?

Passed in by the caller, `CopyOptions.UpdateChannel`. If it is nil, we do nothing different; if it is non-nil, we send updates to it with each copy.

> Is it per-blob?

No, one channel, but each update includes the descriptor. See the extended comment on `UpdateChannel`.

> How do we report progress for the skipped (existing) blobs?

I hadn’t thought about that. I think I would send a single update of 100% on existing blobs. But that can be subject for discussion. I think I would do this first, then do that in a follow-on. One improvement at a time.

> What if the caller wants to track the progress of [PushBytes](https://pkg.go.dev/oras.land/oras-go/v2#PushBytes)? Is there a way to reuse this mechanism?

I don’t see why not. All this is, is a mechanism for sending updates in a channel. The issue here is that there are no “opts” on which to add it

deitch:
Rebased

copy.go
```diff
@@ -112,6 +112,23 @@ type CopyGraphOptions struct {
...
+	// passed with each update. The total size of downloads of all blobs and
+	// manifests is not provided, as it is not known. You can calculate the
+	// percentage downloaded for a particular blob in an individual update
+	// based on the total size of that blob, which is provided in the
+	// descriptor, and the number of bytes copied, which is provided in the
+	// update.
+	// Updates are sent each time a block is copied. The number of bytes copied
+	// depends upon io.Copy, which, by default, is 32KB. As of now, this cannot
+	// be changed. We may provided that capability in a future update.
+	UpdateChannel chan<- CopyUpdate
```
shizhMSFT, line 131: Channels are blocking. If someone sets `opts.UpdateChannel = make(chan<- CopyUpdate)` and never consumes the updates, the `Copy()` will be blocked forever.

copy.go
```diff
@@ -112,6 +112,23 @@ type CopyGraphOptions struct {
...
+	// or manifest, the expected total size, and the descriptor of the blob or
+	// manifest. It is up to the consumer of the channel to differentiate
+	// between updates among different blobs and manifests; no mechanism is
+	// provided for distinguishing between them, other than the descriptor
+	// passed with each update. The total size of downloads of all blobs and
+	// manifests is not provided, as it is not known. You can calculate the
+	// percentage downloaded for a particular blob in an individual update
+	// based on the total size of that blob, which is provided in the
+	// descriptor, and the number of bytes copied, which is provided in the
+	// update.
```
shizhMSFT, line 127: Presenting a progress bar of an operation isn't an easy job. Eventually, we will end up with an [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) pattern if we continue iterating.

Therefore, we need to define data models first for oras operations like `Copy`, `PushBytes`, etc.

copy.go
```diff
@@ -112,6 +112,23 @@ type CopyGraphOptions struct {
...
+	// Updates are sent each time a block is copied. The number of bytes copied
+	// depends upon io.Copy, which, by default, is 32KB. As of now, this cannot
+	// be changed. We may provided that capability in a future update.
```
shizhMSFT, line 130: Not really. The copy operation does not necessarily depends on `io.Copy`. It is purely based on the implementation of the destination Target. For instance, copying from any target to memory / oci targets does not call `io.Copy`.

copy.go
```diff
@@ -266,11 +283,17 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
 	rc, err := src.Fetch(ctx, desc)
 	if err != nil {
 		return err
 	}
+	if ch != nil {
+		rc = &progressReader{
+			c: ch,
+			r: rc,
+		}
+	}
```
shizhMSFT, line 296: Wrapping a `ReadCloser` may have perf penalties as some types may implement `io.WriteTo` which can trigger an optimization by the built-in `io.Copy()` method.

copy.go
shizhMSFT: In fact, `oras` CLI has a data model so as the `buildkit`.

progress_reader.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+type progressReader struct {
+	desc ocispec.Descriptor
+	r    io.ReadCloser
+	c    chan<- CopyUpdate
+}
+
+func (p *progressReader) Close() error {
+	close(p.c)
```
shizhMSFT, line 36: Closing the channel is dangerous as the channel is shared across all nodes. It panics if other nodes get updates.

In other words, the caller has the responsibility to close the channel.

shizhMSFT (commented):
A complete and sound progress updating framework is non-trivial for both oras-go v1 and v2. Thanks for the initial attempt and let's see if we can iterate it to a better design.

copy.go
shizhMSFT: /cc @qweeah

deitch: > Presenting a progress bar of an operation isn't an easy job

True, but I am not concerned with that here. `oras` CLI might, but the `Copy()` and such library calls should not care. They only should worry about some library-centric way of publishing updates. A CLI (like `oras`) or other consumer can do with them what they want.

> we need to define data models

The data model is simple: stream of defined `struct` (or `interface`, if you prefer), each of which contains the amount of bytes transferred and descriptor to which it applies. Anything higher level would be outside the scope of this type of option, I think.

copy.go
deitch: Fair enough. We pass the target an `io.ReadCloser`, so something is calling `Read()` multiple times. We don't really care what it is, as long as it calls `Read()` (which it does). I can update the comment here.

deitch: I updated this comment; take a look.

copy.go
deitch: Are you saying that it is performance impact by wrapping the `Read()` call? Or that by not having the wrapper also  implement `WriterTo`, if the original `ReadCloser` does, then we lose the ability to `WriteTo`? I think the second of these.

What do you suggest? That we check for `WriterTo`, like `io.Copy()` does, and if it does, implement `WriteTo()`.

How would we then capture the updates?

progress_reader.go
deitch: Yeah, very good point. I will change that and add the comment.

deitch: Updated

deitch:
Going to rebase to get it back in line with main; any thoughts on the feedback @shizhMSFT ?

deitch:
Rebased

shizhMSFT:
> Going to rebase to get it back in line with main; any thoughts on the feedback @shizhMSFT ?

@deitch I think this PR is good brainstorming idea and worths an issue or discussion so that we can design it holistically with production-level quality.

deitch:
> PR is good brainstorming idea and worths an issue or discussion so that we can design it holistically with production-level quality

I have no idea what that means. If the overall direction is good, let's fix this and get it in. If not, let's redo it.

shizhMSFT:
It means that we need an issue or discussion for a systematical design as this PR is a feature request proposal. The reason behind it is that `oras-go/v2` is now at a stable state and we should introduce new a feature (i.e. merge to `main`) with the best design we could achieve to avoid future breaking changes as much as we can. (note: breaking changes requires us to introduce `oras-go/v3`).

Since issue and discussion are good place to propose a feature request, would you like to open one with details like the motivation of the feature request (i.e. what would you like to be added? why is this needed for oras?)? We should have an issue template for `oras-go` but it is still in progress (tracked by #641; the `oras` repo has [one](https://github.com/oras-project/oras/issues/new/choose)). With an open issue / discussion, more maintainers like @sajayantony, @TerryHowe, and @sabre1041 can join the discussion and give inputs so that this conversation is not just between you and me or @Wwwsylvia. Of course, you may reference this PR in the new issue / discussion as an implementation candidate (so do others). #650 and #651 by @ktarplee are a good example.

Meanwhile, I would like to continue the discussions on the aspects of API design, concurrency (go-routine safety), performance, and security so that we can iterate your idea and make `oras-go` a better library.

deitch:
Sure. This all started with an extended discussion we had on Slack with @Wwwsylvia . Part of the issue is that this capability existed (in a different way, but still) in v1. The move to v2, with all of its many benefits, lost this capability. This is trying to get it back, so that downstream dependencies that want to switch from v1 to v2 can do so without loss of ability.

BTW, if this PR changes an API, and there is a way to do it without changing it (i.e. backwards-compatible), then by all means. If I recall correctly, that was the main driver behind the variadic options in the main calls in v1. We knew we could add features going forward without changing the API.

sajayantony:
I think this is general goodness and so I'm in support of landing something that is feasible and generally stable. 
Can someone given an code example of how this API might be consumed  with maybe a warning of how it can be abused as well?

deitch:
> Can someone given an code example of how this API might be consumed
> maybe a short hackmd doc or something

I will put something in a comment here. It looks like there is question about whether or not this API is the right one, or even about putting in the changes. Once we have general agreement on adding it and the approach, I am happy to put a formal example and doc anywhere we want.

deitch:
Remove stale

deitch:
@sajayantony how can we bring this back to life?

shizhMSFT:
According to [the discussions](https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1729511967712099?thread_ts=1700489633.277829&cid=CJ1KHJM5Z) on the slack, more topics could be
- What does the view model look like?
- What's the model of an update? How to determine the frequency of the update?
- How to emit the update?
- How do we ensure API flexibility? Not everyone likes channels.
- How do we ensure performance is not degraded?
- ...

@deitch will open a new feature request issue to track properly.

qweeah:
Image Description

It seems the same issue occurred in this PR repeatively. The `stale` label was added and then removed almost immediately. Could this be a CI configuration issue?

TerryHowe:
I thought this was not working out, but reopen if I'm incorrect. Thanks.

PR #665, opened 2024-01-03
feat: add MountFrom and OnMounted to CopyGraphOptions

### LINKED ISSUE #580: Support cross-repository mounting in Copy

ktarplee:
Adds MountFrom and OnMounted to CopyGraphOptions.
Allows for trying to mount from multiple repositories.

Closes #580 

I think this is a better approach than my other PR #632

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -105,6 +105,12 @@ type CopyGraphOptions struct {
 	// OnCopySkipped will be called when the sub-DAG rooted by the current node
 	// is skipped.
 	OnCopySkipped func(ctx context.Context, desc ocispec.Descriptor) error
+	// MountFrom returns the candidate repositories that desc may be mounted from.
+	// The OCI references will be tried in turn.  If mounting fails on all of them,
+	// then it falls back to a copy.
+	MountFrom func(ctx context.Context, desc ocispec.Descriptor) ([]string, error)
+	// OnMounted will be invoked when desc is mounted.
+	OnMounted func(ctx context.Context, desc ocispec.Descriptor) error
 	// FindSuccessors finds the successors of the current node.
 	// fetcher provides cached access to the source storage, and is suitable
 	// for fetching non-leaf nodes like manifests. Since anything fetched from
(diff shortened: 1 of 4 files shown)


copy.go
```diff
@@ -105,6 +105,11 @@ type CopyGraphOptions struct {
 	// OnCopySkipped will be called when the sub-DAG rooted by the current node
 	// is skipped.
 	OnCopySkipped func(ctx context.Context, desc ocispec.Descriptor) error
+	// MountFrom returns the candidate repositories that desc may be mounted from.
+	// The OCI references will be tried in turn.  If mounting fails on all of them, then it falls back to a copy.
```
shizhMSFT, line 109: nit: formating
```suggestion
	// MountFrom returns the candidate repositories that desc may be mounted from.
	// The OCI references will be tried in turn. If mounting fails on all of them,
	// then it falls back to a copy.
```

copy.go
```diff
@@ -259,12 +264,94 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
 	}
 
 	return syncutil.Go(ctx, limiter, fn, root)
 }
 
+// mountOrCopyNode enabled cross repository blob mounting.
+// sourceReference is the repository to use for mounting (the mount point).
+// mounter is the destination for the mount (a well-known implementation of this is *registry.Repository representing the target).
+// OnMounted is called (if provided) when the blob is mounted.
+// The original PreCopy hook is called only on copy, and therefore not when the blob is mounted.
```
shizhMSFT, line 277: nit: it would be better to wrap the documentation at line width of `80` chars.

copy.go
```diff
@@ -259,12 +264,94 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
+	sourceRepositories, err := opts.MountFrom(ctx, desc)
+	if err != nil {
+		// Technically this error is not fatal, we can still attempt to copy the node
+		// But for consistency with the other callbacks we bail out.
+		return err
+	}
+
+	if len(sourceRepositories) == 0 {
+		return copyNode(ctx, src, dst, desc, opts)
+	}
```
shizhMSFT, line 303: nit: could you refactor this code as it has many paths calling `return copyNode(ctx, src, dst, desc, opts)`?

copy.go
```diff
@@ -259,12 +264,94 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
+
+	skipContent := errors.New("skip content")
```
shizhMSFT, line 305: nit: could you make it a global variable so that it is only created once?

copy.go
```diff
@@ -259,12 +264,94 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
+	for i, sourceRepository := range sourceRepositories {
+		// try mounting this source repository
+		var mountFailed bool
+		getContent := func() (io.ReadCloser, error) {
+			// the invocation of getContent indicates that mounting has failed
+			mountFailed = true
+
+			if len(sourceRepositories)-1 == i {
+				// this is the last iteration so we need to actually get the content and do the copy
+
```
shizhMSFT, line 315: nit: remove extra empty line

shizhMSFT (approved):
LGTM with suggestions. I think this design is indeed a better approach than #632.

shizhMSFT:
@ktarplee Meanwhile, could you increase the code coverage so that the check can pass?

copy.go
ktarplee: I think the benefit of the variable's scope being as small as possible overrides the benefit of the optimization.

copy.go
ktarplee: done

copy.go
ktarplee: done

copy.go
ktarplee: done

shizhMSFT (approved):
LGTM

copy.go
```diff
@@ -259,12 +265,85 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
+				if opts.PreCopy != nil {
+					if err := opts.PreCopy(ctx, desc); err != nil {
+						return nil, err
+					}
+				}
+				return src.Fetch(ctx, desc)
+			}
+
+			// We want to return an error that we will test for from mounter.Mount()
+			return nil, skipContent
```
Wwwsylvia, line 318: nit: Would something like this be more readable?

```suggestion
			if i < len(sourceRepositories)-1 {
				// If this is not the last one, skip this source and try next one
				// We want to return an error that we will test for from mounter.Mount()
				return nil, skipContent
			}

			// this is the last iteration so we need to actually get the content and do the copy
			// but first call the PreCopy function
			if opts.PreCopy != nil {
				if err := opts.PreCopy(ctx, desc); err != nil {
					return nil, err
				}
			}
			return src.Fetch(ctx, desc)
```

copy.go
```diff
@@ -259,12 +265,85 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
+		// Technically this error is not fatal, we can still attempt to copy the node
+		// But for consistency with the other callbacks we bail out.
+		return err
+	}
+
+	if len(sourceRepositories) == 0 {
+		return copyNode(ctx, src, dst, desc, opts)
+	}
+
+	skipContent := errors.New("skip content")
```
Wwwsylvia, line 298: nit: `skipContent` sounds like this node is skipped, but this error is used to skip the source? Wondering if it should be called something like `skipSource` or `ignoreSource`, etc.?

copy.go
ktarplee: I like that better as well.

copy.go
ktarplee: Changed to `skipSource`

ktarplee:
@Wwwsylvia @shizhMSFT I pushed my changes.  Sorry about the delay.

Wwwsylvia (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #667, opened 2024-01-03
WIP: fix: support for deprecated image spec release candidates in PackMani…

jlbutler:
This is a WIP PR related to [oras issue 1224](https://github.com/oras-project/oras/pull/1227). See discussion there. 

This PR is to facilitate conversation on the oras issue. If the team thinks this is a good PR, I'll open an issue and tidy things up a bit.

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -66,12 +66,19 @@ const (
 	// PackManifestVersion1_0 represents the OCI Image Manifest defined in
 	// image-spec v1.0.2.
 	// Reference: https://github.com/opencontainers/image-spec/blob/v1.0.2/manifest.md
-	PackManifestVersion1_0 PackManifestVersion = 1
+	PackManifestVersion1_0 PackManifestVersion = iota + 1
 
 	// PackManifestVersion1_1_RC4 represents the OCI Image Manifest defined
 	// in image-spec v1.1.0-rc4.
 	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md
-	PackManifestVersion1_1_RC4 PackManifestVersion = 2
+	PackManifestVersion1_1_RC4
+
+	// PackManifestVersion1_1_RC2 represents the OCI Image Manifest defined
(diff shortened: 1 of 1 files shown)


sajayantony:
@jlbutler could you fix the DCO on this?

pack.go
```diff
@@ -66,12 +66,19 @@ const (
...
 	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md
-	PackManifestVersion1_1_RC4 PackManifestVersion = 2
+	PackManifestVersion1_1_RC4
+
+	// PackManifestVersion1_1_RC2 represents the OCI Image Manifest defined
+	// in image-spec v1.1.0-rc2.
+	// Reference: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/manifest.md
+	// Note: The project may maintain deprecated release candidate versions for expanded
+	//       test coverage during specification development; to be removed when spec is GA.
+	PackManifestVersion1_1_RC2
```
shizhMSFT, line 81: The pack version `PackManifestVersion1_1_RC2` is deprecated and should not be added to `PackManifest`. Instead, users should use the deprecated `Pack()` for those deprecated RC versions.

shizhMSFT: The reason behind it is as simple as we don't encourage new code to produce old RC versions of manifest.

shizhMSFT (changes requested):

pack.go
shizhMSFT: Example implementation can be found at https://github.com/oras-project/oras/pull/1227#discussion_r1441186531

jlbutler: I agree that it's deprecated, and ideally should be pulled back out once 1.1 specs GA. However, the oras client image spec parsing uses these consts. Don't we need to at least add this const for that purpose?

jlbutler:
This is in support of  [oras issue 1224](https://github.com/oras-project/oras/pull/1227) which I just closed. Will close this one as well. Thanks so much for the support and discussion!

PR #670, opened 2024-01-04
doc: add example for digest and parseReference

sajayantony:
Using this PR to discuss the need for importing the hash algorithm. One option is that add this to the examples for digest or parse reference or it possible to have indirect dependencies. I'm leaning towards not having an indirect dependency since this means that if and when a algorithm becomes unacceptable the library has to rev. Instead clients taking the appropriate imports could define the actual hash algorithms that are being used - 
https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1704365828594969

On a side note, how what would be a good pattern to exclude full file examples from license header check? 

Related to - https://github.com/oras-project/oras-go/pull/671

### DIFF
--- a/registry/example_parse_test.go
+++ b/registry/example_parse_test.go
@@ -1,0 +1,32 @@
+package registry_test
+
+import (
+	_ "crypto/sha256"
+	"fmt"
+	"log"
+
+	"oras.land/oras-go/v2/registry"
+)
+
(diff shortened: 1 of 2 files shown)


shizhMSFT:
Closing this PR as it is succeeded by #675 and whole file example can be found on [godoc](https://pkg.go.dev/oras.land/oras-go/v2@v2.3.1-0.20240110064922-c34d275e1ae7/registry#example-ParseReference-Digest).

PR #674, opened 2024-01-09
feat: error type for not finding basic auth creds and export config path

### LINKED ISSUE #676: add exported function to get the path of a DynamicStore's underlying config file
DynamicStore dynamically determines which store to use based on the settings in a config file. Currently there is no way to get the config file path from caller side. 

Suggest adding a function to obtain the path, which will be helpful for ORAS CLI implementation of https://github.com/oras-project/oras/issues/1236.

qweeah:
Two changes proposed in this PR
1) When failed to fetch credentials for basic auth, return a new error
2) Add a new function to DynamicStore

An example usage of the proposed change: oras-project/oras#1235, resolves #676


registry/remote/credentials/store.go
```diff
@@ -193,8 +193,8 @@ func (ds *DynamicStore) getStore(serverAddress string) Store {
 	return fs
 }
 
-// getDockerConfigPath returns the path to the default docker config file.
-func getDockerConfigPath() (string, error) {
+// GetDockerConfigPath returns the path to the default docker config file.
+func GetDockerConfigPath() (string, error) {
```
shizhMSFT, line 197: This method should not be exported as it is not in an internal package.

errdef/errors.go
```diff
@@ -28,4 +28,5 @@ var (
 	ErrSizeExceedsLimit   = errors.New("size exceeds limit")
 	ErrUnsupported        = errors.New("unsupported")
 	ErrUnsupportedVersion = errors.New("unsupported version")
+	ErrBasicCredNotFound  = errors.New("basic credential not found")
```
shizhMSFT, line 31: Should it be put in the `auth` package as it is not a generic error.

registry/remote/credentials/store.go
qweeah: Sorry I don't get it, it's not in an internal package. I need this to default path of docker config. I thought everything under `registry/remote/credentials` are exportable.

registry/remote/credentials/internal/config/config.go
```diff
@@ -155,6 +155,14 @@ func Load(configPath string) (*Config, error) {
 	return cfg, nil
 }
 
+// Path returns the path to the config file.
+func (cfg *Config) Path() (string, error) {
```
Wwwsylvia, line 159: This looks unnecessary to me.

registry/remote/credentials/store.go
```diff
@@ -167,6 +167,11 @@ func (ds *DynamicStore) IsAuthConfigured() bool {
 	return ds.config.IsAuthConfigured()
 }
 
+// GetConfigPath returns the path to the config file.
+func (ds *DynamicStore) GetConfigPath() (string, error) {
+	return ds.config.Path()
```
Wwwsylvia, line 172: I think we don't need to define `config.Path()` and we can just have a `configPath` field in the `DynamicStore` struct and return that here.

registry/remote/credentials/store.go
```diff
@@ -167,6 +167,11 @@ func (ds *DynamicStore) IsAuthConfigured() bool {
 	return ds.config.IsAuthConfigured()
 }
 
+// GetConfigPath returns the path to the config file.
+func (ds *DynamicStore) GetConfigPath() (string, error) {
```
Wwwsylvia, line 171: Wondering if this method should be called `ConfigPath()`? 🤔

registry/remote/credentials/store.go
```diff
@@ -167,6 +169,14 @@ func (ds *DynamicStore) IsAuthConfigured() bool {
 	return ds.config.IsAuthConfigured()
 }
 
+// ConfigPath returns the path to the config file.
+func (ds *DynamicStore) ConfigPath() (string, error) {
+	if ds.config == nil {
+		return "", fmt.Errorf("config is not loaded")
+	}
```
Wwwsylvia, line 176: Is this part necessary? Is there a valid case where `ds.config` is not loaded? If `ds` is nil, every method will panic.

qweeah: Should check both `ds` and `ds.config` since user can
```golang
var ds *DynamicStore
path, err := ds.ConfigPath()
```
or
```
```golang
var ds DynamicStore
path, err := ds.ConfigPath()
```
```

Comment (Wwwsylvia) [reply registry/remote/credentials/store.go]: If one does this, the program will panic.
Methods like  `store.Get` and `store.IsAuthConfigured` will also panic.

Comment (Wwwsylvia) [reply registry/remote/credentials/store.go]: Do we need to tolerate such programming fault?

Comment (qweeah) [reply registry/remote/credentials/store.go]: I am not aware that such error are not handled on oras-go side, since we have existing cases for `store.Get` and `store.IsAuthConfigured`, I am happy to remove it.

**Review by Wwwsylvia — APPROVED** [review]
LGTM.

**Comment by Wwwsylvia** [comment]
I think this change is a `feat` instead of a `chore` as it introduces public types and methods.

File: registry/remote/credentials/store.go [hunk 7]
```diff
@@ -51,6 +51,7 @@ type Store interface {
 // in the config file.
 type DynamicStore struct {
 	config             *config.Config
+	configPath         string
```
shizhMSFT, line 54: `config` has already has the information of the `path`. To avoid duplicated info, `*config.Config` should have a method named
```go
func (cfg *Config) Path() string {
    return cfg.path
}
```
and then
```go
func (ds *DynamicStore) ConfigPath() string {
	return ds.config.Path()
}
```

qweeah: Changed accordingly.

registry/remote/auth/client.go
```diff
@@ -31,6 +31,8 @@ import (
 	"oras.land/oras-go/v2/registry/remote/retry"
 )
 
+var ErrBasicCredentialNotFound = errors.New("basic credential not found")
```
shizhMSFT, line 34: Missing documentation for this error

registry/remote/auth/client.go
```diff
@@ -31,6 +31,10 @@ import (
 	"oras.land/oras-go/v2/registry/remote/retry"
 )
 
+// ErrBasicCredentialNotFound is returned when the basic credential is not
```
Wwwsylvia, line 34: nit: when the credential is not found for basic auth?

Wwwsylvia (approved):
LGTM with a nit suggestion

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

PR #678, opened 2024-01-12
fix: get rid of deprecated types.AuthConfig

ilya-lesikov:
`types.AuthConfig` removed in github.com/docker/docker@v25, can't update to it


shizhMSFT (approved):
LGTM

PR #679, opened 2024-01-16
feat: allow skip unpack on push to file store

### LINKED ISSUE #644: feature: allow for skip unpack on pull
## Problem

Currently, the logic to decide if an archive artifact is unpacked is determiend on push via an annotation. Here is how I set that to a "path" ".":

```content
# contents of annotations.go
{".": {"io.deis.oras.content.unpack": "false"}}

# push, targeting annotations.go
oras push --annotation-file ./annotations.json ${pushto} --plain-http .
```

And the resulting manifest is how I want it.

image

However, the problem is that when it's time to pull, I (as the pulling user) have no control about this unpack. And in fact, I might want different outcomes depending on where I'm doing the extraction and the size of the archive.

## Proposed Solution

While we can default to the preference of the annotation set on push (this makes sense as the pusher knows most about the artifact) ideally I as the user can make this choice at pull time, via a flag:

```bash
oras pull --skip-unpack localhost:5000/hello:v1
```

The above says that I don't want to unpack, regardless of what the annotation says. Since this is globally applied to the layers, I would propose we apply to all layers where it's relevant. The use case of wanting to fine grained control the unpack might be addressed if/when an actual user calls for it.

## Use Case

My use case is the oras-operator - https://github.com/converged-computing/oras-operator. I did an analysis today using something called hpctoolkit that generates huuuge databases and results, and I don't want it to unpack on my local machine. But depending on where I'm pulling it, I might want it to. The point being, I should be able to decide on pull.

From @FeynmanZhou: 

> Thanks for sharing the use case. Your scenario looks interesting and new! Now you are unblocked in customizing annotation, but are looking for a switch to control the unpack behavior in oras pull and corresponding AnnotationUnpack function in oras-go?

And yes, this is spot on. Thank you!

wangxiaoxuan273:
Resolves #644

### DIFF
--- a/content/file/file.go
+++ b/content/file/file.go
@@ -104,6 +104,9 @@ type Store struct {
 	// manifest and config file, while leaving only named layer files.
 	// Default value: false.
 	IgnoreNoName bool
+	// AutoExtract controls if push operations should extract files automatically.
+	// Default value: true.
+	AutoExtract bool
 
 	workingDir   string   // the working directory of the file store
 	closed       int32    // if the store is closed - 0: false, 1: true.
@@ -150,6 +153,7 @@ func NewWithFallbackStorage(workingDir string, fallbackStorage content.Storage) 
 	}
 
 	return &Store{
+		AutoExtract:     true,
 		workingDir:      workingDirAbs,
 		fallbackStorage: fallbackStorage,
 		resolver:        resolver.NewMemory(),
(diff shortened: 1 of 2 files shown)


content/file/file.go
```diff
@@ -104,6 +104,9 @@ type Store struct {
 	// manifest and config file, while leaving only named layer files.
 	// Default value: false.
 	IgnoreNoName bool
+	// AutoExtract controls if push operations should extract files automatically.
+	// Default value: true.
+	AutoExtract bool
```
Wwwsylvia, line 109: I think we should call this `AutoPack` or `SkipUnpack`(with default value: `false`), as the value of `AnnotationUnpack` is `io.deis.oras.content.unpack`.

content/file/file.go
```diff
@@ -482,12 +486,15 @@ func (s *Store) pushDir(name, target string, expected ocispec.Descriptor, conten
...
 	defer bufPool.Put(buf)
-	if err := extractTarGzip(target, name, gzPath, checksum, *buf); err != nil {
-		return fmt.Errorf("failed to extract tar to %s: %w", target, err)
+	if s.AutoExtract {
+		checksum := expected.Annotations[AnnotationDigest]
+		if err := extractTarGzip(target, name, gzPath, checksum, *buf); err != nil {
+			return fmt.Errorf("failed to extract tar to %s: %w", target, err)
+		}
 	}
+
```
Wwwsylvia, line 497: If we don't do auto unpack, we should save the gz file to the target path.

Wwwsylvia (changes requested):

content/file/file.go
wangxiaoxuan273: renamed the variable as `SkipUnpack`

content/file/file.go
wangxiaoxuan273: Changed line 271 to address this.

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

Issue #664, opened 2024-01-02
Manifest should not be added to `index.json`


eiffel-fl:
Hi!

Manifests are added to `index.json`:

```json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:b0eff7f0ec2a27b8d71fe904cf373fa2002c6d669db54c25a9542cf2cb5f46d1",
      "size": 491,
      "annotations": {
        "org.opencontainers.image.ref.name": "docker.io/library/execs:latest"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:662ca58536d5a03864781eb42870f6712927981383f1dab1449d9adf80067313",
      "size": 581,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:f61eef5ff800b2e4366c6d2e7311e95ce8dc3cff5c020a3a810429656fd93efb",
      "size": 581,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```

This behavior can create troubles, as when the manifests in `index.json` will not be automatically deleted when the surrounding `index` is deleted:

```bash
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ sudo ls /var/lib/ig/oci-store/blobs/sha256                                 main % u=
225a8e697e260684d9451ccf8a622cd5548b0ae5fac6e13c416cd1d85247b7b7  b4be31f07444fb9be6dee8cd1349aab3c780d333b63406be10c5c14c1b39dd37
662ca58536d5a03864781eb42870f6712927981383f1dab1449d9adf80067313  c06a86dad7e582686b53bec9c729b92fc56c838b59cc11d0105c2c20f3158dcd
b0eff7f0ec2a27b8d71fe904cf373fa2002c6d669db54c25a9542cf2cb5f46d1  f61eef5ff800b2e4366c6d2e7311e95ce8dc3cff5c020a3a810429656fd93efb
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ sudo -E ./ig image remove execs                                            main % u=
INFO[0000] Experimental features enabled                
Successfully removed execs
[...]

PR #680, opened 2024-01-18
feat: mark untagged manifests as garbage during GC and delete

wangxiaoxuan273:
Related to #664


content/oci/oci.go
```diff
@@ -526,6 +512,38 @@ func (s *Store) GC(ctx context.Context) error {
 	return nil
 }
 
+// GCIndex reloads the index and updates metadata.
+func (s *Store) GCIndex(ctx context.Context) error {
```
Wwwsylvia, line 516: This method should be unexported.

```suggestion
func (s *Store) gcIndex(ctx context.Context) error {
```

Wwwsylvia (changes requested):

content/oci/oci.go
```diff
@@ -190,9 +190,8 @@ func (s *Store) Delete(ctx context.Context, target ocispec.Descriptor) error {
 		}
 		if s.AutoGC {
 			for _, d := range danglings {
-				// do not delete existing manifests in tagResolver
-				_, err = s.tagResolver.Resolve(ctx, string(d.Digest))
-				if errors.Is(err, errdef.ErrNotFound) {
+				// do not delete existing tagged manifests
+				if !s.isTagged(d) {
```
Wwwsylvia, line 194: Since we now automatically delete untagged successor manifests when `Store.AutoGC` is enabled, we can further automatically delete referrers as well. That is, we no longer need the `s.AutoDeleteReferrers` option. 
Instead, enabling `Store.AutoGC` will mean:

- Automatically delete untagged successor manifests while keeping tagged successor manifests
- Automatically delete referrers, no mater it is tagged or not

If we agree on this, we will need another PR to make this change. /cc @shizhMSFT

content/oci/oci.go
```diff
@@ -526,6 +512,38 @@ func (s *Store) GC(ctx context.Context) error {
 	return nil
 }
 
+// gcIndex reloads the index and updates metadata.
```
Wwwsylvia, line 515: nit: Should we improve this comment a bit? Like mentioning that we clean up the untagged descriptors and leave tagged ones.

Wwwsylvia (approved):
LGTM with a minor suggestion

content/oci/oci.go
wangxiaoxuan273: Improved the doc.

content/oci/oci.go
wangxiaoxuan273: See #683

internal/resolver/memory.go
```diff
@@ -19,48 +19,93 @@ import (
...
-		return true
-	})
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	res := make(map[string]ocispec.Descriptor, len(m.index))
+	for key, value := range m.index {
+		res[key] = value
+	}
+	return res
```
shizhMSFT, line 97: FYI: [maps.Clone](https://pkg.go.dev/maps#Clone) is introduced since `go 1.21`. Based on the `oras-go` support window, we can start using `maps.Clone` after the `go 1.22` release.

shizhMSFT (approved):
LGTM

internal/resolver/memory.go
shizhMSFT: /cc @Wwwsylvia

content/oci/oci.go
```diff
@@ -526,6 +512,39 @@ func (s *Store) GC(ctx context.Context) error {
...
 }
 
+// gcIndex reloads the index and updates metadata. Information of untagged blobs
+// are cleaned and only tagged blobs remain.
+func (s *Store) gcIndex(ctx context.Context) error {
+	refMap := s.tagResolver.Map()
+	s.tagResolver = resolver.NewMemory()
+	s.graph = graph.NewMemory()
+	for ref, desc := range refMap {
+		if ref == desc.Digest.String() {
```
Wwwsylvia, line 522: Just realized, for the following graph in the OCI store,

```mermaid
graph TD
    Referrer --> Manifest
```
Where `Manifest` is tagged while `Referrer` is not tagged.
If we run `GC()`, will `Referrer` be deleted?

Wwwsylvia (changes requested):

internal/graph/memory.go
```diff
@@ -186,3 +186,10 @@ func (m *Memory) index(ctx context.Context, fetcher content.Fetcher, node ocispe
 	}
 	return successors, nil
 }
+
+// Exists checks if the node exists in the graph
+func (m *Memory) Exists(node ocispec.Descriptor) bool {
+	nodeKey := descriptor.FromOCI(node)
+	_, exists := m.nodes[nodeKey]
```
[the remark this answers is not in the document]
wangxiaoxuan273: added the lock, thanks for catching.

internal/manifestutil/parser.go
```diff
@@ -61,3 +62,23 @@ func Manifests(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descri
...
+		}
+		var manifest struct {
+			Subject *ocispec.Descriptor `json:"subject,omitempty"`
+		}
+		if err := json.Unmarshal(content, &manifest); err != nil {
+			return nil, err
+		}
+		return manifest.Subject, nil
+	default:
+		return nil, nil
```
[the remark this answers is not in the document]
wangxiaoxuan273: I think when the media type does not have a subject, it can just be silently ignored. We don't need to return an error.

content/oci/oci.go
wangxiaoxuan273: Updated our implementation, so that now when `manifest` is tagged and `referrer` is not tagged, the referrer will not be deleted during `GC`. The rule is that nodes reachable (by referrer and successor relation) from tagged manifests will not be deleted.

wangxiaoxuan273:
> Hi!
> 
> Thank you for this PR! As advised in #664, I tested this PR in a [branch](https://github.com/inspektor-gadget/inspektor-gadget/compare/main...francis/rmi-upstream):
> 
> ```shell
> francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ sudo -E ./ig image build -t opens -f gadgets/trace_open/gadget.yaml gadgets/trace_open
> INFO[0000] Experimental features enabled                
> Successfully built ghcr.io/inspektor-gadget/gadget/opens:latest@sha256:a5de3655d6c7640eb6d43f7d9d7182b233ac86aedddfe6c132cba6b876264d97
> francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ sudo cat /var/lib/ig/oci-store/index.json | jq             francis/rmi-upstream % u=
> {
>   "schemaVersion": 2,
>   "manifests": [
>     {
>       "mediaType": "application/vnd.oci.image.index.v1+json",
>       "digest": "sha256:a5de3655d6c7640eb6d43f7d9d7182b233ac86aedddfe6c132cba6b876264d97",
>       "size": 491,
>       "annotations": {
>         "org.opencontainers.image.ref.name": "ghcr.io/inspektor-gadget/gadget/opens:latest"
>       }
>     },
>     {
>       "mediaType": "application/vnd.oci.image.manifest.v1+json",
>       "digest": "sha256:6fc908439cd210687651e36693d4fa96e59259403826c7cf929ac5382c8d6a33",
>       "size": 582,
>       "platform": {
>         "architecture": "amd64",
>         "os": "linux"
>       }
>     },
>     {
>       "mediaType": "application/vnd.oci.image.manifest.v1+json",
>       "digest": "sha256:b4c03573365a929cbde0bf6e92ed91e067647501851a6d84a6f2a1a072bf22ce",
>       "size": 582,
>       "platform": {
>         "architecture": "arm64",
>         "os": "linux"
>       }
>     }
>   ]
> }
> francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ sudo -E ./ig image remove opens                            francis/rmi-upstream % u=
> INFO[0000] Experimental features enabled                
> Successfully removed opens
> francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ sudo cat /var/lib/ig/oci-store/index.json | jq             francis/rmi-upstream % u=
> {
>   "schemaVersion": 2,
>   "manifests": [
>     {
>       "mediaType": "application/vnd.oci.image.index.v1+json",
>       "digest": "sha256:a5de3655d6c7640eb6d43f7d9d7182b233ac86aedddfe6c132cba6b876264d97",
>       "size": 491
>     },
>     {
>       "mediaType": "application/vnd.oci.image.manifest.v1+json",
>       "digest": "sha256:6fc908439cd210687651e36693d4fa96e59259403826c7cf929ac5382c8d6a33",
>       "size": 582,
>       "platform": {
>         "architecture": "amd64",
>         "os": "linux"
>       }
>     },
>     {
>       "mediaType": "application/vnd.oci.image.manifest.v1+json",
>       "digest": "sha256:b4c03573365a929cbde0bf6e92ed91e067647501851a6d84a6f2a1a072bf22ce",
>       "size": 582,
>       "platform": {
>         "architecture": "arm64",
>         "os": "linux"
>       }
>     }
>   ]
> }
> ```
> 
> As a result, the tag was just removed from the index but nothing no blobs were actually removed. Am I missing something?
> 
> With regard to the code itself, I found two nits.
> 
> Best regards.

Hi @eiffel-fl,

If I understand your scenario correctly, when you delete an image index, you also want the manifests in the index to be automatically removed, given that the manifests are untagged and do not have other tagged predecessors. 

I have added a simple [unit test](https://github.com/oras-project/oras-go/pull/680/files#diff-ea481518062f08e79710c25e05d0f3c1160963c9e0ca218d0e7ff5565c10b34e:~:text=func%20TestStore_GCAndDeleteOnIndex(t%20*testing.T)%20%7B) to cover the scenario of deleting an image index. It looks like it's working as expected. In this unit test, untagged manifests are all removed and the tagged one remains.

I see that in your branch, you replaced `deleteTree` (which deletes all successors of `root` if they have no other predecessors?) with `Delete` and `GC`. Logically it should work. Do the manifests have no other tagged predecessors, other than the image index?

content/oci/oci.go
```diff
@@ -526,6 +514,72 @@ func (s *Store) GC(ctx context.Context) error {
...
+		// check if the referrers manifest can traverse to the existing graph
+		subject := &desc
+		for {
+			subject, err := manifestutil.Subject(ctx, s.storage, *subject)
+			if err != nil {
+				return err
+			}
+			if subject == nil {
+				break
+			}
```
shizhMSFT, line 556: Can you add a test to cover this path?

wangxiaoxuan273: added

shizhMSFT (approved):
LGTM

content/oci/oci.go
```diff
@@ -57,8 +58,8 @@ type Store struct {
 
 	// AutoGC controls if the OCI store will automatically clean newly produced
 	// dangling (unreferenced) blobs during Delete() operation. For example the
-	// blobs whose manifests have been deleted. Manifests in index.json will not
-	// be deleted.
+	// blobs whose manifests have been deleted. Tagged manifests will not be
```
Wwwsylvia, line 61: Just a note, we will need to revise this doc in #683.

content/oci/oci.go
```diff
@@ -526,6 +514,72 @@ func (s *Store) GC(ctx context.Context) error {
...
 
+// gcIndex reloads the index and updates metadata. Information of untagged blobs
+// are cleaned and only tagged blobs remain.
+func (s *Store) gcIndex(ctx context.Context) error {
+	// index tagged manifests
+	refMap := s.tagResolver.Map()
+	tagResolver := resolver.NewMemory()
+	graph := graph.NewMemory()
+	tagged := set.New[digest.Digest]()
+	for ref, desc := range refMap {
```
Wwwsylvia, line 525: nit: something like this might be more readable

```suggestion
	tagResolver := resolver.NewMemory()
	graph := graph.NewMemory()
	tagged := set.New[digest.Digest]()

	// index tagged manifests
	refMap := s.tagResolver.Map()
```

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

Issue #685, opened 2024-01-19
Upgrade to `distribution-spec v1.1.0-rc4`


shizhMSFT:
[`distribution-spec v1.1.0-rc4`](https://github.com/opencontainers/distribution-spec/releases/tag/v1.1.0-rc4) (alias: [`distribution-spec v1.1.0-rc.4`](https://github.com/opencontainers/distribution-spec/releases/tag/v1.1.0-rc.4)) has just been released.

The changes related to `oras-go` include:
- No longer require checking 404 on Referrers API but checking the `OCI-Subject` header on pushing manifest with `subject`.
- The `Content-Type` header of the Referrers API response MUST be set to `application/vnd.oci.image.index.v1+json`.
  - Fixed by #635

PR #690, opened 2024-01-23
refactor: upgrade to `distribution-spec v1.1.0-rc4`

Wwwsylvia:
1. If the `OCI-Subject` header is not set in the response of manifest push, do not ping the Referrers API (see `repository.go`)
2. Update tests accordingly
3. Update the spec version in comments

Resolve: #685


shizhMSFT (approved):
LGTM

registry/remote/repository.go
```diff
@@ -1423,22 +1431,17 @@ func (s *manifestStore) indexReferrersForPush(ctx context.Context, desc ocispec.
 		return nil
 	}
 
-	ok, err := s.repo.pingReferrers(ctx)
```
shizhMSFT, line 1426: Since `pingReferrers` is now only used in the delete path, we can remove the `zeroDigest` hack.

Wwwsylvia: Found the old converstation: https://github.com/opencontainers/distribution-spec/issues/359#issuecomment-1294902270

This design was for performance consideration. We used the digest of a non-existing manifest to avoid server to query the referrers.

PR #715, opened 2024-02-27
Bump go version

dextrot:
This PR fixes #713
- bumps go version in `README.md`, `go.mod` and `GitHub Actions`.

PR #725, opened 2024-03-19
test: add a new case for UnwrapNopCloser

shizhMSFT:
Add a new unit test case by the change to `UnwrapNopCloser` introduced by #715


Wwwsylvia (approved):
LGTM

Issue #713, opened 2024-02-22
Shift Go Support Window to `[1.21, 1.22]`


shizhMSFT:
[Go 1.22](https://go.dev/blog/go1.22) is released with [release notes](https://go.dev/doc/go1.22)!

It is time to shift the window for go support to `[1.21, 1.22]` by updating

- `go. mod`
- GitHub Actions
- `README.md`

After shifting the golang support window, we can remove / update some internal implementations using the latest golang built-in libraries introduced in golang `1.21`.

- Remove the package [`oras.land/oras-go/internal/slices`](https://github.com/oras-project/oras-go/blob/v2.4.0/internal/slices/slice.go) and replace it with the golang built-in package [`slices`](https://pkg.go.dev/slices).
- Replace map clone logic in `Map()` and `TagSet()` of [`internal/resolver/memory.go`](https://github.com/oras-project/oras-go/blob/v2.4.0/internal/resolver/memory.go) with the built-in [`maps.Clone`](https://pkg.go.dev/maps#Clone) to improve performance.

dextrot:
Hi @shizhMSFT can you please assign this issue to me?

shizhMSFT:
@dextrot Sure. Thanks for contributing to `oras-go`.

You may follow #587 for references.

dextrot:
> @dextrot Sure. Thanks for contributing to `oras-go`.
> 
> You may follow #587 for references.

Thank you @shizhMSFT

shizhMSFT:
@dextrot There is a build error in #715. Could you fix it according to https://github.com/oras-project/oras-go/pull/715#pullrequestreview-1919141893 as we need it in the next version `v2.5.0`, which is planned to be released in the near future? It is OK if you want to delegate it to other contributors.

dextrot:
> @dextrot There is a build error in #715. Could you fix it according to [#715 (review)](https://github.com/oras-project/oras-go/pull/715#pullrequestreview-1919141893) as we need it in the next version `v2.5.0`, which is planned to be released in the near future? It is OK if you want to delegate it to other contributors.

Apologies, taking a look now.

PR #726, opened 2024-03-19
perf: use new built-in methods introduced before go 1.22

shizhMSFT:
Changes:
* chore: use built-in slices
* perf: use built-in map clone
* chore: use built-in map copy
* perf: use slices.Sort instead of sort.Strings

Resolves: #713


wangxiaoxuan273:
LGTM although IANAM

Wwwsylvia (approved):
LGTM

PR #728, opened 2024-03-19
fix(DynamicStore): retry `setCredsStore` on next `PUT`

### LINKED ISSUE #727: Subsequent `DynamicStore.Put` calls do not set credential store once failure
In the following code block:
https://github.com/oras-project/oras-go/blob/8d139f0be9190ee4e48fe00cda08e0185ccd2494/registry/remote/credentials/store.go#L136-L152

If `ds.setCredsStoreOnce` fails on the first call to `Put`, the subsequent calls to `Put` will never attempt to set the credential store again but return success.

shizhMSFT:
Fix #727


internal/syncutil/once.go
```diff
@@ -68,3 +72,31 @@ func (o *Once) Do(ctx context.Context, f func() (interface{}, error)) (bool, int
...
+	// slow path
+	o.lock.Lock()
+	defer o.lock.Unlock()
+
+	if o.done.Load() {
+		return nil
+	}
+	if err := f(); err != nil {
+		return err
+	}
```
Wwwsylvia, line 99: This is cool!
nit: The name `OnceOrRetry` might be confusing as it does not convey the idea that the function will only run once successful. How about something like `OnceSuccess`?🤔

shizhMSFT: SGTM

shizhMSFT: There is a problem that `OnceSuccess` is not correct in grammar. On the other hand, `OnceSuccess` has ambiguity on the return value of the subsequent calls after the success. It could mean "succeed once and fail later".

Wwwsylvia: OK. It's an internal struct so never mind😂

Wwwsylvia (approved):
LGTM

PR #739, opened 2024-04-07
fix: cancel goroutine before the next one is created

wangxiaoxuan273:
To fix #738

### DIFF
--- a/copy_test.go
+++ b/copy_test.go
@@ -1779,7 +1779,9 @@ func TestCopyGraph_WithOptions(t *testing.T) {
 	t.Run("MountFrom error", func(t *testing.T) {
 		root = descs[6]
 		dst := &countingStorage{storage: cas.NewMemory()}
-		opts = oras.CopyGraphOptions{}
+		opts = oras.CopyGraphOptions{
+			Concurrency: 1,
+		}
 		var numMountFrom atomic.Int64
 		e := errors.New("mountFrom error")
 		opts.MountFrom = func(ctx context.Context, desc ocispec.Descriptor) ([]string, error) {
@@ -1790,7 +1792,8 @@ func TestCopyGraph_WithOptions(t *testing.T) {
 			t.Fatalf("CopyGraph() error = %v, wantErr %v", err, e)
 		}
 
-		if got, expected := dst.numExists.Load(), int64(7); got != expected {
+		// with a very low probability, dst.numExists may be 3
+		if got, expected := dst.numExists.Load(), int64(4); got != expected {
 			t.Errorf("count(Exists()) = %d, want %d", got, expected)
 		}
 		if got, expected := dst.numFetch.Load(), int64(0); got != expected {
@@ -1799,7 +1802,7 @@ func TestCopyGraph_WithOptions(t *testing.T) {
 		if got, expected := dst.numPush.Load(), int64(0); got != expected {
 			t.Errorf("count(Push()) = %d, want %d", got, expected)
 		}
-		if got, expected := numMountFrom.Load(), int64(4); got != expected {
+		if got, expected := numMountFrom.Load(), int64(1); got != expected {
 			t.Errorf("count(MountFrom()) = %d, want %d", got, expected)
 		}
 	})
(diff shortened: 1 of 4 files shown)


copy_test.go
```diff
@@ -1790,7 +1792,8 @@ func TestCopyGraph_WithOptions(t *testing.T) {
 			t.Fatalf("CopyGraph() error = %v, wantErr %v", err, e)
 		}
 
-		if got, expected := dst.numExists.Load(), int64(7); got != expected {
+		// with a very low probability, dst.numExists may be 3
```
Wwwsylvia, line 1795: Why?🤔

copy_test.go
```diff
@@ -1779,7 +1779,9 @@ func TestCopyGraph_WithOptions(t *testing.T) {
 	t.Run("MountFrom error", func(t *testing.T) {
 		root = descs[6]
 		dst := &countingStorage{storage: cas.NewMemory()}
-		opts = oras.CopyGraphOptions{}
+		opts = oras.CopyGraphOptions{
+			Concurrency: 1,
+		}
```
Wwwsylvia, line 1784: What is the consideration?

copy_test.go
wangxiaoxuan273: Due to the structure of the graph being copied, there are two possible orders of visiting each node. With a very low probability, one of the orders will make `dst.numExists` 3. Under most of the time the value will be 4.

copy_test.go
wangxiaoxuan273: We want to make the run result deterministic. So we want to make sure that only one goroutine is run at a time, basically make it sequential.

shizhMSFT (approved):
LGTM

copy_test.go
Wwwsylvia: > Due to the structure of the graph being copied, there are two possible orders of visiting each node. With a very low probability, one of the orders will make dst.numExists 3.

Would this impact other unit tests or user code as well?

copy_test.go
Wwwsylvia: Got it. Can you add this explanation to the comment so that others looking at the code can understand?

copy_test.go
wangxiaoxuan273: > > Due to the structure of the graph being copied, there are two possible orders of visiting each node. With a very low probability, one of the orders will make dst.numExists 3.
> 
> Would this impact other unit tests or user code as well?

Under our current knowledge, no, this is the only place impacted.

copy_test.go
wangxiaoxuan273: Added a comment

copy_test.go
wangxiaoxuan273: Oops, looks like the value can also be 5, under `go 1.21`.
image

internal/syncutil/limit.go
```diff
@@ -68,15 +69,24 @@ type GoFunc[T any] func(ctx context.Context, region *LimitedRegion, t T) error
...
 		eg.Go(func(t T) func() error {
 			return func() error {
 				defer region.End()
-				return fn(egCtx, region, t)
+				// cancel the gorountine before the next goroutine is created
+				err := fn(egCtx, region, t)
+				if err != nil {
+					egErr.CompareAndSwap(nil, err)
+				}
+				return err
```
JeyJeyGao, line 89: nit:
```suggestion
				if err != nil {
					egErr.CompareAndSwap(nil, err)
					return err
				}
				return nil
```

internal/syncutil/limit.go
```diff
@@ -68,15 +69,24 @@ type GoFunc[T any] func(ctx context.Context, region *LimitedRegion, t T) error
...
+	var egErr atomic.Value
 	for _, item := range items {
-		region := LimitRegion(ctx, limiter)
+		region := LimitRegion(egCtx, limiter)
 		if err := region.Start(); err != nil {
+			if egErr, ok := egErr.Load().(error); ok && egErr != nil {
+				return egErr
+			}
 			return err
 		}
```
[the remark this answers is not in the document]
wangxiaoxuan273: I don't think it is possible, since we only change `egErr` when `err != nil` in Line 87.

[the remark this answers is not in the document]
JeyJeyGao: I don’t quite understand why we need to check `egErr` when `err` is not nil. If we don’t check it, will it still return the error and cancel the other goroutines?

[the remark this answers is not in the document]
wangxiaoxuan273: > I don’t quite understand why we need to check `egErr` when `err` is not nil. If we don’t check it, will it still return the error and cancel the other goroutines?

If we don't check it, the error returned will always be context cancelled error. We check it to return the exact error that causes the error cancellation.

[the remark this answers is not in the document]
JeyJeyGao: I see. Thanks for your explanation.

JeyJeyGao (approved):
LGTM. But I am not a maintainer.

Wwwsylvia (approved):
LGTM

PR #749, opened 2024-04-23
docs: document PackManifestOptions to make PackManifest reproducible

### LINKED ISSUE #748: docs: document `PackManifestOptions` to make `PackManifest` reproducible
[PackManifest](https://pkg.go.dev/oras.land/oras-go/v2#PackManifest) packs an artifact by generating a manifest with annotation `org.opencontainers.image.created`.

Each time calling `PackManifest`, a new time stamp is generated for `org.opencontainers.image.created` even with the same content. Therefore, building an artifact is not [reproducible](https://en.wikipedia.org/wiki/Reproducible_builds).

The `PackManifest` method can be _reproducible_ when `ocispec.AnnotationCreated` is set. However, this fact is not called out in the docs of [PackManifestOptions](https://pkg.go.dev/oras.land/oras-go/v2#PackManifestOptions) as well as [PackManifest](https://pkg.go.dev/oras.land/oras-go/v2#PackManifest).

This request is to enhance the documentation for reproducibility of `PackManifest`.

wangxiaoxuan273:
Resolves #748

### DIFF
--- a/example_pack_test.go
+++ b/example_pack_test.go
@@ -35,7 +35,8 @@ func ExamplePackManifest_imageV11() {
 	opts := oras.PackManifestOptions{
 		ManifestAnnotations: map[string]string{
 			// this timestamp will be automatically generated if not specified
-			// use a fixed value here in order to test the output
+			// use a fixed value here in order to test the output and make the
+			// pack reproducible.
 			ocispec.AnnotationCreated: "2000-01-01T00:00:00Z",
 		},
 	}
@@ -71,7 +72,8 @@ func ExamplePackManifest_imageV10() {
 	opts := oras.PackManifestOptions{
 		ManifestAnnotations: map[string]string{
 			// this timestamp will be automatically generated if not specified
-			// use a fixed value here in order to test the output
+			// use a fixed value here in order to test the output and make the
+			// pack reproducible.
 			ocispec.AnnotationCreated: "2000-01-01T00:00:00Z",
 		},
 	}
--- a/example_pack_test.go
+++ b/example_pack_test.go
@@ -34,7 +34,7 @@ func ExamplePackManifest_imageV11() {
 	// 1. Set optional parameters
 	opts := oras.PackManifestOptions{
 		ManifestAnnotations: map[string]string{
-			// this timestamp will be automatically generated if not specified
+			// this time stamp will be automatically generated if not specified
 			// use a fixed value here to make the pack result reproducible
 			ocispec.AnnotationCreated: "2000-01-01T00:00:00Z",
 		},
@@ -70,7 +70,7 @@ func ExamplePackManifest_imageV10() {
 	// 1. Set optional parameters
 	opts := oras.PackManifestOptions{
 		ManifestAnnotations: map[string]string{
-			// this timestamp will be automatically generated if not specified
+			// this time stamp will be automatically generated if not specified
 			// use a fixed value here to make the pack result reproducible
 			ocispec.AnnotationCreated: "2000-01-01T00:00:00Z",
 		},
--- a/example_pack_test.go
+++ b/example_pack_test.go
@@ -35,8 +35,7 @@ func ExamplePackManifest_imageV11() {
 	opts := oras.PackManifestOptions{
 		ManifestAnnotations: map[string]string{
 			// this timestamp will be automatically generated if not specified
-			// use a fixed value here in order to test the output and make the
-			// pack reproducible.
+			// use a fixed value here to make the pack result reproducible
 			ocispec.AnnotationCreated: "2000-01-01T00:00:00Z",
 		},
 	}
@@ -72,8 +71,7 @@ func ExamplePackManifest_imageV10() {
 	opts := oras.PackManifestOptions{
 		ManifestAnnotations: map[string]string{
 			// this timestamp will be automatically generated if not specified
-			// use a fixed value here in order to test the output and make the
-			// pack reproducible.
+			// use a fixed value here to make the pack result reproducible
 			ocispec.AnnotationCreated: "2000-01-01T00:00:00Z",
 		},
 	}
--- a/pack.go
+++ b/pack.go
@@ -94,8 +94,9 @@ type PackManifestOptions struct {
 	Layers []ocispec.Descriptor
 
 	// ManifestAnnotations is the annotation map of the manifest. In order to
-	// make PackManifest reproducible, set the key ocispec.AnnotationCreated to
-	// a fixed value. The value must conform to RFC 3339.
+	// make [PackManifest] reproducible, set the key ocispec.AnnotationCreated
+	// (i.e. "org.opencontainers.image.created") to a fixed value. The value
+	// must conform to RFC 3339.
 	ManifestAnnotations map[string]string
 
 	// ConfigDescriptor is a pointer to the descriptor of the config blob.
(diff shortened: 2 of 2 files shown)


pack.go
```diff
@@ -126,6 +128,11 @@ var mediaTypeRegexp = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126
 //
 // artifactType and opts.ConfigDescriptor.MediaType MUST comply with RFC 6838.
 //
+// Each time when PackManifest is called, it generates a new time stamp in the
+// manifest annotations with the key "org.opencontainers.image.created". To make
```
Wwwsylvia, line 132: We also need to mention that if "org.opencontainers.image.created" is specified, it must conform to [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html).

pack.go
```diff
@@ -93,7 +93,9 @@ type PackManifestOptions struct {
 	// Layers is the layers of the manifest.
 	Layers []ocispec.Descriptor
 
-	// ManifestAnnotations is the annotation map of the manifest.
+	// ManifestAnnotations is the annotation map of the manifest. Set
+	// "org.opencontainers.image.created" with a value to make PackManifest
+	// reproducible.
```
Wwwsylvia, line 98: The timestamp must be in [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) format.

pack.go
Wwwsylvia, line 1: We may also want to update the doc of `ErrInvalidDateTimeFormat` and replace `AnnotationCreated ` with the literal string "org.opencontainers.image.created".

Wwwsylvia (changes requested):
We also need to rephrase the related comment a bit for `ExamplePackManifest_imageV10` and `ExamplePackManifest_imageV11` in `example_pack_test.go`.

Wwwsylvia:
BTW, the PR title should start with `docs` (end with `s`) as per the [conventional commit guidance](https://www.conventionalcommits.org/en/v1.0.0/#summary). 😀

pack.go
```diff
@@ -93,7 +93,9 @@ type PackManifestOptions struct {
 	// Layers is the layers of the manifest.
 	Layers []ocispec.Descriptor
 
-	// ManifestAnnotations is the annotation map of the manifest.
+	// ManifestAnnotations is the annotation map of the manifest. Set
+	// "org.opencontainers.image.created" with a value to make PackManifest
```
shizhMSFT, line 97: Developers is not encouraged to use string literals but constants. Please refer the developers to use constants from ocispec.

pack.go
```diff
@@ -126,6 +128,11 @@ var mediaTypeRegexp = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126
 //
 // artifactType and opts.ConfigDescriptor.MediaType MUST comply with RFC 6838.
 //
+// Each time when PackManifest is called, it generates a new time stamp in the
+// manifest annotations with the key "org.opencontainers.image.created". To make
+// PackManifest reproducible, set "org.opencontainers.image.created" with a value
```
shizhMSFT, line 133: Same here for consts.

shizhMSFT: This comment is a duplicate of https://github.com/oras-project/oras-go/pull/749/files#r1575970704

pack.go
shizhMSFT: This comment is a duplicate of https://github.com/oras-project/oras-go/pull/749/files#r1575970704

shizhMSFT (commented):
You also need to update the examples in `example_pack_test.go`.

pack.go
Wwwsylvia: Actually we are talking about different stuffs😀 In [another comment](https://github.com/oras-project/oras-go/pull/749#discussion_r1575970704) my concern was that users may not understand what `AnnotationArtifactCreated` and `AnnotationCreated` are as they are not defined in `oras-go`, and we may want to use string literals in the comment.

https://github.com/oras-project/oras-go/blob/6fce64e956e850a3af7ea16e85959c23e3eddc4b/pack.go#L50-L54

pack.go
wangxiaoxuan273: Added a comment.

pack.go
wangxiaoxuan273: Added a comment.

pack.go
wangxiaoxuan273: Replaced both `AnnotationArtifactCreated` and `AnnotationCreated` with the corresponding literal strings.

pack.go
wangxiaoxuan273: > Actually we are talking about different stuffs😀 In [another comment](https://github.com/oras-project/oras-go/pull/749#discussion_r1575970704) my concern was that users may not understand what `AnnotationArtifactCreated` and `AnnotationCreated` are as they are not defined in `oras-go`, and we may want to use string literals in the comment.
> 
> https://github.com/oras-project/oras-go/blob/6fce64e956e850a3af7ea16e85959c23e3eddc4b/pack.go#L50-L54

Has a point. Kept using the string literal in the comment for the error value, and replaced with the variable name `ocispec.AnnotationCreated` for comments of the function and annotation map.

pack.go
wangxiaoxuan273: Replaced the const with the variable name.

wangxiaoxuan273:
> We also need to rephrase the related comment a bit for `ExamplePackManifest_imageV10` and `ExamplePackManifest_imageV11` in `example_pack_test.go`.

updated the comments  in `example_pack_test.go`.

example_pack_test.go
```diff
@@ -35,7 +35,8 @@ func ExamplePackManifest_imageV11() {
 	opts := oras.PackManifestOptions{
 		ManifestAnnotations: map[string]string{
 			// this timestamp will be automatically generated if not specified
-			// use a fixed value here in order to test the output
+			// use a fixed value here in order to test the output and make the
```
Wwwsylvia, line 38: nit: How about:
```suggestion
			// use a fixed value here to make the pack result reproducible
```
?

pack.go
```diff
@@ -126,6 +128,11 @@ var mediaTypeRegexp = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126
 //
 // artifactType and opts.ConfigDescriptor.MediaType MUST comply with RFC 6838.
 //
+// Each time when PackManifest is called, it generates a new time stamp in the
+// manifest annotations with the key ocispec.AnnotationCreated. To make
```
Wwwsylvia, line 132: nit: Do we need to explicitly call out that the timestamp is generated only when `ocispec.AnnotationCreated` annotation is not specified?

pack.go
```diff
@@ -93,7 +93,9 @@ type PackManifestOptions struct {
 	// Layers is the layers of the manifest.
 	Layers []ocispec.Descriptor
 
-	// ManifestAnnotations is the annotation map of the manifest.
+	// ManifestAnnotations is the annotation map of the manifest. In order to
+	// make PackManifest reproducible, set the key ocispec.AnnotationCreated to
```
shizhMSFT, line 97:
```suggestion
	// make [PackManifest] reproducible, set the key ocispec.AnnotationCreated to
```

pack.go
shizhMSFT: Consider the wording like `ocispec.AnnotationCreated (i.e. "org.opencontainers.image.created")`

pack.go
wangxiaoxuan273: Modified the comment

shizhMSFT (approved):
LGTM

Wwwsylvia (approved):
LGTM

Wwwsylvia (approved):
LGTM

Issue #750, opened 2024-04-24
exec oras pull, and set allowpathTraversalonwrite True, and get --output to other directory, but is pull to orginal path


lht2017:
when see code in content/file.go
in the method  resolveWritePath, when allowpathTraversalonwrite  is True, then return the original path to store the pull files
is some bug? or it should like this bellow:

![image](https://github.com/oras-project/oras-go/assets/30956002/2ee11d77-acd4-4e88-a08f-978911c4dda1)

Wwwsylvia:
Hi @lht2017 , thanks for opening this issue. 

The purpose of the `resolveWritePath` function is to resolve and validate the target write path of the file name (`name`). If the target path (`path`) is outside of the working directory, it returns `ErrPathTraversalDisallowed`. Otherwise, it returns the absolute path (`path`) of the file name (`name`).
That is, when `AllowPathTraversalOnWrite` is `true`, this function returns the **absolute** path obtained at line `606` (`path := s.absPath(name)`).

https://github.com/oras-project/oras-go/blob/9b6f32158776a699b23edb3db86d053623619b60/content/file/file.go#L604-L633

If you have observed any unexpected behavior, could you please provide additional details? Specifically:

1. What actions did you take, and what were your expectations?
2. What was your environment?
3. How can we reproduce the issue?
4. Are there any relevant error logs?

(We hope to have an [issue template](https://github.com/oras-project/oras-go/issues/641), but it is still [working in progress](https://github.com/oras-project/oras-go/pull/716).)

lht2017:
Thanks for replay!
1. What actions did you take, and what were your expectations?
first I user oras push a file , like /usr/local/tmp/a.txt  to a remote registry ,
then use oras command to pull files from the remote registry, and use param -o , to other directory   like /pull/data/a.txt , but the file a.txt not write in /pull/data/a.txt , actually it always write in the push path, like /usr/local/tmp/a.txt
2 What was your environment?
linux el7
3How can we reproduce the issue?
use step 1

Wwwsylvia:
Hi @lht2017 , this behavior is by design. The `--output` parameter is only applicable to **relative paths**, and the `--allow-path-traversal` means to allow storing files **out of the output directory**.
That is to say, if one specified an absolute path on push, for example `/usr/local/tmp/a.txt`, 
```sh
$ oras push $myrepo:$mytag /usr/local/tmp/a.txt --disable-path-validation
```

[...]

PR #751, opened 2024-04-24
fix bug, when oras pull --output & allowpathtraversalonwrite true

lht2017:


Wwwsylvia:
Hi @lht2017 , thanks for the PR. Before we can review the PR, let's discuss the issue in #750 .

PR #793, opened 2024-07-21
fix: update error returned when `oci.NewFromTar` is used with a non-tar file

### LINKED ISSUE #640: improve the returned error of `oci.NewFromTar` when the passed-in path is not a tarfile
[oci.NewFromTar](https://pkg.go.dev/oras.land/oras-go/v2/content/oci#NewFromTar) can be used to create a new read-only OCI store from a tarfile path. If the passed-in path doesn't point to a tarfile, currently(v2.3.1) below errors will be returned:
- If the length of the file is zero, a `failed to open OCI layout file:  oci-layout: file does not exist` error will be returned
- If the length of the file is not zero, an [io.ErrUnexpectedEOF](https://pkg.go.dev/io#ErrUnexpectedEOF) will be returned

As a user of oras-go SDK, I would like this function to return a unified and exported error type. Also, the error should indicate that the passed-in path doesn't point to an expected tarfile.

ADorigi:
fixes #640 

This pull request includes the following changes:
- Declaring a new error `ErrNotTarFile`
- Condition to check if the given path points to a `.tar` file
- Corresponding tests


internal/fs/tarfs/tarfs.go
```diff
@@ -48,6 +48,13 @@ func New(path string) (*TarFS, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve absolute path for %s: %w", path, err)
 	}
+
+	extension := filepath.Ext(pathAbs)
+
+	if extension != ".tar" {
```
TerryHowe, line 54: Are compressed formats supported here like `tgz`?

ADorigi: No, I believe they will not be.
Is there a list of extensions we need to support?

ADorigi: I can add them to the check and add corresponding tests for them too.

ADorigi: The current tests work with the files `hello-world.tar` and `test.tar`. From my understanding, `.tar`s are the only ones supported as of now.

Wwwsylvia: In `tfs.indexEntries()`, maybe we can return a friendly error if `tr.Next()` fails?

Wwwsylvia (changes requested):

content/oci/readonlyoci_test.go
```diff
@@ -836,3 +837,15 @@ func Test_deleteAnnotationRefName(t *testing.T) {
 		})
 	}
 }
+
+func TestNewFromTar_NonTarFile(t *testing.T) {
+
+	emptyfile := "testdata/emptyfile"
```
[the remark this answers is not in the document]
ADorigi: Yes, apologies. I will correct this.

internal/fs/tarfs/tarfs.go
ADorigi: I think `tfs.indexEntries` would indeed be a better place for this check. We can check if `EOF` is reached before any entries are added to `tfs`. This would infer it is an empty file. While a non-tar file will give us an `io.ErrUnexpectedEOF` while reading it.

We can then give the user our custom exported error.

ADorigi:
Hi @Wwwsylvia 
Please check the updates. 
Now the check for non-tar file takes place in `tfs.indexEntries()`

internal/fs/tarfs/testdata/nonempty.tar
Wwwsylvia, line 1: nit: Since this file is a non-tar file ending with `.tar`, would it be better to call it something like "nontar.tar"?

internal/fs/tarfs/tarfs_test.go
```diff
@@ -268,3 +268,28 @@ func TestTarFS_Stat_Unsupported(t *testing.T) {
...
 }
+
+func TestTarFS_New_NonTarFile(t *testing.T) {
+
+	emptyfile := "testdata/emptyfile.tar"
+	nonemptyfile := "testdata/nonempty.tar"
+	tarwithoutextension := "testdata/tarwithoutextension"
+
+	_, err := New(emptyfile)
+
```
Wwwsylvia, line 279: nit: I think it's more readable without this empty line.

errdef/errors.go
```diff
@@ -28,4 +28,5 @@ var (
 	ErrSizeExceedsLimit   = errors.New("size exceeds limit")
 	ErrUnsupported        = errors.New("unsupported")
 	ErrUnsupportedVersion = errors.New("unsupported version")
+	ErrNotTarFile         = errors.New("non-empty tar file expected")
```
Wwwsylvia, line 31: nit: The error message can be just "not a tar file" or something like that

internal/fs/tarfs/tarfs.go
```diff
@@ -135,8 +136,16 @@ func (tfs *TarFS) indexEntries() error {
...
 			if errors.Is(err, io.EOF) {
+				if len(tfs.entries) == 0 {
+					// indicates the given file is empty
+					return errdef.ErrNotTarFile
+				}
 				break
 			}
+			if errors.Is(err, io.ErrUnexpectedEOF) {
+				// indicates that its either not a tarfile or it is a corrupted one
+				return errdef.ErrNotTarFile
```
Wwwsylvia, line 147: We can also wrap this error with a friendly error message like "{path} is not a valid tar file"

Wwwsylvia:
Hi @ADorigi , the updated PR looks good overall. I left a few nit comments.

internal/fs/tarfs/testdata/nonempty.tar
ADorigi: yes, `nontar.tar` better depicts the file's content. I will change it.

internal/fs/tarfs/tarfs_test.go
ADorigi: Removed the line.

errdef/errors.go
ADorigi: Changed the error string to `invalid tar file`

internal/fs/tarfs/tarfs.go
ADorigi: Wrapped the error with the path of the file.

ADorigi:
Hi @Wwwsylvia
I made the minor changes and updated the error message.

internal/fs/tarfs/tarfs.go
```diff
@@ -135,8 +136,16 @@ func (tfs *TarFS) indexEntries() error {
 		header, err := tr.Next()
 		if err != nil {
 			if errors.Is(err, io.EOF) {
+				if len(tfs.entries) == 0 {
+					// indicates the given file is empty
+					return fmt.Errorf("{%s}: %w", tfs.path, errdef.ErrInvalidTarFile)
```
Wwwsylvia, line 141: nit: Should we further call out that the file is empty?

internal/fs/tarfs/tarfs.go
```diff
@@ -135,8 +136,16 @@ func (tfs *TarFS) indexEntries() error {
...
+				}
 				break
 			}
+			if errors.Is(err, io.ErrUnexpectedEOF) {
+				// indicates that its either not a tarfile or it is a corrupted one
+				return fmt.Errorf("{%s}: %w", tfs.path, errdef.ErrInvalidTarFile)
```
Wwwsylvia, line 147: Oh the `{}` isn't necessary, sorry for the misleading info.

internal/fs/tarfs/tarfs_test.go
```diff
@@ -268,3 +268,25 @@ func TestTarFS_Stat_Unsupported(t *testing.T) {
...
+	}
+
+	_, err = New(nontar)
+	if want := errdef.ErrInvalidTarFile; !errors.Is(err, want) {
+		t.Errorf("TarFS.New(%s) error = %v, wantErr %v", nontar, err, want)
+	}
+
+	_, err = New(tarwithoutextension)
+	if err != nil {
+		t.Errorf("TarFS.New(%s) error = %v, wantErr %v", tarwithoutextension, err, nil)
```
Wwwsylvia, line 290: nit: The test code can be simplified. You may refer to other tests in this file.

internal/fs/tarfs/tarfs.go
ADorigi: yes a simple `file is empty` would be clearer.

internal/fs/tarfs/tarfs.go
ADorigi: Will remove the braces.

internal/fs/tarfs/tarfs_test.go
ADorigi: Will condense the tests similar to the others.

internal/fs/tarfs/tarfs_test.go
```diff
@@ -268,3 +268,22 @@ func TestTarFS_Stat_Unsupported(t *testing.T) {
...
+		_, err := New(name)
+		if want := errdef.ErrInvalidTarFile; !errors.Is(err, want) {
+			t.Errorf("TarFS.New(%s) error = %v, wantErr %v", name, err, want)
+		}
+	}
+
+	tarwithoutextension := "testdata/tarwithoutextension"
+	_, err := New(tarwithoutextension)
+	if err != nil {
+		t.Errorf("TarFS.New(%s) error = %v, wantErr %v", tarwithoutextension, err, nil)
```
Wwwsylvia, line 287: Just realized, should this be put in another test function? As this test function is for non-tar files.

ADorigi: Yes this would be better in its own test. I will create a new test `TestTarFS_New_TarWithoutExtension`.

Wwwsylvia (approved):
LGTM with some thoughts. @shizhMSFT Please also take a look at this PR when you get a chance.

Wwwsylvia:
cc @qweeah who opened the original issue

internal/fs/tarfs/tarfs.go
```diff
@@ -135,8 +136,16 @@ func (tfs *TarFS) indexEntries() error {
...
+				if len(tfs.entries) == 0 {
+					// indicates the given file is empty
+					return fmt.Errorf("%s: file is empty: %w", tfs.path, errdef.ErrInvalidTarFile)
+				}
 				break
 			}
+			if errors.Is(err, io.ErrUnexpectedEOF) {
+				// indicates that its either not a tarfile or it is a corrupted one
+				return fmt.Errorf("%s: %w", tfs.path, errdef.ErrInvalidTarFile)
+			}
```
shizhMSFT, line 148: The `io.ErrUnexpectedEOF` error itself indicates the tar file is corrupted. There is no need to wrap it to `errdef.ErrInvalidTarFile`.

shizhMSFT (changes requested):
Set to `Request changes` to block merge as the original issue might be invalid.

PR #799, opened 2024-07-25
fix: add missing operation to fs path error

### LINKED ISSUE #640: improve the returned error of `oci.NewFromTar` when the passed-in path is not a tarfile

qweeah:
When opening a folder as OCI image layout and there is no `oci-layout` file exists, the error would be

> failed to open OCI layout file:  oci-layout: file does not exist

There is a duplicated space in the error string, because the `Op` field is missing in the returned `fs.PathError`. With fix in this PR, the return error would be
> failed to open OCI layout file: open oci-layout: file does not exist

Fixes #640

### DIFF
--- a/internal/fs/tarfs/tarfs.go
+++ b/internal/fs/tarfs/tarfs.go
@@ -67,7 +67,7 @@ func New(path string) (*TarFS, error) {
 // ValidPath(name), returning a *PathError with Err set to
 // ErrInvalid or ErrNotExist.
 func (tfs *TarFS) Open(name string) (file fs.File, openErr error) {
-	entry, err := tfs.getEntry(name)
+	entry, err := tfs.getEntry("open", name)
 	if err != nil {
 		return nil, err
 	}
@@ -98,7 +98,7 @@ func (tfs *TarFS) Open(name string) (file fs.File, openErr error) {
 // Stat returns a FileInfo describing the file.
 // If there is an error, it should be of type *PathError.
 func (tfs *TarFS) Stat(name string) (fs.FileInfo, error) {
-	entry, err := tfs.getEntry(name)
+	entry, err := tfs.getEntry("stat", name)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


internal/fs/tarfs/tarfs.go
```diff
@@ -108,11 +108,11 @@ func (tfs *TarFS) Stat(name string) (fs.FileInfo, error) {
 // getEntry returns the named entry.
 func (tfs *TarFS) getEntry(name string) (*entry, error) {
 	if !fs.ValidPath(name) {
-		return nil, &fs.PathError{Path: name, Err: fs.ErrInvalid}
+		return nil, &fs.PathError{Op: "open", Path: name, Err: fs.ErrInvalid}
 	}
 	entry, ok := tfs.entries[name]
 	if !ok {
-		return nil, &fs.PathError{Path: name, Err: fs.ErrNotExist}
+		return nil, &fs.PathError{Op: "open", Path: name, Err: fs.ErrNotExist}
```
Wwwsylvia, line 115: Looks like we should define a constant for "open".

qweeah: I don't know, just check `os` package and seems they doesn't add a const
https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/os/file_windows.go

shizhMSFT, line 115: `Stat` also invoke `getEntry`. We should distinguish them.

qweeah: Add operation as input of the getEntry method.

Wwwsylvia (approved):
LGTM

internal/fs/tarfs/tarfs.go
Wwwsylvia: Nevermind, the comment became invalid now.

shizhMSFT (approved):
LGTM

PR #807, opened 2024-08-14
feat(registry): allow repository to use content.ExtendedCopy

### LINKED ISSUE #806: regitry.Repository does not implement content.ReadOnlyGraphStorage
currently `remote.NewRepository("myreg/myrepo")` returns a remote.Repository object
but i use `reg, err := remote.NewRegistry("myreg")` and then call `reg.Repository("myrepo")` it returns a registry.Repository interface. 

Unfortunately this interface is missing the `Predecessors()` method which means that it does not implement the content.ReadOnlyGraphStorage

this means that while i can use `content.CopyGraph`, i cannot use `content.ExtendedCopyGraph`

JefeDavis:
Fixes #806

### DIFF
--- a/registry/repository.go
+++ b/registry/repository.go
@@ -45,6 +45,7 @@ type Repository interface {
 	content.Storage
 	content.Deleter
 	content.TagResolver
+	content.PredecessorFinder
 	ReferenceFetcher
 	ReferencePusher
 	ReferrerLister


shizhMSFT (changes requested):
This is a breaking change. See https://github.com/oras-project/oras-go/issues/806#issuecomment-2288032292 for details.

Issue #812, opened 2024-08-28
Shift Go Support Window to [1.22, 1.23]


Wwwsylvia:
[Go 1.23](https://go.dev/doc/go1.23) has been released on 2024-08-13. It is time to shift the window for go support to [1.22, 1.23].

PR #820, opened 2024-08-30
build(go): shift Go support window to [1.22, 1.23]

Wwwsylvia:
1. Upgrading `go.mod` to `1.22`
2. Using new `for` syntax like `for i := range N`
3. Migrating random generator from package [`math/rand`](https://pkg.go.dev/math/rand) to package [`math/rand/v2`](https://pkg.go.dev/math/rand/v2@go1.22.0)

References:
- [Go 1.22 release notes](https://tip.golang.org/doc/go1.22)
- [Secure Randomness in Go 1.22](https://tip.golang.org/blog/chacha8rand)

Resolves: #812


registry/remote/retry/policy.go
```diff
@@ -88,15 +88,15 @@ type Backoff func(attempt int, resp *http.Response) time.Duration
...
 // The HTTP response is checked for a Retry-After header. If it is present, the
 // value is used as the backoff duration.
 func ExponentialBackoff(backoff time.Duration, factor, jitter float64) Backoff {
 	return func(attempt int, resp *http.Response) time.Duration {
-		var h maphash.Hash
-		h.SetSeed(maphash.MakeSeed())
-		rand := rand.New(rand.NewSource(int64(h.Sum64())))
+		hashSeed := seedFromMapHash()
+		timeSeed := uint64(time.Now().UnixNano())
+		rand := rand.New(rand.NewPCG(hashSeed, timeSeed))
```
shizhMSFT, line 99: What's the advantage of moving to PCG for the randomness of the exponential backoff scenario?

shizhMSFT (changes requested):

shizhMSFT (approved):
LGTM

registry/remote/retry/policy.go
Wwwsylvia: It's new and seems not harmful?

Issue #841, opened 2024-11-07
`Repository.ReferrersListPageSize` may not work for all registries


Wwwsylvia:
The `Repository.ReferrersListPageSize` property was introduced in `oras-go v2.0.0-alpha` to address the issue #42, initially aligning with the [ORAS Artifact Spec](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md#paging-results) for listing referrers. 

The ORAS Artifact Spec was later deprecated with the release of [distribution-spec v1.1.0](https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#listing-referrers), which does not define a pagination protocol for listing referrers. Despite this, some registries still support pagination for the Referrers API, so the `Repository.ReferrersListPageSize` property remains in `oras-go`. It is important to note that this property may not work with all registries. To avoid confusion, we should update our documentation to call out this limitation.

https://github.com/oras-project/oras-go/blob/9b6f32158776a699b23edb3db86d053623619b60/registry/remote/repository.go#L118-L122

PR #842, opened 2024-11-07
docs: update doc for ReferrerListPageSize

Wwwsylvia:
Resolves: #841


Wwwsylvia:
> Do you want to mention the legacy oras artifact spec?

@shizhMSFT I concerned that might introduce confusion as the ORAS artifact spec has been deprecated and users don't need to know it. But maybe I can mention the related issue.

shizhMSFT (approved):
LGTM

Issue #840, opened 2024-11-04
Support narrowly scoped registry authentication


arewm:
Global access might not be granted for an entire registry. Instead, multiple service accounts/robots may be used for narrowly scoped push/pull actions. Registry authentication should be done from most-specific to least-specific.

> When checking for available credentials, the relevant repository is matched against available keys in its hierarchical order, going from most-specific to least-specific. For example, an image pull for my-registry.local/namespace/user/image:latest will result in a lookup in auth.json in the following order:
> * my-registry.local/namespace/user/image
> * my-registry.local/namespace/user
> * my-registry.local/namespace
> * my-registry.local

https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md#format

This is similar to the proposed change in https://github.com/google/go-containerregistry/pull/1966.

shizhMSFT:
Related to #836

shizhMSFT:
This requires a redesign of the auth module, which further requires a design revisit.

WoozyMasta:
For anyone who needs this on `oras-go/v2` today: I ended up building a small external adapter while this is waiting for v3: [orascope](https://github.com/WoozyMasta/orascope)
It adds most-specific repository-path lookup for inline `auths`, keeps helpers host-scoped, and isolates the ORAS auth cache by repository.

It is just a pragmatic v2 workaround, not an alternative proposal for the v3 design.

TerryHowe:
Closing as superseded by #1348, which covers this in more detail.

#1348 keeps the ask here intact — hierarchical, most-specific-first credential lookup following the containers `auth.json` semantics — and adds the two pieces it turns out to depend on: a `RegistryScope` type that can represent `host/namespace` at all (`properties.NewReference` rejects a bare registry and accepts tags, so neither end matches), and a `Login` that doesn't key the credential store host-only.

The relevant finding for this issue: the store-side plumbing already landed. `credentials.StoreOptions.Hierarchical` and `configfile.GetAuthConfigHierarchical` implement exactly the longest-prefix walk quoted above. What's missing is that nothing ever passes it a path — `credentials.CredentialFunc` is still `func(ctx, hostport string)` and `auth.Client` calls it with the bare host, so the walk only ever has a hostname to match. That gap is tracked as blocker #3 in #1348.

[...]

PR #850, opened 2024-11-27
feat: allow loading config from io.Reader

programmer04:
**What this PR does / why we need it:**

Currently, struct `Config` that represents a docker configuration file can be populated only with the function 

`Load(configPath string) (*Config, error)`

that expects to have a file available on the disk and read it. Sometimes, this is a redundant step and a security risk when file content is available in memory. Here is a real-world example

- https://github.com/Kong/gateway-operator/issues/521

Hence, to overcome the aforementioned problems, this PR introduces

`LoadFromReader(r io.Reader)` and `NewFileStoreFromConfig(cfg *config.Config)`

that allow passing anything that implements the reader interface in a non-breaking way. The whole existing API and behavior don't change, some additional methods are introduced, like `SetPath` and `Path` allowing saving config populated this way. So this change won't break anybody's code.

The root cause of such a workaround that I proposed is that `Config` is tightly coupled with the idea of getting configuration only directly from a file, thus to extend but not break, we need to make some compromises. I'm open to suggestions for this PR.

Since there are plans for `v3` version of this library and work related to registry authentication e.g. https://github.com/oras-project/oras-go/issues/840 I think that the whole API for parsing a config can be changed during this transition in a breaking way. So getting rid of coupling to files on disk should be feasible. If this PR is accepted, I can create a dedicated issue to get it sorted out in the `v3`.


Wwwsylvia:
Hi @programmer04, thank you for the PR!

From the original issue https://github.com/Kong/gateway-operator/issues/521, it looks like your scenario involves reading plaintext credentials from `regcred` without needing to update or delete them. Is that correct?

If so, rather than modifying the original credentials file store implementation, we could introduce something like `ReadOnlyFileStore`, which allows creating a store from a reader and does not support write operations on the credentials.

To get unblocked, you can consider implementing the `ReadOnlyFileStore` type in your code base. Since the plan is to support read operations only on the `auth` entry, the implementation would be much more straightforward than the current implementation in `oras-go` (no need to worry about concurrent writes, accidental overwrites, etc.).

If we feel like this functionality is necessary to be in `oras-go`, we can open an issue to track the feature request. What do you think?

Wwwsylvia:
> The problem is that there is a path field in Config (this structure maps JSON Docker creds). It's hard to break that coupling, this is why I proposed a workaround in PR to avoid breaking anything. Maybe you have an idea of how to tackle it differently?

The file store and the internal Config struct was designed for reading/writing Docker config files. If read-only is sufficient, we can implement a new store like `ReadOnlyFileStore` that loads the config from a reader. To support such store, we also need to implement a new Config like `ReadOnlyConfig`. This way we won't add complexity to the existing implementation and can gain better performance in the read-only scenario. 

It would look somewhat like:

```go
type ReadOnlyConfig struct {
	Auths map[string]AuthConfig
}

func LoadFromReader(reader io.Reader) (*ReadOnlyConfig, error) {
	// TODO: read from the reader and decode the content into Auths
	return &ReadOnlyConfig{}, nil
}

func (c *ReadOnlyConfig) GetCredential(serverAddress string) (auth.Credential, error) {
	// TODO: get the credential from Auths with best effort
	return auth.Credential{}, nil
}
```

```go
type ReadOnlyFileStore struct {
	*config.ReadOnlyConfig
}

func NewReadOnlyFileStoreFromReader(reader io.Reader) (*ReadOnlyFileStore, error) {
	cfg, err := config.LoadFromReader(reader)
	if err != nil {
		return nil, err
	}
	return &ReadOnlyFileStore{ReadOnlyConfig: cfg}, nil
}

func (fs *ReadOnlyFileStore) Get(serverAddress string) (auth.Credential, error) {
	return fs.ReadOnlyConfig.GetCredential(serverAddress)
}

func (fs *ReadOnlyFileStore) Put(serverAddress string, cred auth.Credential) error {
	return nil
}

func (fs *ReadOnlyFileStore) Delete(serverAddress string) error {
	return nil
}
```

programmer04:
Thanks for the detailed guidance @Wwwsylvia I've updated PR (code, title, description) according to the solution proposed by you in https://github.com/oras-project/oras-go/pull/850#issuecomment-2551014069. Please take a look

registry/remote/credentials/internal/config/config.go
```diff
@@ -160,13 +160,18 @@ func (cfg *Config) GetCredential(serverAddress string) (auth.Credential, error)
 	cfg.rwLock.RLock()
 	defer cfg.rwLock.RUnlock()
 
-	authCfgBytes, ok := cfg.authsCache[serverAddress]
+	return getCredentialFromCache(cfg.authsCache, serverAddress)
+}
+
+// getCredentialsFromCache is a helper function to get the credential for serverAddress from authsCache in a raw format.
+func getCredentialFromCache(authsCache map[string]json.RawMessage, serverAddress string) (auth.Credential, error) {
```
Wwwsylvia, line 167: We can refactor this in another way using generics, so that it can be reused by both config and readonly config:
Something like:
```go
func matchAuth[V any](auths map[string]V, serverAddress string) (V, bool) {
	if v, ok := auths[serverAddress]; ok {
		return v, true
	}
	for addr, v := range auths {
		if toHostname(addr) == serverAddress {
			return v, true
		}
	}
	var zero V
	return zero, false
}
```

registry/remote/credentials/file_store.go
```diff
@@ -44,6 +45,9 @@ var (
 	// ErrBadCredentialFormat is returned by Put() when the credential format
 	// is bad.
 	ErrBadCredentialFormat = errors.New("bad credential format")
+	// ErrReadOnlyStore is returned for operations
+	// Put(...) and Delete(...) for read-only store.
+	ErrReadOnlyStore = errors.New("cannot modify content of the read-only store")
```
Wwwsylvia, line 50: nit: How about: "operation not allowed: the store is read-only"?

registry/remote/credentials/file_store.go
```diff
@@ -95,3 +99,36 @@ func validateCredentialFormat(cred auth.Credential) error {
 	}
 	return nil
 }
+
+// ReadOnlyFileStore implements a credentials store using the docker configuration file
+// as an input. It supports only Get operation that works in the same way as for standard
+// FileStore.
+type ReadOnlyFileStore struct {
```
Wwwsylvia, line 106: Should we put this in a separate file for better readability?

Wwwsylvia (changes requested):

registry/remote/credentials/internal/config/config.go
programmer04: It's currently used by both stores. For readonly in `LoadFromReader` where I perform parsing all entries to `auth.Credential` because they won't change during the lifetime of such a config, it can be done here. There is info whether the whole config is valid and no computation is needed for getting a credential.

For file backedd config it's used in method `GetCredential`. They need to be parsed and looked up on each call because they may change.

registry/remote/credentials/internal/config/readonly_config.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	for serverAddress := range authsCache {
+		creds, err := getCredentialFromCache(authsCache, serverAddress)
+		if err != nil {
+			return nil, err
+		}
+		cfg.auths[serverAddress] = creds
+	}
+
+	return &cfg, nil
```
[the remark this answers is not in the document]
programmer04: Exactly, my reasoning is that since it's read-only, the whole parsing can be done during creation because entries won't change in the runtime so I can have a map of `auth.Credential` ready for use directly in `GetCredential`, see also the comment in https://github.com/oras-project/oras-go/pull/850#discussion_r1893876225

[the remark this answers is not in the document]
Wwwsylvia: @programmer04 Sorry for the delayed response. The idea is that we can parse the content directly into `map[string]AuthConfig`, without the intermediate `map[string]json.RawMessage`. The original file store uses `map[string]json.RawMessage` because it has to persist the other fields so that it can write it back later. But for the read-only file store there is no such need so the code can be simpler.

registry/remote/credentials/internal/config/config.go
Wwwsylvia: With this the readonly store can do something like:

```
func (c *ReadOnlyConfig) GetCredential(serverAddress string) (auth.Credential, error) {
	authCfg, ok := matchAuth(c.auths, serverAddress)
	if !ok {
		return auth.EmptyCredential, nil
	}
	return authCfg.Credential()
}
```

This way we don't need to parse the raw message into AuthConfig for the read only store.

registry/remote/credentials/internal/config/readonly_config.go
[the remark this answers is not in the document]
programmer04: Thanks, @Wwwsylvia, for the review and guidance. I think now everything is implemented according to your suggestions. Please do a rereview

registry/remote/credentials/internal/config/helpers.go
```diff
@@ -0,0 +1,15 @@
+package config
```
Wwwsylvia, line 1: The ORAS copyright header is required for all files.

registry/remote/credentials/internal/config/helpers.go
Wwwsylvia, line 1: nit: maybe call this file `utils` to be consistent with other similar files in the code base?

registry/remote/credentials/file_store_test.go
```diff
@@ -179,6 +204,27 @@ func TestFileStore_Get_validConfig(t *testing.T) {
...
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("ReadOnlyFileStore.Get() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestReadOnlyFileStore_Create_fromInvalidConfig(t *testing.T) {
```
Wwwsylvia, line 220: We should put the tests for readonly file store in a separate test file `readonly_file_store_test.go` for better readability.

registry/remote/credentials/internal/config/readonly_config.go
Wwwsylvia, line 1: Can we have some unit tests for this file?

registry/remote/credentials/readonly_file_store.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Get always returns ErrReadOnlyStore. It's present to satisfy the Store interface.
+func (fs *ReadOnlyFileStore) Put(_ context.Context, _ string, _ auth.Credential) error {
+	return ErrReadOnlyStore
+}
+
+// Delete always returns ErrReadOnlyStore. It's present to satisfy the Store interface.
+func (fs *ReadOnlyFileStore) Delete(_ context.Context, _ string) error {
+	return ErrReadOnlyStore
+}
```
Wwwsylvia, line 62: For type safety, maybe we should introduce a new interface `ReadOnlyStore` that only contains the `Get` method.
```go
type ReadOnlyStore interface {
	// Get retrieves credentials from the store for the given server address.
	Get(ctx context.Context, serverAddress string) (auth.Credential, error)
}
```

We can also change the signature of `Credential()` from

https://github.com/oras-project/oras-go/blob/eeb21fc96e7b765d78bf94ac06edb737975af205/registry/remote/credentials/registry.go#L72

to
```go
func Credential(store ReadOnlyStore) auth.CredentialFunc {
```


@programmer04 Do you use the `Credential()` function in your code? 
@shizhMSFT What do you think of adding a new interface?

Wwwsylvia (changes requested):

registry/remote/credentials/readonly_file_store.go
programmer04: I think this is a great idea! I implemented this and tested it with my code - https://github.com/Kong/gateway-operator/pull/940 (there is a usage of `Credential()`) and it works nicely. 

Furthermore, I implemented all other suggestions. PTAL

Wwwsylvia (approved):
LGTM

registry/remote/credentials/registry.go
```diff
@@ -69,7 +69,7 @@ func Logout(ctx context.Context, store Store, registryName string) error {
 }
 
 // Credential returns a Credential() function that can be used by auth.Client.
-func Credential(store Store) auth.CredentialFunc {
+func Credential(store ReadOnlyStore) auth.CredentialFunc {
```
shizhMSFT, line 72: Technically, it's a breaking change since the function signature is changed.

registry/remote/credentials/readonly_file_store.go
```diff
@@ -0,0 +1,47 @@
...
+	"io"
+
+	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials/internal/config"
+)
+
+// ReadOnlyFileStore implements a credentials store using the docker configuration file
+// as an input. It supports only Get operation that works in the same way as for standard
+// FileStore.
+type ReadOnlyFileStore struct {
```
shizhMSFT, line 29: As you may notice, it is no longer a file store. Instead, it is a `ReadOnlyMemoryStore` now.

How about naming it `MemoryStore`?
```go
type MemoryStore struct {
    // fields are for demo only. it depends on the implementation
    auths sync.Map // map[string]AuthConfig
}

func NewMemoryStore() (*MemoryStore) {

}

func LoadMemoryStoreFromConfig(config []byte) (*MemoryStore, error) {

}

// Implement Get/Put/Delete
```

shizhMSFT (changes requested):

registry/remote/credentials/readonly_file_store.go
shizhMSFT: Ooops, we do have `NewMemoryStore`. All we need is an extra `LoadMemoryStoreFromConfig`.

registry/remote/credentials/registry.go
Wwwsylvia: The design wasn't good enough. We have to leave this signature change in v3 then.

registry/remote/credentials/readonly_file_store.go
Wwwsylvia: Okay, this would bring more extensibility - users can load credentials from config and update it in memory.
But since we are loading from config which could contain legacy auth keys, we need to normalize the keys when storing them.

How about something like:

```go
func LoadMemoryStoreFromConfig(config []byte) (Store, error) {
	var cfg struct {
		Auths map[string]configPkg.AuthConfig `json:"auths"`
	}
	if err := json.Unmarshal(config, &cfg); err != nil {
		return nil, fmt.Errorf("failed to decode config: %w", err)
	}

	s := &memoryStore{}
	for addr, auth := range cfg.Auths {
		// normalize the auth key to hostname
		hostname := configPkg.ToHostname(addr)
		cred, err := auth.Credential()
		if err != nil {
			return nil, err
		}
		s.store.Store(hostname, cred)
	}
	return s, nil
}
```

We need to expose the `toHostName` in the internal config package though.

programmer04: Thanks for the review. I updated the whole PR (description and code) according to @Wwwsylvia and @shizhMSFT  newest suggestions. PTAL

registry/remote/credentials/memory_store.go
```diff
@@ -17,23 +17,50 @@ package credentials
...
+	for addr, auth := range cfg.Auths {
+		// Normalize the auth key to hostname.
+		hostname := config.ToHostname(addr)
+		cred, err := auth.Credential()
+		if err != nil {
+			return nil, err
+		}
+		s.store.Store(hostname, cred)
+	}
+	return s, nil
```
shizhMSFT, line 59: It changes the behavior. We should also allow exact match.

shizhMSFT: Consider
```suggestion
	s := &MemoryStore{}
	for addr, auth := range cfg.Auths {
		cred, err := auth.Credential()
		if err != nil {
			return nil, err
		}
		s.store.Store(addr, cred)
		// normalize the auth key to hostname.
		_, _ = s.store.LoadOrStore(hostname, cred)
	}
	return s, nil
```

shizhMSFT: I use `LoadOrStore` to prevent overwriting.

registry/remote/credentials/memory_store.go
```diff
@@ -17,23 +17,50 @@ package credentials
...
 	"sync"
 
 	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials/internal/config"
 )
 
-// memoryStore is a store that keeps credentials in memory.
-type memoryStore struct {
+// MemoryStore is a store that keeps credentials in memory.
+type MemoryStore struct {
```
shizhMSFT, line 29: @Wwwsylvia What do you think about exposing the MemoryStore?

registry/remote/credentials/memory_store.go
```diff
@@ -17,23 +17,50 @@ package credentials
...
 // NewMemoryStore creates a new in-memory credentials store.
 func NewMemoryStore() Store {
-	return &memoryStore{}
+	return &MemoryStore{}
+}
+
+// NewMemoryStoreFromConfig creates a new in-memory credentials store from the given configuration.
+//
+// Reference: https://docs.docker.com/engine/reference/commandline/cli/#docker-cli-configuration-file-configjson-properties
+func NewMemoryStoreFromConfig(c []byte) (*MemoryStore, error) {
```
shizhMSFT, line 41: It is better to call out it is docker config.
```suggestion
func NewMemoryStoreFromDockerConfig(c []byte) (*MemoryStore, error) {
```

shizhMSFT (changes requested):

registry/remote/credentials/memory_store.go
shizhMSFT: /cc @Wwwsylvia for comments

Wwwsylvia: I realized that we can just normalize the host name, as the auth module only looks for the host name. Actually my memory was wrong, the original implementation did not do exact match.

registry/remote/credentials/memory_store.go
Wwwsylvia: I don't think exposing the type is necessary at this moment.

registry/remote/credentials/memory_store.go
shizhMSFT: Oh, OK. Let me close this comment.

registry/remote/credentials/memory_store.go
```diff
@@ -17,23 +17,50 @@ package credentials
...
+
+	s := &MemoryStore{}
+	for addr, auth := range cfg.Auths {
+		// Normalize the auth key to hostname.
+		hostname := config.ToHostname(addr)
+		cred, err := auth.Credential()
+		if err != nil {
+			return nil, err
+		}
+		s.store.Store(hostname, cred)
```
shizhMSFT, line 57: @Wwwsylvia To keep the original behavior, should we keep the first item instead of the last item if we have conflict normalized hostname?

```suggestion
		_, _ = s.store.LoadOrStore(hostname, cred)
```

Wwwsylvia: LGTM

programmer04:
Thanks, @Wwwsylvia for the great collaboration and review. It needs approval from @shizhMSFT too to merge it since he requested changes. And we're done

registry/remote/credentials/memory_store.go
```diff
@@ -32,6 +35,30 @@ func NewMemoryStore() Store {
 	return &memoryStore{}
 }
 
+// NewMemoryStoreFromDockerConfig creates a new in-memory credentials store from the given configuration.
+//
+// Reference: https://docs.docker.com/engine/reference/commandline/cli/#docker-cli-configuration-file-configjson-properties
+func NewMemoryStoreFromDockerConfig(c []byte) (*memoryStore, error) {
```
shizhMSFT, line 41: Public method should return public type
```suggestion
func NewMemoryStoreFromDockerConfig(c []byte) (Store, error) {
```

programmer04: Done

shizhMSFT (approved):
LGTM

programmer04:
Thanks @shizhMSFT for the great collaboration and review. It seems I can't merge it, someone with write permissions has to approve CI run for it and click the merge button

Wwwsylvia (approved):
LGTM

Wwwsylvia:
@programmer04 Merged. 🎉Thanks a lot for your efforts and the contribution!!

programmer04:
Thanks @Wwwsylvia! The last question is when do you release a new version? The last one is almost a year old, and looking at the `main` branch, it shows that many nice improvements have been made since then. In my opinion, it's high time to make a new release. WDYT?

Wwwsylvia:
@programmer04 I agree, I think we are pretty close to a new release. We will discuss it and hopefully we can make a release next week.

Wwwsylvia:
Hi @programmer04 , just a heads up -  We still have to finish all documentation work in the [v2.6.0 milestone](https://github.com/oras-project/oras-go/milestone/19). We have set the target date of the release to the end of February.

PR #852, opened 2024-12-02
feat: track progress

### LINKED ISSUE #839: Copy progress - feature request
As discussed in [this comment](https://github.com/oras-project/oras-go/pull/661#issuecomment-1871019328) (with thanks to @shizhMSFT for shepherding it through).

## Summary

* Request: Add support for ongoing progress updates when running `oras.Copy()`
* Context: if the downloaded blobs/manifests are large (and sometimes even if not), being able to know status updates, and that things are not stuck, is very useful
* Boundaries: this should not be concerned with UI or other aspects, except, potentially, for oras CLI, and that is out of scope at this time. The entire interface should be one that lets updates be sent by code to some consumer, which handles UI (if any)

## Proposed design

The signature for [oras.Copy](https://pkg.go.dev/oras.land/oras-go/v2#Copy) includes [oras.CopyOptions](https://pkg.go.dev/oras.land/oras-go/v2#CopyOptions) as the last parameter. If this were variadic, I would suggest adding another `WithProgressUpdate()` or similar, but we use a single `CopyOptions`, so I propose adding another property: `CopyOptions.Progress`. The type depends on the design choice.

There are two ways to do this:

* function: `CopyOptions.Progress func(Update)`. With each update, `Copy()` (or its underlying functions) would call the passed function, passing it the `Update` (see below).
* channel: `CopyOptions.Progress chan<- Update`. With each update `Copy()` would send an `Update` to the channel

If `Progress` is nil, then this functionality is bypassed.

Despite my linked PR using channels, I have no strong preference for channel over function.

### Preventing Blocking

With both functions and channels, there is a concern that it might block. In principle, I think that if someone calls `Copy()` and passes it a blocking function or unbuffered channel, that is their issue. However, it can cause us headaches to support them.

I would consider having each call that sends to the channel or calls `func()` to be in a separate short-lived goroutine, which calls fund or sends to the channel, wrapped in a timeout.

### Frequency of Update

My initial design is to rely on the underlying `io.CopyN()`. Whatever we use for that, we use for writing updates. However, that can overwhelm if the default `io.Copy()` is used. If I recall correctly, `io.Copy()` defaults to 32K. With a 100MB blob, that is ~3000 updates. That may or may not be good.

However we control the update frequency, I think it should be byte-based, not time-based. I.e. "updates every x KB" instead of "updates every y seconds." That is more useful, and also easier to implement.

In terms of controlling the update frequency, the simplest way is  `CopyOption.ProgresssFrequency uint`. If it is 0, stick to the default.

An alternative is to have `CopyOption.Progress` be a struct with both the channel/func (whichever is chosen) and an update frequency property.

A third method - and probably the simplest - is not to control it at all, but rather have it be part of `CopyOption.Progress`. Our call `Copy()` calls that / sends to channel, and it buffers as often as it wants. This is the simplest, but is subject to making our "blocking control", i.e. goroutines, being overwhelmed.

Open to ideas.

### Structure of update message

The `oras.Update` should be simple and contain only 2 properties:

```go
type Update struct {
   Copied int64
   Descriptor descriptor
}
```

The descriptor is important for 2 reasons:

1. To know the total expected size (which lets the consumer calculate the percentage complete)
2. To know what downloading blob/manifest this is for. The func can be called / channel can receive messages multiple times, even in parallel. This lets the consumer know exactly what the update is for.

## Sample

Channel:

```go
ch := make(chan oras.Update, 1000)
opts := oras.CopyOptions{
    ProgressChannel: ch        
}
desc, err := oras.Copy(ctx, src, tagName, dst, tagName, opts)
go func(ch <- chan CopyUpdate) {
    for msg := range ch {
        fmt.Printf("copied %d bytes out of %d total for %s\n", msg.Copied, msg.Descriptor.Size, msg.Descriptor.Digest)
    }
}(ch)
```

Func:

```go
f := func(msg oras.Update) {
        fmt.Printf("copied %d bytes out of %d total for %s\n", msg.Copied, msg.Descriptor.Size, msg.Descriptor.Digest)
}
opts := oras.CopyOptions{
    ProgressHandler: f
}
desc, err := oras.Copy(ctx, src, tagName, dst, tagName, opts)
```

shizhMSFT:
> [!IMPORTANT]
> This PR resolves #839 and is experimental.
> This PR is also a duplicate of the `internal/progress` package in https://github.com/oras-project/oras/pull/1524.

This PR introduce a new package `progress` to track the progress of reading content.


shizhMSFT:
Discussions should go to #839

deitch:
@shizhMSFT any progress on this?

shizhMSFT:
@deitch I have good news and bad news for this track progress effort.

The good news is that we have this package integrated in the `oras` CLI [v1.3.0-beta.2](https://github.com/oras-project/oras/releases/tag/v1.3.0-beta.2) (see [code](https://github.com/oras-project/oras/tree/v1.3.0-beta.2/internal/progress)). You can check the code and test it out.

The bad news is that the code / design in this PR does not match the quality for generic use in `oras-go`. There are several issues found when we did the integration:
1. The current progress model only covers the blob downloading operation and does not cover things like resolving the tags. Discussion is there in https://github.com/oras-project/oras-go/issues/839#issuecomment-2510765034
2. Degraded performance. If an `oras-go` user invokes the implementation of this PR directly, the performance will be roughly 10x slower due to too frequent progress event report and rendering. The `oras` CLI resolves this by introducing two strategies for different state reports. Precisely, transmitting events can be dropped if there is no CPU cycle for them (stateless like UDP) and other events will block until the render consumer picks it (stateful like TCP) so that critical events won't be missed (see [code](https://github.com/oras-project/oras/blob/v1.3.0-beta.2/cmd/oras/internal/display/status/progress/messenger.go)). However, which strategy should be used for an event (note: an event can be user defined) is purely application specific. Currently, I don't have a solution for making it generic.

Therefore, I'd like to abandon this PR and try to find a better design for the `progress` package in `oras-go` in the future.

Our overall strategy for having a new package in `oras-go` is that we first experiment in `oras` CLI and iterate it to be mature as well as validate the scenarios for real requirements. Once the new package is mature, we can promote it from the `oras` CLI to `oras-go`.

PR #867, opened 2025-01-08
fix: overwrite symlinks when extracting tarballs

### LINKED ISSUE #865: Bug: symbolic links are not automatically overwritten when extracted from tar archive to File store
Original issue in ORAS CLI repo: https://github.com/oras-project/oras/issues/1593

Currently, when extracting a tar archive in File store, we do not handle the case where a symbolic link contained in the archive already exists in the file system. When such a case happens, a LinkError is thrown by the `os.Symlink(target, path)` function call on line 190.

https://github.com/oras-project/oras-go/blob/e4dc64e9ffcdee2343dc0b4f6fff546c178f9da0/content/file/utils.go#L187-L191

Since we automatically overwrite other files by default, we should keep it consistent for symbolic links as well.

Wwwsylvia:
1. Fix the bug that symbolic links are not automatically overwritten when extracted from a tar archive to the File store.
2. Rename some functions and variables for better readability.
3. Add corresponding unit tests.
4. Improve docs and error messages.

Fixes #865


content/file/utils.go
```diff
@@ -61,6 +61,7 @@ func tarDirectory(ctx context.Context, root, prefix string, w io.Writer, removeT
 		name = filepath.ToSlash(name)
 
 		// Generate header
+		// NOTE: We don't support hard links and treat it as regular files
```
TerryHowe, line 64: If there wqs a duplicate blob in this with the same shasum, a hard link would be ideal. Is that a possibility with this method?

content/file/utils.go
```diff
@@ -180,15 +181,27 @@ func extractTarDirectory(dir, prefix string, r io.Reader, buf []byte) error {
 		case tar.TypeDir:
 			err = os.MkdirAll(path, header.FileInfo().Mode())
 		case tar.TypeLink:
+			// NOTE: ORAS does not generate hard links when creating tarballs.
+			// If a hard link is found in the tarball, it will be extracted.
+			// If the target link already exists, os.Link will throw an error.
+			// This is a known limitation and will not be addressed.
```
TerryHowe, line 187: If supporting links is possible and maybe desirable in the futre, maybe just make an issue out of this

content/file/utils.go
Wwwsylvia: Do you mean users may push a directory containing two files, one of them is a hark link to another?

content/file/utils.go
Wwwsylvia: Hard links in tarballs are tricky to support and can potentially introduce issues. For example, it requires exhaustive handling/tests on different OS. If there is a valid use case, we can evaluate the scenario and see if we need an issue for this.

content/file/utils.go
```diff
@@ -166,29 +167,41 @@ func extractTarDirectory(dir, prefix string, r io.Reader, buf []byte) error {
...
+			if err != nil {
+				return err
+			}
+			if _, err := os.Lstat(filePath); err == nil {
+				// link already exists, remove it first
+				if err := os.Remove(filePath); err != nil {
+					return err
+				}
 			}
+			err = os.Symlink(target, filePath)
```
shizhMSFT, line 204: We should assume the link does not exist. If `os.Symlink()` fails, we should then remove and retry.

shizhMSFT (changes requested):

content/file/utils.go
Wwwsylvia: This makes sense, updated.

shizhMSFT (approved):
LGTM

Issue #851, opened 2024-11-27
Oras does not read re-archived docker TAR files correctly


Atharex:
Details & original discussion thread here:
https://github.com/oras-project/oras/discussions/1539

Wwwsylvia:
Hi @Atharex, before we implement the enhancement, can you try the following tar command as a workaround?

Suppose you have a local OCI layout directory `repo` that is extracted from `repo.tar`. After making some changes on the directory `repo`, you can archive it back by running:

```shell
cd repo && rm -r ingest/ && tar cvf ../repo_modified.tar * && cd ..
```

Atharex:
Hi @Wwwsylvia, this workaround actually works!

After running tar within the folder, the archive is properly packaged and can be read with oras as normal

```
$ oras discover --oci-layout modified_local.tar:$TAG 
modified_local.tar@sha256:7e3b87ef0305b2734e28a5167505d67ee07ef7e9abd9b2a22a3af587a87047e0
└── my-favorite-artifact/type
    └── sha256:44a365ff03ad9dce2ae7898556446bd2e0708ab11bc68ec7b3ed08507bb9fb94

```

Wwwsylvia:
@Atharex Sounds good! Currently we have to package paths without the `./` prefix. The way to produce such tar files is a bit tricky though. 🫤
We will improve this in a future release!

PR #875, opened 2025-01-17
fix: handle dot-prefixed path for tarfs

Wwwsylvia:
1. Handle dot-prefixed paths such as `./index.json` when indexing and reading entries in tarfs
2. Add corresponding tests
3. Refactor original tests

Fixes: #851


internal/fs/tarfs/tarfs.go
```diff
@@ -143,15 +144,21 @@ func (tfs *TarFS) indexEntries() error {
...
 			pos:    pos - blockSize,
 		}
 	}
-
 	return nil
 }
 
+// getEntryKey cleans the name and returns a key for the entry.
+func getEntryKey(name string) string {
+	return filepath.Clean(name)
```
shizhMSFT: Should we use `path.Clean()` or `filepath.Clean()`?

shizhMSFT (approved):
LGTM

Issue #886, opened 2025-02-06
file permissions are not retained


sammy-da:
Files that are outside of directories don't get packed when `Add()`'ed to the file store and subesquently `Copy()`'ed, and their permissions (i.e. `rwx` file mode bits) don't get retained.

Also, for files within (sub-) directories, their mode bits don't get _fully_ retained when pulling either.
Seems like the un-packing is akin to `tar -x` rather than `tar -x --preserve-permissions`.
It would be nice to have an option to preserve the permissions

sammy-da:
Willing to contribute a PR.

sammy-da:
> Files that are outside of directories don't get packed when Add()'ed to the file store and subesquently Copy()'ed, and their permissions (i.e. rwx file mode bits) don't get retained.

for those, i came up with my own scheme for recording the file info in my app, which by setting annotations on the layer corrospoding to each file when pushing.

Wwwsylvia:
Hi @sammy-da, could you elaborate a bit more on this?

> Also, for files within (sub-) directories, their mode bits don't get fully retained when pulling either.

Can you share the reproduction steps?

It looks like `tar -x --preserve-permissions` is used to preserve the original file permissions exactly as they are, regardless of the current user's umask. Is this the behavior you are looking for in `oras-go`?

sammy-da:
> It looks like tar -x --preserve-permissions is used to preserve the original file permissions exactly as they are, regardless of the current user's umask. Is this the behavior you are looking for in oras-go?

yep, exactly

sammy-da:
> Can you share the reproduction steps?

```shell
umask 0022
mkdir foo
echo bar > foo/bar.txt
chmod +x 777 foo/bar.txt
oras push localhost:5000/foo:latest foo

oras pull localhost:5000/foo:latest -o out
ls -al out/foo/bar.txt 
# shows rwxr--r-- instead of expect `rwxrwxrwx`
```

Btw... I submitted https://github.com/oras-project/oras-go/pull/891 to address the "unpacked files perms" part of this issue.

Wwwsylvia:
Got it. Thanks @sammy-da , I've reviewed the PR and left some comments.

Wwwsylvia:
The PR is merged, thank you! @sammy-da Do you need this feature on the CLI side? If so, you may open an issue [there](https://github.com/oras-project/oras).

PR #891, opened 2025-02-14
feat: option to have File store preserve permissions when extracting

sammy-da:
This PR takes care of https://github.com/oras-project/oras-go/issues/886, or rather part of it.
It adds an option to the File store that's similar to tar's `--preserve-permissions`.

closes https://github.com/oras-project/oras-go/issues/886

### DIFF
--- a/content/file/errors.go
+++ b/content/file/errors.go
@@ -23,6 +23,7 @@ var (
 	ErrPathTraversalDisallowed = errors.New("path traversal disallowed")
 	ErrOverwriteDisallowed     = errors.New("overwrite disallowed")
 	ErrStoreClosed             = errors.New("store already closed")
+	ErrPreservePermissions     = errors.New("failed to restore permissions during unpacking")
 )
 
 var errSkipUnnamed = errors.New("unnamed descriptor skipped")
--- a/content/file/errors.go
+++ b/content/file/errors.go
@@ -23,7 +23,6 @@ var (
 	ErrPathTraversalDisallowed = errors.New("path traversal disallowed")
 	ErrOverwriteDisallowed     = errors.New("overwrite disallowed")
 	ErrStoreClosed             = errors.New("store already closed")
-	ErrPreservePermissions     = errors.New("failed to restore permissions during unpacking")
 )
 
 var errSkipUnnamed = errors.New("unnamed descriptor skipped")
--- a/content/file/file.go
+++ b/content/file/file.go
@@ -108,6 +108,9 @@ type Store struct {
 	// value overrides the [AnnotationUnpack].
 	// Default value: false.
 	SkipUnpack bool
+	// PreservePermissions controls whether to preserve file permissions when unpacking,
+	// disregarding the active umask, similar to tar's `--preserve-permissions`
+	PreservePermissions bool
 
 	workingDir   string   // the working directory of the file store
 	closed       int32    // if the store is closed - 0: false, 1: true.
@@ -499,7 +502,7 @@ func (s *Store) pushDir(name, target string, expected ocispec.Descriptor, conten
 	checksum := expected.Annotations[AnnotationDigest]
 	buf := bufPool.Get().(*[]byte)
 	defer bufPool.Put(buf)
-	if err := extractTarGzip(target, name, gzPath, checksum, *buf); err != nil {
+	if err := extractTarGzip(target, name, gzPath, checksum, *buf, s.PreservePermissions); err != nil {
 		return fmt.Errorf("failed to extract tar to %s: %w", target, err)
 	}
 	return nil
--- a/content/file/utils.go
+++ b/content/file/utils.go
@@ -219,7 +219,7 @@ func extractTarDirectory(dirPath, dirName string, r io.Reader, buf []byte, prese
 		// Restore full mode bits
 		if preservePermissions && (header.Typeflag == tar.TypeReg || header.Typeflag == tar.TypeDir) {
 			if err := os.Chmod(filePath, os.FileMode(header.Mode)); err != nil {
-				return fmt.Errorf("%w: %w", ErrPreservePermissions, err)
+				return err
 			}
 		}
 	}
--- a/content/file/utils.go
+++ b/content/file/utils.go
@@ -113,7 +113,7 @@ func tarDirectory(ctx context.Context, root, prefix string, w io.Writer, removeT
 
 // extractTarGzip decompresses the gzip
 // and extracts tar file to a directory specified by the `dir` parameter.
-func extractTarGzip(dirPath, dirName, gzPath, checksum string, buf []byte) (err error) {
+func extractTarGzip(dirPath, dirName, gzPath, checksum string, buf []byte, preservePermissions bool) (err error) {
 	fp, err := os.Open(gzPath)
 	if err != nil {
 		return err
@@ -144,7 +144,7 @@ func extractTarGzip(dirPath, dirName, gzPath, checksum string, buf []byte) (err 
 			r = io.TeeReader(r, verifier)
 		}
 	}
-	if err := extractTarDirectory(dirPath, dirName, r, buf); err != nil {
+	if err := extractTarDirectory(dirPath, dirName, r, buf, preservePermissions); err != nil {
 		return err
 	}
 	if verifier != nil && !verifier.Verified() {
@@ -156,7 +156,7 @@ func extractTarGzip(dirPath, dirName, gzPath, checksum string, buf []byte) (err 
 // extractTarDirectory extracts tar file to a directory specified by the `dir`
 // parameter. The file name prefix is ensured to be the string specified by the
 // `prefix` parameter and is trimmed.
-func extractTarDirectory(dirPath, dirName string, r io.Reader, buf []byte) error {
+func extractTarDirectory(dirPath, dirName string, r io.Reader, buf []byte, preservePermissions bool) error {
 	tr := tar.NewReader(r)
 	for {
 		header, err := tr.Next()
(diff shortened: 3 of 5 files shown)


content/file/utils_test.go
```diff
@@ -327,12 +344,33 @@ func Test_extractTarDirectory(t *testing.T) {
 					if string(gotContent) != wantContent {
 						t.Errorf("file content = %s, want %s", gotContent, wantContent)
 					}
+
+					header := headers[path]
```
TerryHowe, line 348: I'm not a maintainer, but this is the kind of thing that make table tests bad. This may in fact be the way to go, but the hardlink tests make a lot more sense to me.

content/file/file.go
```diff
@@ -108,6 +108,8 @@ type Store struct {
 	// value overrides the [AnnotationUnpack].
 	// Default value: false.
 	SkipUnpack bool
+	// PreserveModeBits controls whether to preserve file permissions when unpacking
+	PreserveModeBits bool
```
Wwwsylvia, line 112: nit: How about just calling it `PreservePermission` so users know that it's similar to the `tar -x --preserve-permission`?

content/file/utils.go
```diff
@@ -214,7 +214,12 @@ func extractTarDirectory(dirPath, dirName string, r io.Reader, buf []byte) error
 		}
 
 		// Change access time and modification time if possible (error ignored)
-		os.Chtimes(filePath, header.AccessTime, header.ModTime)
+		_ = os.Chtimes(filePath, header.AccessTime, header.ModTime)
+
+		// Restore full mode bits (error ignored)
+		if preserveModeBits && (header.Typeflag == tar.TypeReg || header.Typeflag == tar.TypeDir) {
+			_ = os.Chmod(filePath, os.FileMode(header.Mode))
```
Wwwsylvia, line 221: Since the user set the option, they may want to get an error if the permission cannot be kept. I think we can return a dedicated error type for users to catch.
What do you think?

content/file/utils_test.go
Wwwsylvia: +1 to Terry's point, can we move this test case to a separate test function to make it clearer?

content/file/file.go
```diff
@@ -108,6 +108,8 @@ type Store struct {
 	// value overrides the [AnnotationUnpack].
 	// Default value: false.
 	SkipUnpack bool
+	// PreserveModeBits controls whether to preserve file permissions when unpacking
```
Wwwsylvia, line 111: We may want to make it clearer and call out that umask will be ignored.

content/file/utils.go
sammy-da: Yea, I like that.
Fixed

content/file/utils_test.go
sammy-da: 👍 fixed.

content/file/file.go
sammy-da: 👍 fixed.

content/file/errors.go
```diff
@@ -23,6 +23,7 @@ var (
 	ErrPathTraversalDisallowed = errors.New("path traversal disallowed")
 	ErrOverwriteDisallowed     = errors.New("overwrite disallowed")
 	ErrStoreClosed             = errors.New("store already closed")
+	ErrPreservePermissions     = errors.New("failed to restore permission during unpacking")
```
Wwwsylvia, line 26: Since this error is file store specific, it should be put the `file` package.
BTW, should it be "restore permission**s**" in the error message?

content/file/file.go
```diff
@@ -108,6 +108,9 @@ type Store struct {
 	// value overrides the [AnnotationUnpack].
 	// Default value: false.
 	SkipUnpack bool
+	// PreservePermission controls whether to preserve file permissions when unpacking,
+	// disregarding the active umask, similar to tar's `--preserve-permissions`
+	PreservePermission bool
```
Wwwsylvia, line 113: nit: Do you mean PreservePermission**s**?

content/file/utils_test.go
```diff
@@ -372,12 +372,49 @@ func Test_extractTarDirectory_HardLink(t *testing.T) {
...
+		{name: "base/test.txt", content: "hello world", mode: 0771},
+		{name: "base/file_symlink", linkname: "test.txt", mode: os.ModeSymlink | 0771},
+	})
+
+	tempDir := t.TempDir()
+	dirName := "base"
+	dirPath := filepath.Join(tempDir, dirName)
+	fileContent := "hello world"
+	fileMode := os.FileMode(0771)
+	buf := make([]byte, 1024)
```
Wwwsylvia, line 393: nit: 

```suggestion

        fileContent := "hello world"
	fileMode := os.FileMode(0771)
	tarData := createTar(t, []tarEntry{
		{name: "base/", mode: os.ModeDir | 0777},
		{name: "base/test.txt", content: fileContent, mode: fileMode},
	})

	tempDir := t.TempDir()
	dirName := "base"
	dirPath := filepath.Join(tempDir, dirName)
	buf := make([]byte, 1024)
```

content/file/utils_test.go
```diff
@@ -372,12 +372,49 @@ func Test_extractTarDirectory_HardLink(t *testing.T) {
...
 		})
 
-		if err := extractTarDirectory(dirPath, dirName, bytes.NewReader(tarData), buf); err == nil {
+		if err := extractTarDirectory(dirPath, dirName, bytes.NewReader(tarData), buf, false); err == nil {
 			t.Error("extractTarDirectory() error = nil, wantErr = true")
 		}
 	})
 }
 
+func Test_extractTarDirectory_PreservePermissions(t *testing.T) {
```
Wwwsylvia, line 381: This test won't work in Windows... We need to move it to `utils_unix_test.go`.

Wwwsylvia (changes requested):

content/file/errors.go
sammy-da: this _is_ the file package.

>BTW, should it be "restore permissions" in the error message?

yep, typo

sammy-da: btw, not sure how to increase the code coverage for this.
To do that, i think i would need to test the chmod error case. It isn't obvious to me how to trigger such failure, other than concurrently trying to delete a file before its chmod'ed or something (i.e. a race, but that would be flaky test)

content/file/utils.go
```diff
@@ -113,7 +113,7 @@ func tarDirectory(ctx context.Context, root, prefix string, w io.Writer, removeT
 
 // extractTarGzip decompresses the gzip
 // and extracts tar file to a directory specified by the `dir` parameter.
-func extractTarGzip(dirPath, dirName, gzPath, checksum string, buf []byte) (err error) {
+func extractTarGzip(dirPath, dirName, gzPath, checksum string, buf []byte, preservePermission bool) (err error) {
```
Wwwsylvia, line 116: nit: preservePermission**s** ? Similar to other places.

Wwwsylvia (commented):
LGTM. Just a nit.

content/file/errors.go
Wwwsylvia: > this is the file package. (?)

Oh sorry I was blind😂
Yeah it would be very tricky to test the chmod error case... I think we can bypass it. /cc @shizhMSFT

Wwwsylvia:
@sammy-da Could you also [link](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) the PR to the issue using GitHub keywords?

sammy-da:
> Could you also [link](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) the PR to the issue using GitHub keywords?

It doesn't fully close it though. This will only take care of files that end up getting packed.
I guess i'll mark this as closing [the issue](https://github.com/oras-project/oras-go/issues/886) anyway 🤷

content/file/file_unix_test.go
```diff
@@ -476,3 +476,39 @@ func TestStore_Dir_OverwriteSymlink_RemovalFailed(t *testing.T) {
 		t.Fatal("error calling Chmod(), error =", err)
 	}
 }
+
+func Test_extractTarDirectory_PreservePermissions(t *testing.T) {
```
Wwwsylvia, line 480: Forgot to mention, since this function is defined in `utils.go`, it's better to test it in `utils_unix_test.go` instead of `file_unix_test.go`.
A side note: we need to include the oras copyright header at the beginning of each file.

shizhMSFT:
@sammy-da DCO checking is failing. Could you fix it?

content/file/utils.go
```diff
@@ -214,7 +214,14 @@ func extractTarDirectory(dirPath, dirName string, r io.Reader, buf []byte) error
 		}
 
 		// Change access time and modification time if possible (error ignored)
-		os.Chtimes(filePath, header.AccessTime, header.ModTime)
+		_ = os.Chtimes(filePath, header.AccessTime, header.ModTime)
+
+		// Restore full mode bits
+		if preservePermissions && (header.Typeflag == tar.TypeReg || header.Typeflag == tar.TypeDir) {
+			if err := os.Chmod(filePath, os.FileMode(header.Mode)); err != nil {
+				return fmt.Errorf("%w: %w", ErrPreservePermissions, err)
```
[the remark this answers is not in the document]
sammy-da: @Wwwsylvia Given that we're having problem even coming up with a way to trigger this error in a test, I think I agree with this. In hindsight i think this dedicated Err is overkill

content/file/errors.go
sammy-da: How do I bypass it?
do I ignore the entire file?

Wwwsylvia: We will ignore the patch CI failure and force merge...

Wwwsylvia (approved):
LGTM

Wwwsylvia:
@sammy-da Could you help to rebase?

sammy-da:
@Wwwsylvia rebased

shizhMSFT (approved):
LGTM

Issue #880, opened 2025-01-27
OCI Store generates old version of Index


shizhMSFT:
The OCI store generates `v1.0.0` version of the Index file with `mediaType` missing.

https://github.com/oras-project/oras-go/blob/bb3d17803e9670f688387f14739e0e94ab35299f/content/oci/oci.go#L378-L383

Although the context is clear, it is better to generate `v1.0.1` and above version of the index file. To be compliant, `oras-go` should generate `v1.1.0` version of the Index as it is backward compatible.

PR #894, opened 2025-02-18
fix: add media type in the generated index.json file

Wwwsylvia:
Add a media type field in the `index.json` file generated during the OCI store initialization.

Fixes: #880


shizhMSFT (approved):
LGTM

PR #909, opened 2025-03-04
chore: update to `image-spec v1.1.1`

Wwwsylvia:
1. Use the media type regex defined in [`image-spec v1.1.1`](https://github.com/opencontainers/image-spec/blob/v1.1.1/schema/defs-descriptor.json#L7) in `pack.go`
2. Update the doc links to reference `image-spec v1.1.1` and `distribution-spec v1.1.1`

Resolves: #910


shizhMSFT (approved):
LGTM

PR #933, opened 2025-03-31
feat: introduce `CopyError` to identify error source in copy operations

Wwwsylvia:
Resolves: 677

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -131,10 +131,10 @@ type CopyGraphOptions struct {
 // Returns the descriptor of the root node on successful copy.
 func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, dstRef string, opts CopyOptions) (ocispec.Descriptor, error) {
 	if src == nil {
-		return ocispec.Descriptor{}, NewCopyError("validation", CopyErrorOriginSource, errors.New("nil source target"))
+		return ocispec.Descriptor{}, newCopyError("validation", CopyErrorOriginSource, errors.New("nil source target"))
 	}
 	if dst == nil {
-		return ocispec.Descriptor{}, NewCopyError("validation", CopyErrorOriginDestination, errors.New("nil destination target"))
+		return ocispec.Descriptor{}, newCopyError("validation", CopyErrorOriginDestination, errors.New("nil destination target"))
 	}
 	if dstRef == "" {
 		dstRef = srcRef
@@ -147,14 +147,14 @@ func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, ds
 	proxy := cas.NewProxyWithLimit(src, cas.NewMemory(), opts.MaxMetadataBytes)
 	root, err := resolveRoot(ctx, src, srcRef, proxy)
 	if err != nil {
-		return ocispec.Descriptor{}, NewCopyError("resolveRoot", CopyErrorOriginSource, err)
+		return ocispec.Descriptor{}, newCopyError("resolveRoot", CopyErrorOriginSource, err)
 	}
 
 	if opts.MapRoot != nil {
 		proxy.StopCaching = true
 		root, err = opts.MapRoot(ctx, proxy, root)
 		if err != nil {
-			return ocispec.Descriptor{}, NewCopyError("MapRoot", CopyErrorOriginSource, err)
+			return ocispec.Descriptor{}, newCopyError("MapRoot", CopyErrorOriginSource, err)
 		}
 		proxy.StopCaching = false
 	}
@@ -222,7 +222,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 		// skip if a rooted sub-DAG exists
 		exists, err := dst.Exists(ctx, desc)
 		if err != nil {
-			return NewCopyError("Exists", CopyErrorOriginDestination, err)
+			return newCopyError("Exists", CopyErrorOriginDestination, err)
 		}
 		if exists {
 			if opts.OnCopySkipped != nil {
@@ -236,7 +236,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 		// find successors while non-leaf nodes will be fetched and cached
 		successors, err := opts.FindSuccessors(ctx, proxy, desc)
 		if err != nil {
-			return NewCopyError("FindSuccessors", CopyErrorOriginSource, err)
+			return newCopyError("FindSuccessors", CopyErrorOriginSource, err)
 		}
 		successors = removeForeignLayers(successors)
 
@@ -250,12 +250,12 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 				done, committed := tracker.TryCommit(node)
 				if committed {
 					err := fmt.Errorf("%s: %s: successor not committed", desc.Digest, node.Digest)
-					return NewCopyError("copyGraph", CopyErrorOriginUnknown, err)
+					return newCopyError("copyGraph", CopyErrorOriginInternal, err)
 				}
 				select {
 				case <-done:
 				case <-ctx.Done():
-					NewCopyError("copyGraph", CopyErrorOriginUnknown, ctx.Err())
+					newCopyError("copyGraph", CopyErrorOriginInternal, ctx.Err())
 				}
 			}
 			if err := region.Start(); err != nil {
@@ -265,7 +265,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 
 		exists, err = proxy.Cache.Exists(ctx, desc)
 		if err != nil {
-			return NewCopyError("Cache.Exists", CopyErrorOriginSource, err)
+			return newCopyError("Cache.Exists", CopyErrorOriginSource, err)
 		}
 		if exists {
 			return copyNode(ctx, proxy.Cache, dst, desc, opts)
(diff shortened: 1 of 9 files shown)


copyerror.go
shizhMSFT, line 1: This file should be merged into `copy.go`.

shizhMSFT: Same for other related files.

copyerror.go
```diff
@@ -0,0 +1,61 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package oras
+
+import "fmt"
+
+// CopyErrorOrigin defines the source of a copy error.
+type CopyErrorOrigin string
```
shizhMSFT, line 21: Should it just be an int?

shizhMSFT: or simply remove this type?

copy.go
```diff
@@ -319,12 +320,16 @@ func mountOrCopyNode(ctx context.Context, src content.ReadOnlyStorage, dst conte
 					return nil, err
 				}
 			}
-			return src.Fetch(ctx, desc)
+			rc, err := src.Fetch(ctx, desc)
+			if err != nil {
+				return nil, err
+			}
+			return rc, nil
```
shizhMSFT, line 327: This change seems unnecessary?

copy.go
```diff
@@ -264,7 +265,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 
 		exists, err = proxy.Cache.Exists(ctx, desc)
 		if err != nil {
-			return err
+			return newCopyError("Cache.Exists", CopyErrorOriginSource, err)
```
shizhMSFT, line 268: This operation seems not right.

copy.go
```diff
@@ -249,12 +249,13 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
...
 				if committed {
-					return fmt.Errorf("%s: %s: successor not committed", desc.Digest, node.Digest)
+					err := fmt.Errorf("%s: %s: successor not committed", desc.Digest, node.Digest)
+					return newCopyError("copyGraph", CopyErrorOriginInternal, err)
 				}
 				select {
 				case <-done:
 				case <-ctx.Done():
-					return ctx.Err()
+					return newCopyError("copyGraph", CopyErrorOriginInternal, ctx.Err())
```
shizhMSFT, line 258: "copyGraph" is internal. It is not that meaningful to end users.

copy.go
```diff
@@ -147,14 +147,14 @@ func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, ds
 	proxy := cas.NewProxyWithLimit(src, cas.NewMemory(), opts.MaxMetadataBytes)
 	root, err := resolveRoot(ctx, src, srcRef, proxy)
 	if err != nil {
-		return ocispec.Descriptor{}, fmt.Errorf("failed to resolve %s: %w", srcRef, err)
+		return ocispec.Descriptor{}, newCopyError("resolveRoot", CopyErrorOriginSource, err)
```
shizhMSFT, line 150: "resolveRoot" is internal. It is not that meaningful to end users.

copy.go
```diff
@@ -131,10 +131,10 @@ type CopyGraphOptions struct {
 // Returns the descriptor of the root node on successful copy.
 func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, dstRef string, opts CopyOptions) (ocispec.Descriptor, error) {
 	if src == nil {
-		return ocispec.Descriptor{}, errors.New("nil source target")
+		return ocispec.Descriptor{}, newCopyError("validation", CopyErrorOriginSource, errors.New("nil source target"))
 	}
 	if dst == nil {
-		return ocispec.Descriptor{}, errors.New("nil destination target")
+		return ocispec.Descriptor{}, newCopyError("validation", CopyErrorOriginDestination, errors.New("nil destination target"))
```
shizhMSFT, line 137: "validation" is internal. It is not that meaningful to end users.

copyerror.go
```diff
@@ -0,0 +1,61 @@
...
+	// CopyErrorOriginDestination indicates the error occurred at the destination side.
+	CopyErrorOriginDestination CopyErrorOrigin = "destination"
+
+	// CopyErrorOriginInternal indicates the error occurred internally.
+	CopyErrorOriginInternal CopyErrorOrigin = "internal"
+)
+
+// CopyError represents an error encountered during a copy operation.
+type CopyError struct {
+	Op     string
```
wangxiaoxuan273, line 36: It looks like most `Op` values are not very meaningful to end users 😂

copyerror.go
Wwwsylvia: I still prefer the current implementation, because:

1. A dedicated type ensures type safety and limits the possible values
2. The `string` type is more convenient for formatting the error message (see the `Error()` function)

copyerror.go
Wwwsylvia: I refactored the code a little bit, and now it should be able to help users to get a basic idea on where the error occurred.

copy.go
Wwwsylvia: refactored

copy.go
Wwwsylvia: refactored

copy.go
Wwwsylvia: refactored

copy.go
Wwwsylvia: refactored

copy.go
Wwwsylvia: Good catch, this was left from the previous iteration...

copyerror.go
Wwwsylvia: I found it clearer to keep it in a separate file, as there is 60+ lines of code and defines not only struct but methods. Besides, we need a separate test file `copyerror_test.go` so that we can make the tests under the `oras` package instead of `oras_test` package and test some internal methods (`newCopyError`).

copy.go
```diff
@@ -264,7 +270,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 
 		exists, err = proxy.Cache.Exists(ctx, desc)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to check cache existence: %s: %w", desc.Digest, err)
```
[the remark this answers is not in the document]
Wwwsylvia: This is cache error and not a source error

example_copyerror_test.go
```diff
@@ -0,0 +1,58 @@
...
+	dst := memory.New()
+	ctx := context.Background()
+
+	// Try to copy a non-existent reference
+	nonExistentRef := "non-existent-reference"
+	_, err := oras.Copy(ctx, src, nonExistentRef, dst, "", oras.DefaultCopyOptions)
+	if err == nil {
+		fmt.Println("copy succeeded")
+		return
+	}
```
shizhMSFT, line 39: Let's just panic on `err == nil` or we should still follow the pattern of `if err != nil {` since it is an example code and folks may just copy and paste.

copyerror.go
shizhMSFT: Since this is library code, I would prefer an `int`-based enum with a `String()` method. It would be less error prone than `string`-based as you can have something like
```go
switch e {
case CopyErrorOriginUnknown:
  return "unknown"
case CopyErrorOriginSource:
  return "source"
case CopyErrorOriginDestination:
  return "destination"
default:
  return "invalid"
} 
```
Also, the perf is better than `string`-based error.

copyerror.go
```diff
@@ -0,0 +1,61 @@
...
+	return &CopyError{
+		Op:     op,
+		Origin: origin,
+		Err:    err,
+	}
+}
+
+// Error implements the error interface for CopyError.
+func (e *CopyError) Error() string {
+	return fmt.Sprintf("%s error: failed to perform %q: %v", e.Origin, e.Op, e.Err)
```
shizhMSFT, line 55: The word `error` seems to be redundant.
```suggestion
	return fmt.Sprintf("%s: failed to perform %q: %v", e.Origin, e.Op, e.Err)
```

copyerror.go
qweeah: Caller won't need to know the real type of an error origin, e.g.
```golang
if err = oras.CopyGraph(ctx, src, dst, desc, extendedCopyOptions.CopyGraphOptions); err != nil {
  var re *oras.CopyError
  if !errors.As(err, &re){
    switch re.Code {
      case oras.CopyErrorOriginSource:
        return ocispec.Descriptor{}, fmt.Errorf("failed to copy from %s: %w", opts.From.Reference, err)
      case oras.CopyErrorOriginDestination:
        return ocispec.Descriptor{}, fmt.Errorf("failed to copy to %s: %w", opts.To.Reference, err)
    }
  }
}

But int performs better.

File: example_copyerror_test.go (hunk 9 above)
Comment (Wwwsylvia) [reply example_copyerror_test.go]: Updated

File: copyerror.go (hunk 1 above)
Comment (Wwwsylvia) [reply copyerror.go]: Refactored

File: copyerror.go (hunk 10 above)
Comment (Wwwsylvia) [reply copyerror.go]: Updated

File: copy.go [hunk 11]
```diff
@@ -434,7 +444,7 @@ func resolveRoot(ctx context.Context, src ReadOnlyTarget, srcRef string, proxy *
 }
 
 // prepareCopy prepares the hooks for copy.
-func prepareCopy(ctx context.Context, dst Target, dstRef string, proxy *cas.Proxy, root ocispec.Descriptor, opts *CopyOptions) error {
+func prepareCopy(_ context.Context, dst Target, dstRef string, proxy *cas.Proxy, root ocispec.Descriptor, opts *CopyOptions) error {
```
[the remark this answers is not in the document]
Wwwsylvia: `dst.Tag(ctx, ...)` is not from `prepareCopy`'s context

example_copyerror_test.go
```diff
@@ -0,0 +1,58 @@
...
+		var copyErr *oras.CopyError
+		if errors.As(err, &copyErr) {
+			fmt.Println("copyErr.Origin:", copyErr.Origin)
+			fmt.Println("copyErr.Op:", copyErr.Op)
+			fmt.Println("copyErr.Err:", copyErr.Err)
+			fmt.Println("copyErr.Error():", copyErr.Error())
+		} else {
+			fmt.Println("err is not a CopyError:", err)
+		}
+		return
```
shizhMSFT, line 48: nit:
```suggestion
			return
		}

		fmt.Println("err is not a CopyError:", err)
		return
```

shizhMSFT (approved):
LGTM with suggestions

example_copyerror_test.go
Wwwsylvia: Updated

wangxiaoxuan273 (approved):
LGTM

shizhMSFT (approved):
LGTM

PR #959, opened 2025-05-25
fix: Support empty credentials file

TerryHowe:
Updating helm from oras-go v1 to v2 and I discovered oras-go v2 does not support an empty credentials file.

There was a test file empty.json that was just not being used. It seems like a reasonable use case. With a Docker config.json, I've always just deleted everything in the file to clear my creds.


registry/remote/credentials/internal/config/config.go
```diff
@@ -128,7 +129,16 @@ func Load(configPath string) (*Config, error) {
...
+			return nil, fmt.Errorf("failed to decode config file %s: %w: %v", configPath, ErrInvalidConfigFormat, err)
+		}
+
+		// Check to see if the credential file was an empty file
+		fileInfo, statErr := configFile.Stat()
+		if statErr != nil || fileInfo.Size() != 0 {
+			// Parse error
+			return nil, fmt.Errorf("failed to decode config file %s: %w: %v", configPath, ErrInvalidConfigFormat, err)
+		}
 	}
```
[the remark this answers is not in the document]
TerryHowe: Done, thanks.

registry/remote/credentials/internal/config/config_test.go
```diff
@@ -346,6 +346,47 @@ func TestConfig_GetCredential_emptyConfig(t *testing.T) {
...
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Config.GetCredential() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestConfig_GetCredential_whiteSpaceFile(t *testing.T) {
```
Wwwsylvia, line 382: If we allow whitespaces, we can merge this into the above test and rename it to "emptyOrWhitespace"

registry/remote/credentials/testdata/whitespace.json
```diff
@@ -0,0 +1 @@
+
```
Wwwsylvia, line 1: nit: We should put some white spaces here if we allow it

registry/remote/credentials/internal/config/config.go
```diff
@@ -128,7 +129,21 @@ func Load(configPath string) (*Config, error) {
...
+		}
+
+		// Check to see if the credential file was an empty file
+		fileInfo, statErr := configFile.Stat()
+		if statErr != nil || fileInfo.Size() != 0 {
+			// Parse error
+			return nil, fmt.Errorf("failed to decode config file %s: %w: %v", configPath, ErrInvalidConfigFormat, err)
+		}
+
+		// init content and caches if the content file does not exist
```
Wwwsylvia, line 143: nit: Do you mean
```suggestion
		// init content and caches if the content file is empty
```

registry/remote/credentials/internal/config/config_test.go
```diff
@@ -346,6 +346,47 @@ func TestConfig_GetCredential_emptyConfig(t *testing.T) {
 	}
 }
 
+func TestConfig_GetCredential_emptyFile(t *testing.T) {
```
Wwwsylvia, line 349: Just realized that the above test `TestConfig_GetCredential_emptyConfig` already does the same thing, but the referenced file name is wrong, oops. (Should be `empty.json` instead of `empty_config.json`)

registry/remote/credentials/internal/config/config.go
```diff
@@ -128,7 +129,16 @@ func Load(configPath string) (*Config, error) {
 
 	// decode config content if the config file exists
 	if err := json.NewDecoder(configFile).Decode(&cfg.content); err != nil {
-		return nil, fmt.Errorf("failed to decode config file at %s: %w: %v", configPath, ErrInvalidConfigFormat, err)
+		if !errors.Is(err, io.EOF) {
+			return nil, fmt.Errorf("failed to decode config file %s: %w: %v", configPath, ErrInvalidConfigFormat, err)
+		}
+
+		// Check to see if the credential file was an empty file
+		fileInfo, statErr := configFile.Stat()
```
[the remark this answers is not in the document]
TerryHowe: Done.

registry/remote/credentials/internal/config/config.go
TerryHowe: Done

registry/remote/credentials/internal/config/config_test.go
TerryHowe: One test reads one file is the pattern here 🤷

registry/remote/credentials/testdata/whitespace.json
TerryHowe: There is whitespace in there

registry/remote/credentials/internal/config/config.go
```diff
@@ -128,7 +129,23 @@ func Load(configPath string) (*Config, error) {
...
+		if statErr != nil || fileInfo.Size() != 0 {
+			fileContent, readErr := os.ReadFile(configPath)
+			if readErr != nil || strings.TrimSpace(string(fileContent)) != "" {
+				return nil, fmt.Errorf("failed to decode config file %s: %w: %v", configPath, ErrInvalidConfigFormat, err)
```
TerryHowe, line 141: The only case that `Decode` returns io.EOF is all whitespace, so this code looks good, but is not testable. In the case of a file containing `{` for example, it will return "Unexpected EOF" as reproduced in the test below. Keep this code or get rid of it?

Wwwsylvia: > In the case of a file containing { for example, it will return "Unexpected EOF" as reproduced in the test below.

A file containing `{` should be considered corrupted.
The implementation in #962 is actually cleaner and can pass all tests.

registry/remote/credentials/internal/config/config_test.go
```diff
@@ -346,6 +346,47 @@ func TestConfig_GetCredential_emptyConfig(t *testing.T) {
...
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Config.GetCredential() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestConfig_GetCredential_noCloseCurly(t *testing.T) {
```
Wwwsylvia, line 382: nit: Looks like this can be merged into `TestLoad_badFormat`

registry/remote/credentials/internal/config/config.go
TerryHowe: There is no need to stat the file because empty or just whitespace are the only cases that return `io.EOF`. If there are any other characters in the file it returns `io.ErrUnexpectedEOF`

PR #962, opened 2025-05-27
Handle empty config files

### LINKED ISSUE #961: Load config file when it's an empty file
When the credentials config file is an empty file (i.e. "") there is an error and it's not loaded.

This is different from v1 of the package that handled this similar to the way v2 handles a missing file. When there is a missing file, default config is initialized and used.

mattfarina:
ORAS handles config files that do not exist by initializing the cfg object. A similar case is when config files are empty files. ORAS v1 used the docker CLI packages that handled the io.EOF error case gracefully by initializing the configuration.

This change adds in fault tolerance handling by handling the EOF case.

Closes #961

### DIFF
--- a/registry/remote/credentials/internal/config/config.go
+++ b/registry/remote/credentials/internal/config/config.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


mattfarina:
@FeynmanZhou it looks like they try to handle the same issue. I didn't notice Terrys PR when I created this one.

PR #966, opened 2025-05-30
fix: credentials key for docker registry-1

TerryHowe:
Login and logout should be supported for the OCI registry server `registry-1.docker.io` as well as the alias `docker.io` The key used is the legacy server for backward compatibility.

Related: https://github.com/helm/helm/issues/30929


TerryHowe:
> The changes LGTM. Can we add a few more test cases in `registry_test.go`?

Done

Wwwsylvia (approved):
LGTM

Issue #972, opened 2025-06-06
fetchBearerToken reverts too easily to basic authentication


TerryHowe:
The fetchBearerToken method seems to revert to basic authentication too easily. With docker.io for instance, they claim to only support bearer auth in their 'WWW-Authenticate', but they do not support RefresherToken as a password or personal access token. You can use bearer authentication with as pasword with either a password or a personal access token though.

https://github.com/oras-project/oras-go/blob/main/registry/remote/auth/client.go#L305

I am getting reports of other registries that request bearer via the WWW-Authenticate and auth fails with ORAS because it reverts to basic auth which is not supported.

TerryHowe:
Related https://github.com/helm/helm/issues/30970

Wwwsylvia:
> I am getting reports of other registries that request bearer via the WWW-Authenticate and auth fails with ORAS because it reverts to basic auth which is not supported.

Does it help to set `ForceAttemptOAuth2` to `true`?

Wwwsylvia:
> From a user perspective, it makes a lot more sense to have a `ForceBasicAuth` for bearer auth rather than a variable to actually use bearer auth for bearer auth.

@TerryHowe We should avoid changing the default behavior as it would be breaking. We can consider introducing an option for the new behavior if needed, maybe something like `PreferOAuth2`? 🤔

Anyway, before making any changes, we need to identify the requirements and expected behavior here.

TerryHowe:
I think there are two options here, we could do as you suggest and create something like `PreferOAuth2` (although I don't love that name) or cut a major release for the breaking change.

TerryHowe:
The reason I'm kind of in favor or new major release is it just isn't right to favor basic auth when the registry requests bearer.

Wwwsylvia:
@TerryHowe We do have a plan for v3 which may include significant refactoring of the auth client. But we don't think it is the right moment to cut a v3 release just for this feature.

If the proposal is to attempt OAuth2 whenever credentials are provided (#973), that’s effectively the same as setting `ForceAttemptOAuth2 = true`.

One option to consider is updating `DefaultClient` to have `ForceAttemptOAuth2` enabled by default. Since this change isn’t considered breaking, it could be included in the next minor release `v2.7.0`.

/cc @shizhMSFT for comments on planning and the security concern

shizhMSFT:
[...]

PR #89, opened 2022-01-05
WIP: Backport of tag listing to v1

sabre1041:
Backport of v2 registry tag listing to support Helm as outlined in [HIP-6](https://github.com/helm/community/blob/main/hips/hip-0006.md)

Also includes changes proposed in #88 

CC: @scottrigby @deitch @jdolitsky @scottrigby @SteveLasker

PR #973, opened 2025-06-07
fix: allow user password for bearer auth

TerryHowe:
Allow user/pass for bearer authentication

Closes: https://github.com/oras-project/oras-go/issues/972

### DIFF
--- a/registry/remote/auth/client.go
+++ b/registry/remote/auth/client.go
@@ -302,7 +302,7 @@ func (c *Client) fetchBearerToken(ctx context.Context, registry, realm, service 
 	if cred.AccessToken != "" {
 		return cred.AccessToken, nil
 	}
-	if cred == EmptyCredential || (cred.RefreshToken == "" && !c.ForceAttemptOAuth2) {
+	if cred == EmptyCredential {
 		return c.fetchDistributionToken(ctx, realm, service, scopes, cred.Username, cred.Password)
 	}
 	return c.fetchOAuth2Token(ctx, realm, service, scopes, cred)
--- a/registry/remote/auth/client_test.go
+++ b/registry/remote/auth/client_test.go
@@ -661,24 +661,44 @@ func TestClient_Do_Bearer_Auth(t *testing.T) {
 		"repository:src:pull",
 	}
 	as := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.Method != http.MethodGet || r.URL.Path != "/" {
+		if r.Method != http.MethodPost || r.URL.Path != "/" {
 			t.Error("unexecuted attempt of authorization service")
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
-		header := "Basic " + base64.StdEncoding.EncodeToString([]byte(username+":"+password))
-		if auth := r.Header.Get("Authorization"); auth != header {
-			t.Errorf("unexpected auth: got %s, want %s", auth, header)
+		if err := r.ParseForm(); err != nil {
+			t.Errorf("failed to parse form: %v", err)
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
-		if got := r.URL.Query().Get("service"); got != service {
-			t.Errorf("unexpected service: got %s, want %s", got, service)
+		if got := r.PostForm.Get("grant_type"); got != "password" {
+			t.Errorf("unexpected grant type: %v, want %v", got, "password")
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
-		if got := r.URL.Query()["scope"]; !reflect.DeepEqual(got, scopes) {
-			t.Errorf("unexpected scope: got %s, want %s", got, scopes)
+		if got := r.PostForm.Get("service"); got != service {
+			t.Errorf("unexpected service: %v, want %v", got, service)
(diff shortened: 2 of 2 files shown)


registry/remote/auth/client.go
```diff
@@ -302,7 +302,7 @@ func (c *Client) fetchBearerToken(ctx context.Context, registry, realm, service
 	if cred.AccessToken != "" {
 		return cred.AccessToken, nil
 	}
-	if cred == EmptyCredential || (cred.RefreshToken == "" && !c.ForceAttemptOAuth2) {
+	if cred == EmptyCredential {
```
gjenkins8: `ForceAttemptOAuth2` will be unused I think after this change. Which seems unusual, as presumably it exists for a reason?

Perhaps:
```suggestion
	if cred == EmptyCredential && !c.ForceAttemptOAuth2 {
```

(And to be clear, I'm speculating that if the user has set `c.ForceAttemptOAuth2 = true`, the user would expect the `c.fetchOAuth2Token(...)` to be invoked. Based on the naming only.)

TerryHowe: That or mark it deprecated

gjenkins8: certainly, if `ForceAttemptOAuth2` will serve no function now (my proposal doesn't make sense), then should mark it deprecated

sabre1041: Agree that with this it would be good to mark `ForceAttemptOAuth2` as deprecated

Wwwsylvia: In the Helm scenario, if the OAuth2 flow is preferred, can we just set `ForceAttemptOAuth2` to true? 🤔 This way:

- If the credential is empty, use basic auth.
- If not, use bearer auth.

TerryHowe: I think from a security perspective alone, we should favor bearer auth. The current implementation favors basic auth.

Setting `ForceAttemptOAuth2` also makes it appear we aren't going to support basic auth if needed.

Why would we not allow by default grant type `password` using username/password?

Wwwsylvia: @TerryHowe  The current implementation was designed to align with the [Moby implmentation](https://github.com/opencontainers/wg-auth/blob/main/docs/implementations/moby.md).  /cc @shizhMSFT Correct me if I'm wrong.
There is a concern that changing this behavior could be considered a breaking change.

Looking at the original [helm issue](https://github.com/helm/helm/issues/30970), the regression was introduced in helm v3.18.x. But Helm v3.17.3 [used](https://github.com/helm/helm/blob/e4da49785aa6e6ee2b86efd5dd9e43400318262b/pkg/registry/client.go#L144-L145) oras-go v1.2.5, which shared the same auth logic as v2. Why did upgrading to v2 break Helm's auth flow? Did I miss anything? 

oras-go v1.2.5 auth logic:
https://github.com/oras-project/oras-go/blob/9d9e2cb1efbad77745b57601947a3b537313ba7f/pkg/registry/remote/auth/client.go#L258-L261

TerryHowe: Huh, when I was running it, it was in some containerd code. I'll need to look at that again.

Wwwsylvia: @TerryHowe This PR is becoming stale. Did you find anything suspicious?

TerryHowe: v1 has a backport of the same logic https://github.com/oras-project/oras-go/pull/89 but helm used the original https://github.com/oras-project/oras-go/tree/v1/pkg/auth it uses containerd which is what I was seeing https://github.com/oras-project/oras-go/blob/v1/pkg/auth/client.go#L23

TerryHowe:
This should work like https://github.com/helm/helm/pull/31212

Wwwsylvia (commented):
Is this PR targeting `v2`?

registry/remote/auth/client.go
```diff
@@ -164,6 +166,11 @@ func (c *Client) SetUserAgent(userAgent string) {
 	c.Header.Set("User-Agent", userAgent)
 }
 
+// SetLegacyMode favors basic auth even when the registry requests bearer.
+func (c *Client) SetLegacyMode() {
+	c.ForceAttemptOAuth2 = false
```
Wwwsylvia, line 171: Wondering what's the scenario of this? This field is defaulted to be `false`.🤔

registry/remote/auth/client.go
```diff
@@ -119,6 +119,8 @@ type Client struct {
 	// References:
 	// - https://distribution.github.io/distribution/spec/auth/jwt/
 	// - https://distribution.github.io/distribution/spec/auth/oauth/
+	// Deprecated: This will be removed in the future.
+	// Calling SetLegacyMode() will preserve ForceAttemptOAuth2 set to false
```
Wwwsylvia, line 123: Do we really want to deprecate this when there's no substitute solution?

TerryHowe: This deprecation notice is just to let people know if they are using this variable that it is going away.

registry/remote/auth/client.go
TerryHowe: This is a method for people to use in the future to preserve the old behavior (favor basic auth).

registry/remote/auth/client.go
TerryHowe: I thought this was what was wanted, but I'll create another PR that is a breaking change

PR #1004, opened 2025-09-12
feature: make referrers API const public

TerryHowe:
Exposing this makes things easier for the user.


Wwwsylvia (commented):
Why do users need to access these constants? We don't expose referrersState.

TerryHowe:
> Why do users need to access these constants? We don't expose referrersState.

These are the three possible states and if the ORAS CLI had this, it wouldn't have to replicate this structure.

PR #1015, opened 2025-10-06
feature: token expiration tracking

TerryHowe:
Track token expiration so refresh can happen on time.

### DIFF
--- a/registry/remote/auth/cache.go
+++ b/registry/remote/auth/cache.go
@@ -56,7 +56,7 @@ type Cache interface {
 // cacheEntry is a cache entry for a single registry.
 type cacheEntry struct {
 	scheme Scheme
-	tokens sync.Map // map[string]string
+	tokens sync.Map // map[string]*tokenEntry
 }
 
 // concurrentCache is a cache suitable for concurrent invocation.
@@ -80,7 +80,8 @@ func (cc *concurrentCache) GetScheme(ctx context.Context, registry string) (Sche
 }
 
 // GetToken returns the auth-token part cached for the given registry of a given
-// scheme.
+// scheme. It automatically checks token expiration and removes expired tokens
+// from the cache.
 func (cc *concurrentCache) GetToken(ctx context.Context, registry string, scheme Scheme, key string) (string, error) {
 	entryValue, ok := cc.cache.Load(registry)
 	if !ok {
@@ -90,8 +91,14 @@ func (cc *concurrentCache) GetToken(ctx context.Context, registry string, scheme
 	if entry.scheme != scheme {
 		return "", errdef.ErrNotFound
 	}
-	if token, ok := entry.tokens.Load(key); ok {
-		return token.(string), nil
+	if tokenValue, ok := entry.tokens.Load(key); ok {
+		te := tokenValue.(*tokenEntry)
+		if te.isExpired() {
+			// Token has expired, remove from cache
+			entry.tokens.Delete(key)
+			return "", errdef.ErrNotFound
+		}
+		return te.token, nil
 	}
 	return "", errdef.ErrNotFound
 }
@@ -123,7 +130,7 @@ func (cc *concurrentCache) Set(ctx context.Context, registry string, scheme Sche
 		return token, nil
 	}
 
-	// cache token
+	// cache token with expiration
 	newEntry := &cacheEntry{
 		scheme: scheme,
 	}
@@ -135,7 +142,7 @@ func (cc *concurrentCache) Set(ctx context.Context, registry string, scheme Sche
 		entry = newEntry
 		cc.cache.Store(registry, entry)
 	}
-	entry.tokens.Store(key, token)
+	entry.tokens.Store(key, newTokenEntry(token))
 
 	return token, nil
 }
--- a/registry/remote/auth/token.go
+++ b/registry/remote/auth/token.go
@@ -1,0 +1,100 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package auth
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+	"time"
+)
+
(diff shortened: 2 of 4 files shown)


registry/remote/auth/token.go
```diff
@@ -0,0 +1,100 @@
...
+// tokens that don't contain expiration information (like basic auth tokens).
+// For JWT tokens without an expiration claim, it returns a default expiration
+// of 60 seconds from now.
+func parseTokenExpiration(token string) time.Time {
+	// Try to parse as JWT token
+	// JWT format: header.payload.signature
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		// Not a JWT token, return default expiration
+		return time.Now().Add(60 * time.Second)
```
Wwwsylvia: We should make the default expiration a constant

registry/remote/auth/token.go
```diff
@@ -0,0 +1,100 @@
...
+// isExpired checks if a token entry has expired.
+// It includes a 10-second grace period to avoid using tokens that are
+// about to expire.
+func (te *tokenEntry) isExpired() bool {
+	if te.expiresAt.IsZero() {
+		// No expiration set, consider it never expires
+		return false
+	}
+	// Add 10-second grace period
+	return time.Now().Add(10 * time.Second).After(te.expiresAt)
```
Wwwsylvia: We should make the grace period a constant

registry/remote/auth/token_test.go
```diff
@@ -0,0 +1,296 @@
...
+			}
+		})
+	}
+}
+
+func TestTokenEntry_ExpirationGracePeriod(t *testing.T) {
+	// Create a token that expires in exactly 10 seconds
+	te := &tokenEntry{
+		token:     "test-token",
+		expiresAt: time.Now().Add(10 * time.Second),
```
Wwwsylvia: This could make the test flaky.

TerryHowe: Done

registry/remote/auth/token.go
TerryHowe: Done

registry/remote/auth/token.go
TerryHowe: Done.

registry/remote/auth/cache.go
```diff
@@ -90,8 +91,15 @@ func (cc *concurrentCache) GetToken(ctx context.Context, registry string, scheme
...
-	if token, ok := entry.tokens.Load(key); ok {
-		return token.(string), nil
+	if tokenValue, ok := entry.tokens.LoadAndDelete(key); ok {
+		te := tokenValue.(*tokenEntry)
+		if te.isExpired() {
+			// Token has expired, already removed from cache
+			return "", errdef.ErrNotFound
+		}
+		// Token is still valid, re-store it for future use
+		entry.tokens.Store(key, te)
```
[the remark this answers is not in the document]
Wwwsylvia: How about something like:

```go
te := tokenValue.(*tokenEntry)
if te.isExpired() {
    entry.tokens.CompareAndDelete(key, tokenValue)
    return "", errdef.ErrNotFound
}
```

registry/remote/auth/token.go
```diff
@@ -0,0 +1,104 @@
...
+	// token is the actual authentication token
+	token string
+	// expiresAt is when the token expires
+	expiresAt time.Time
+}
+
+// jwtClaims represents the standard claims in a JWT token.
+// We only parse the fields we need for expiration handling.
+type jwtClaims struct {
+	// Exp is the expiration time (seconds since Unix epoch)
```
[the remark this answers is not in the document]
TerryHowe: Done.

registry/remote/auth/cache.go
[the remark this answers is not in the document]
TerryHowe: I like that approach. Done.

Wwwsylvia (approved):
LGTM

registry/remote/auth/token.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+	"time"
+)
+
+const (
+	gracePeriodSeconds       = 10
+	defaultExpirationSeconds = 60
```
shizhMSFT, line 27: The default expiration time is too small. Vendors like ACR usually have access token valid for more than 1 hour.

Well, you may say ACR tokens are JWT tokens and they are fine but I still think we should leave the expiry check to the server. That is, if the client cannot determine the expiry time, we follow the previous behavior (i.e. let the server decide). By default, the auth client relies on 401 for token refresh.

shizhMSFT: So, it leaves another question, do we really need to track token expiration?

shizhMSFT: Here's the AI-generated summary of how the current re-auth on expired token works:

---

# ORAS-Go Authentication Token Expiry Strategy

## Current Implementation: Reactive/Passive Expiry Handling

ORAS-Go uses a **reactive approach** to token expiry rather than proactive time-based expiration.

### How It Works

1. **Token Caching**
   - Tokens are cached indefinitely in memory (`sync.Map`) with no TTL tracking
   - Cache key includes: registry host, auth scheme (Basic/Bearer), and scopes
   - No expiration metadata is stored or checked

2. **Expiry Detection**
   - Token expiry is detected when the registry returns `HTTP 401 Unauthorized`
   - The client attempts cached tokens first, then falls back to re-authentication on 401

3. **Re-authentication Flow**
   
   Request with cached token → 401 response → Fetch new token → Retry request

4. **Token Response Parsing**
   - Distribution spec tokens: extracts `token` or `access_token` fields
   - OAuth2 tokens: extracts `access_token` field
   - **Does NOT parse or use `expires_in` field** (if present in response)

### Trade-offs

**Advantages:**
- Simple implementation, no time tracking overhead
- Works universally without assumptions about token lifetime
- No clock synchronization issues between client/server

**Disadvantages:**
- Extra round-trip when tokens expire (request → 401 → re-auth → retry)
- Cannot proactively refresh tokens before expiry
- May cause latency spikes when tokens expire during critical operations

### Code References

- **Cache implementation**: `registry/remote/auth/cache.go`
  - `concurrentCache` struct uses `sync.Map` with no expiration
- **Client authentication**: `registry/remote/auth/client.go`
  - `Client.Do()` method handles 401 responses and re-authentication
  - Lines 351-360: Distribution token response parsing (no `expires_in`)
  - Lines 410-420: OAuth2 token response parsing (no `expires_in`)
- **Test validation**: `registry/remote/auth/client_test.go`
  - `TestClient_Do_Token_Expire` validates reactive expiry handling

### Compliance

- ✅ **OCI Distribution Spec**: Token format is compliant
- ⚠️ **OAuth2 RFC 6749**: Does not utilize optional `expires_in` field
- ✅ **Functional**: Works correctly with all compliant registries

PR #1038, opened 2025-11-11
feat!: remove ForceAttemptOAuth2 add SetLegacyMode to auth client

TerryHowe:
Remove ForceAttemptOAuth2 and add SetLegacyMode to the auth client.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,10 +18,12 @@ on:
     branches: 
       - main
       - release-*
+      - v2
   pull_request:
     branches:
       - main
       - release-*
+      - v2
 
 jobs:
   build:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -18,10 +18,12 @@ on:
     branches:
       - main
       - release-*
+      - v2
   pull_request:
     branches:
       - main
       - release-*
+      - v2
   schedule:
     - cron: '34 13 * * 3'
 
--- a/.github/workflows/license-checker.yml
+++ b/.github/workflows/license-checker.yml
@@ -18,10 +18,12 @@ on:
     branches:
       - main
       - release-*
+      - v2
   pull_request:
     branches:
       - main
       - release-*
+      - v2
 
 permissions:
   contents: write
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@ debug
 # Jetbrains
 .idea
 
+# Vi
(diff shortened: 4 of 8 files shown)


registry/remote/auth/client.go
```diff
@@ -121,13 +121,14 @@ type Client struct {
 	// Reference: https://distribution.github.io/distribution/spec/auth/oauth/#getting-a-token
 	ClientID string
 
-	// ForceAttemptOAuth2 controls whether to follow OAuth2 with password grant
-	// instead the distribution spec when authenticating using username and
-	// password.
+	// legacyMode controls whether to use the legacy distribution spec
+	// instead of OAuth2 with password grant when authenticating using
+	// username and password.
```
Wwwsylvia, line 126: nit: Would be nice to clarify that this is the approach used in v1 and v2.

registry/remote/auth/client.go
```diff
@@ -172,6 +173,21 @@ func (c *Client) SetUserAgent(userAgent string) {
 	c.Header.Set(headerUserAgent, userAgent)
 }
 
+// SetLegacyMode sets whether to use legacy distribution spec authentication
+// instead of OAuth2 with password grant when authenticating using username
+// and password.
+//
+// When legacy is true, the client uses the legacy distribution spec for
+// authentication. When legacy is false (default), the client uses OAuth2
+// with password grant.
```
Wwwsylvia: Same here

TerryHowe: Done.

registry/remote/auth/client.go
TerryHowe: Done.

registry/remote/auth/client.go
```diff
@@ -121,13 +121,25 @@ type Client struct {
...
+	//
+	// When legacyMode is false (default), the client uses OAuth2 with password
+	// grant.  If the registry says it supports bearer authentication, bearer
+	// authentication will be performed.
+	//
 	// References:
 	// - https://distribution.github.io/distribution/spec/auth/jwt/
 	// - https://distribution.github.io/distribution/spec/auth/oauth/
-	ForceAttemptOAuth2 bool
+	legacyMode bool
```
shizhMSFT, line 142: If we do want `legacyMode`, why not make it an exported field like `LegacyMode` so that we don't need getter and setter.

TerryHowe:
Comments have been used in feat/everything branch. Some future PR will cover this.

PR #1042, opened 2025-11-14
feat!: add support for tag and digest in Reference

TerryHowe:
This change adds support to capture both the tag and the digest in the registry Reference object.

* Renames Digest() method GetDigest()
* Adds Tag and Digest member for Form B
* Deprecates Reference.Reference member
* Adds GetReference() method for convenience and to replace Reference member

The reason the name is DigestString is to avoid conflict with the Digest() method which has an unfortunate name. It should be GetDigest()

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,10 +18,12 @@ on:
     branches: 
       - main
       - release-*
+      - v2
   pull_request:
     branches:
       - main
       - release-*
+      - v2
 
 jobs:
   build:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -18,10 +18,12 @@ on:
     branches:
       - main
       - release-*
+      - v2
   pull_request:
     branches:
       - main
       - release-*
+      - v2
   schedule:
     - cron: '34 13 * * 3'
 
--- a/.github/workflows/license-checker.yml
+++ b/.github/workflows/license-checker.yml
@@ -18,10 +18,12 @@ on:
     branches:
       - main
       - release-*
+      - v2
   pull_request:
     branches:
       - main
       - release-*
+      - v2
 
 permissions:
   contents: write
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@ debug
 # Jetbrains
 .idea
 
+# Vi
+*.sw*
+
 # Custom
 coverage.txt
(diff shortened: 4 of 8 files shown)


registry/reference.go
```diff
@@ -259,6 +285,12 @@ func (r Reference) Digest() (digest.Digest, error) {
 	return digest.Parse(r.Reference)
 }
 
+// GetReference returns the reference string (either a tag or a digest).
+// This method should be used instead of directly accessing the deprecated Reference field.
+func (r Reference) GetReference() string {
+	return r.Reference
+}
```
Wwwsylvia, line 299: What do we need this function for?

registry/reference.go
```diff
@@ -60,7 +60,18 @@ type Reference struct {
...
+	// For specific access to tag or digest values, use the Tag or DigestString fields.
 	Reference string
+
+	// Tag is the tag of the object in the repository, if a tag was provided.
+	// This field is empty if only a digest was provided (Form A) or if no reference was provided (Form D).
+	Tag string
+
+	// DigestString is the digest of the object in the repository, if a digest was provided.
+	// This field is empty if only a tag was provided (Form C) or if no reference was provided (Form D).
+	DigestString string
```
Wwwsylvia: Since we’re adding this, we might want to revisit the `ParseReference` implementation and refactor where needed, especially the validation logic. We should also consider updating the `Digest()` and `String()` implementations.

TerryHowe: Yes, I guess it is a matter of how many breaking changes we want.

registry/reference.go
TerryHowe: I forgot to implement this. Fixed.

registry/reference.go
TerryHowe: I agree the validation could use some clean up, but I think that could be another PR. I created https://github.com/oras-project/oras-go/issues/1056 to track that, comment away with what you were thinking.

I just made another change to this pull request to rename `Digest()` to `GetDigest()` and change `DigestString` to `Digest` which is a breaking change, but perhaps worth it.

TerryHowe: What were you thinking on `String()`? Probably another PR on that as well.

TerryHowe: I took this idea and made some changes that will be in another PR, thanks.

TerryHowe:
New PR will follow

PR #1047, opened 2025-11-16
refactor: split out methods from ParseReference

TerryHowe:
The ParseReference method is great for that, but the code in there could be shared to parse registries and repositories aka untagged references.

### DIFF
--- a/registry/reference.go
+++ b/registry/reference.go
@@ -168,7 +168,7 @@ func splitRepository(path string) (string, string, bool) {
 		return repository, reference, false
 	}
 
-	if index = strings.Index(path, ":"); index != -1 {
+	if index := strings.Index(path, ":"); index != -1 {
 		// `tag` found; Valid Form C
 		return path[:index], path[index+1:], true
 	}


registry/reference.go
```diff
@@ -176,6 +152,38 @@ func ParseReference(artifact string) (Reference, error) {
...
+		// `tag` found; Valid Form C
+		isTag = true
+		repository = path[:index]
+		reference = path[index+1:]
+	} else {
+		// empty `reference`; Valid Form D
+		repository = path
+	}
+	return repository, reference, isTag
+}
```
shizhMSFT: Try not use named return values. Consider

```suggestion
// splitRepository splits the path into repository and reference, indicating
// whether the reference is a tag or not.
func splitRepository(path string) (string, string, bool) {
	if index := strings.Index(path, "@"); index != -1 {
		// `digest` found; Valid Form A (if not B)
		repository := path[:index]
		reference := path[index+1:]

		if index = strings.Index(repository, ":"); index != -1 {
			// `tag` found (and now dropped without validation) since `the
			// `digest` already present; Valid Form B
			repository = repository[:index]
		}
		return repository, reference, false
	}

	if index = strings.Index(path, ":"); index != -1 {
		// `tag` found; Valid Form C
		return path[:index], path[index+1:], true
	}

	// empty `reference`; Valid Form D
	return path, "", false
}
```

shizhMSFT:
The trivial compile error needs to be fixed.

PR #1057, opened 2025-11-24
docs: add v3 migration doc

TerryHowe:


TerryHowe:
I've changed this so it creates a directory so we won't need to deal with merge conflicts. Each breaking change can have a short document and at the end we'll take them all add them to the docs site.

Issue #881, opened 2025-01-30
Move cache store from the oras cli to oras-go


AustinAbro321:
I am evaluating using the ORAS library in my product to enable image pulling. One feature I'd like is to enable users to cache images between runs. From my testing it seems like https://github.com/oras-project/oras/tree/main/internal/cache would be a good fit for this, but it's internal. I feel this could make sense to move to oras-go and allow library users to cache images, if so I can open up a PR, but I'd like to gauge thoughts first.

FeynmanZhou:
cc @oras-project/oras-go-maintainers @oras-project/oras-cli for help

Wwwsylvia:
Hi @AustinAbro321 , thank you for opening the issue! 

The cache implementation you see is dedicated to the ORAS CLI and may not work for other clients. As a library, oras-go should provide generic functions that are designed with care. Therefore, we need to consider this feature request carefully.

Could you elaborate on your scenarios for using the cache? Are you looking to cache blobs only, or do you also want to cache the tag resolving results? It would be helpful if you could share some code snippets where you would like to integrate a cache, so we can discuss how to implement this properly.

AustinAbro321:
Makes sense, it seems to work well for my purposes so far, but I have not tested it extensively. 

Here is what I would use it for, essentially to cache the blobs in an oci-format directory. In a real world case the images would be submitted by the user
```go
func main() {
	err := doOrasPull()
	if err != nil {
		panic(err)
	}
}

func doOrasPull() error {
	ctx := context.Background()
	client := auth.DefaultClient
	cwd, err := os.Getwd()
	if err != nil {
		return err
	}
	dst, err := oci.New(filepath.Join(cwd, "download"))
	if err != nil {
		return err
	}
	images := []string{
		"ghcr.io/fluxcd/image-automation-controller:v0.39.0",
		"ghcr.io/fluxcd/image-reflector-controller:v0.33.0",
		"ghcr.io/fluxcd/kustomize-controller:v1.4.0",
		"ghcr.io/fluxcd/notification-controller:v1.4.0",
		"ghcr.io/fluxcd/source-controller:v1.4.1",
	}
	copyOpts := oras.DefaultCopyOptions
	cachePath, err := oci.New(filepath.Join(cwd, "test-cache"))
	if err != nil {
		return err
	}
	for _, image := range images {
		localRepo := &remote.Repository{PlainHTTP: true}
		localRepo.Reference, err = registry.ParseReference(image)
		if err != nil {
[...]

PR #1086, opened 2025-12-27
feat: add cache support

TerryHowe:
Files Created

  1. content/cache/target.go - Core cache wrapper implementation
    - New(source, cache) - Creates a cached target that wraps a ReadOnlyTarget with a content.Storage cache
    - Fetch() - First checks cache, then fetches from source while caching the content
    - Exists() - Checks both cache and source
    - Supports registry.ReferenceFetcher - If the source implements it, the cached target does too
  2. content/cache/cache.go - Helper for environment variable-based caching
    - CacheEnvVar = "ORAS_CACHE" - Environment variable name
    - NewFromEnv() - Creates a Cache config from the environment variable
    - Cache.Enabled() - Returns true if caching is enabled
    - Cache.CachedTarget(source) - Wraps source with cache if enabled, uses OCI store
  3. content/cache/target_test.go - Comprehensive tests for the cache target
  4. content/cache/cache_test.go - Tests for the Cache helper
  5. content/cache/example_test.go - Example documentation

  Usage

  // Option 1: Environment variable based (like ORAS CLI)
  cache := cache.NewFromEnv() // reads ORAS_CACHE env var
  target, err := cache.CachedTarget(repo)

  // Option 2: Direct cache wrapping
  cachedTarget := cache.New(source, ociStore)

  This matches the pattern used in the ORAS CLI with the ORAS_CACHE environment variable triggering.

Closes: https://github.com/oras-project/oras-go/issues/881

### DIFF
--- a/content/cache/target_test.go
+++ b/content/cache/target_test.go
@@ -322,7 +322,7 @@ func TestReferenceTarget_FetchReference(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create repository:", err)
 	}
-	repo.Registry.PlainHTTP = true
+	repo.PlainHTTP = true
 
 	// create cache
 	cache := memory.New()
@@ -441,7 +441,7 @@ func TestNew_WithReferenceFetcher(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create repository:", err)
 	}
-	repo.Registry.PlainHTTP = true
+	repo.PlainHTTP = true
 
 	// create cache
 	cache := memory.New()
@@ -571,7 +571,7 @@ func TestReferenceTarget_FetchReference_Error(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create repository:", err)
 	}
-	repo.Registry.PlainHTTP = true
+	repo.PlainHTTP = true
 
 	cache := memory.New()
 	ctx := context.Background()
@@ -740,7 +740,7 @@ func TestReferenceTarget_FetchReference_CacheFetchFailsFallsBack(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create repository:", err)
 	}
-	repo.Registry.PlainHTTP = true
+	repo.PlainHTTP = true
 
 	// Cache that always fails on Fetch — simulates a corrupt or unavailable cache.
 	// FetchReference should fall back to the source and succeed.


TerryHowe:
I thought there was another request for cache in Helm, but only found https://github.com/helm/helm/issues/12602

cyberhck:
hey maintainers, what's next for this PR? what can be done to merge it? can we have a review? :)

wangxiaoxuan273:
@shizhMSFT Can you take a look?

content/cache/cache.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"os"
+
+	"github.com/oras-project/oras-go/v3"
+	"github.com/oras-project/oras-go/v3/content/oci"
+)
+
+// CacheEnvVar is the environment variable name for specifying the cache root.
+const CacheEnvVar = "ORAS_CACHE"
```
shizhMSFT: code style: use more readable name: `CacheEnvVar` --> `EnvVarName` and the call site will be `cache.EnvVarName` instead of `cache.CacheEnvVar`.

content/cache/cache.go
```diff
@@ -0,0 +1,60 @@
...
+func NewFromEnv() *Cache {
+	return &Cache{
+		Root: os.Getenv(CacheEnvVar),
+	}
+}
+
+// Enabled returns true if caching is enabled (i.e., Root is non-empty).
+func (c *Cache) Enabled() bool {
+	return c != nil && c.Root != ""
+}
```
shizhMSFT: It is unnecessary to have `Enabled()`. If the user does not want to use cache, the user should not create the cache in the first place.

content/cache/cache.go
```diff
@@ -0,0 +1,60 @@
...
+
+// CachedTarget wraps the source ReadOnlyTarget with a caching layer if
+// caching is enabled. If caching is disabled, the original source is returned.
+func (c *Cache) CachedTarget(source oras.ReadOnlyTarget) (oras.ReadOnlyTarget, error) {
+	if !c.Enabled() {
+		return source, nil
+	}
+
+	ociStore, err := oci.New(c.Root)
```
shizhMSFT: Note that OCI image layout has a `index.json`, which is not process-safe. If multiple processes write to the cache, which is intended for `ORAS_CACHE`, it will have race condition. That's why in the `oras` CLI, we only use `oci.Storage`, which is process-safe.

Although here we are creating cache for `ReadOnlyTarget`, we should honor all clients which uses `ORAS_CACHE`.


```suggestion
	ociStore, err := oci.NewStorage(c.Root)
```

content/cache/cache.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Enabled returns true if caching is enabled (i.e., Root is non-empty).
+func (c *Cache) Enabled() bool {
+	return c != nil && c.Root != ""
+}
+
+// CachedTarget wraps the source ReadOnlyTarget with a caching layer if
+// caching is enabled. If caching is disabled, the original source is returned.
+func (c *Cache) CachedTarget(source oras.ReadOnlyTarget) (oras.ReadOnlyTarget, error) {
```
shizhMSFT: nit: naming: should we have the method name consistent with the type? Like

```suggestion
func (c *Cache) ReadOnlyTarget(source oras.ReadOnlyTarget) (oras.ReadOnlyTarget, error) {
```
or 

```suggestion
func (c *Cache) CacheReadOnlyTarget(source oras.ReadOnlyTarget) (oras.ReadOnlyTarget, error) {
```

content/cache/target.go
```diff
@@ -0,0 +1,151 @@
...
+// target wraps a ReadOnlyTarget with a caching layer.
+type target struct {
+	oras.ReadOnlyTarget
+	cache content.Storage
+}
+
+// New creates a new cached target.
+// The returned target will first check the cache for content before
+// fetching from the source. Fetched content is cached while being read.
+func New(source oras.ReadOnlyTarget, cache content.Storage) oras.ReadOnlyTarget {
```
shizhMSFT:
```suggestion
func CacheReadOnlyTarget(source oras.ReadOnlyTarget, cache content.Storage) oras.ReadOnlyTarget {
```
With the name `CacheReadOnlyTarget` instead of `New`, it is clearer that we are doing read-only cache and the resulted instance may lose methods in addition to `oras.ReadOnlyTarget`.

shizhMSFT: ### Concern: Embedding an interface only promotes the interface’s method set

In Go, embedding an **interface type** inside a struct only promotes the methods **declared in that interface**.  
If the embedded interface contains a concrete value with additional methods, those extra methods are **not promoted**.

```go
package main

type Concrete struct{}

func (Concrete) Start() {}
func (Concrete) Stop()  {}

type Starter interface {
    Start()
}

type Engine struct {
    Starter // embeds the interface type
}

func main() {
    e := Engine{Starter: Concrete{}}

    e.Start() // OK: Start is declared in Starter
    // e.Stop() // ❌ compile error: Engine has no method Stop
}

Comment (shizhMSFT) [reply content/cache/target.go]: Since we are building an SDK, it is better if we can show the above concerns / limitations to the SDK user.

File: content/cache/cache.go (hunk 1 above)
Comment (TerryHowe) [reply content/cache/cache.go]: Done.

File: content/cache/cache.go (hunk 2 above)
Comment (TerryHowe) [reply content/cache/cache.go]: Done.

File: content/cache/cache.go (hunk 3 above)
Comment (TerryHowe) [reply content/cache/cache.go]: Done.

File: content/cache/cache.go (hunk 4 above)
Comment (TerryHowe) [reply content/cache/cache.go]: Done.

File: content/cache/target.go (hunk 5 above)
Comment (TerryHowe) [reply content/cache/target.go]: Done.

File: content/cache/target.go [hunk 6]
```diff
@@ -0,0 +1,151 @@
...
+	if exists {
+		if err := rc.Close(); err != nil {
+			return ocispec.Descriptor{}, nil, err
+		}
+		// get rc from the cache
+		rc, err = t.cache.Fetch(ctx, desc)
+		if err != nil {
+			return ocispec.Descriptor{}, nil, err
+		}
+		return desc, rc, nil
```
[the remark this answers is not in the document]
TerryHowe: Done.

content/cache/example_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/oras-project/oras-go/v3/internal/cas"
+)
+
+// ExampleNew demonstrates how to wrap a ReadOnlyTarget with a cache layer.
+func ExampleNew() {
+	// Create a source store
+	source := memory.New()
+
+	// Create a cache store
+	cacheStore := cas.NewMemory()
```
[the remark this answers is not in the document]
TerryHowe: Done.

content/cache/target_test.go
```diff
@@ -0,0 +1,807 @@
...
+	"net/url"
+	"reflect"
+	"testing"
+
+	"github.com/opencontainers/go-digest"
+	specs "github.com/opencontainers/image-spec/specs-go"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/oras-project/oras-go/v3"
+	"github.com/oras-project/oras-go/v3/content/memory"
+	"github.com/oras-project/oras-go/v3/internal/cas"
```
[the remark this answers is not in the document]
TerryHowe: Done.

content/cache/target.go
```diff
@@ -0,0 +1,151 @@
...
+// referenceTarget extends target with ReferenceFetcher support.
+type referenceTarget struct {
+	*target
+	registry.ReferenceFetcher
+}
+
+// FetchReference fetches the content identified by the reference from the
+// remote and caches the fetched content.
+// Cached content will only be read via Fetch; FetchReference will always
+// fetch from the source first.
```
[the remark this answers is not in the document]
TerryHowe: Done.

PR #1115, opened 2026-03-09
fix: guard GetCredentialFunc against nil store and fix Ingest error cleanup

TerryHowe:
## Summary

This PR fixes two bugs in the `registry/remote` package and adds missing test coverage.

**`GetCredentialFunc` nil store panic** (`registry/remote/auth.go`): When `GetCredentialFunc` was called with a `nil` store, it would panic at runtime when the returned function tried to call methods on the nil store. The fix adds a nil guard that returns a function yielding `EmptyCredential` with no error, matching the documented semantics.

**`Ingest` temp file leak on error** (`registry/remote/internal/ioutil/ioutil.go`): The `Ingest` function creates a temp file and uses a deferred cleanup to remove it on error. However, the error paths for `Chmod` and `io.Copy` previously used bare `return "", err` returns which bypass the named return value `ingestErr`, so the deferred cleanup never triggered and temp files were leaked. The fix switches both error paths to assign `ingestErr` and use a plain `return`, ensuring the defer always sees the error and removes the temp file.

**Typo fix** (`registry/remote/url.go`): Fixed missing space in the `buildRepositoryBlobMountURL` comment (`buildRepositoryBlobMountURLbuilds` → `buildRepositoryBlobMountURL builds`).

## Test plan
- [x] Added `TestGetCredentialFunc_NilStore` to verify nil store returns `EmptyCredential` without error
- [x] Added comprehensive tests for `Ingest`: success, name prefix, invalid dir, reader error with cleanup, empty content, and chmod error with cleanup

### DIFF
--- a/registry/remote/auth.go
+++ b/registry/remote/auth.go
@@ -68,8 +68,15 @@ func Logout(ctx context.Context, store credentials.Store, registryName string) e
 	return nil
 }
 
-// GetCredentialFunc returns a GetCredentialFunc() function that can be used by auth.Client.
+// GetCredentialFunc returns a CredentialFunc that retrieves credentials from
+// the given store. If store is nil, the returned function always returns
+// EmptyCredential without error.
 func GetCredentialFunc(store credentials.Store) credentials.CredentialFunc {
+	if store == nil {
+		return func(context.Context, string) (credentials.Credential, error) {
+			return credentials.EmptyCredential, nil
+		}
+	}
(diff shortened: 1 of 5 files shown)


TerryHowe:
Splitting this into two focused PRs since the fixes are independent:

- #1184 — `fix(remote): guard GetCredentialFunc against nil store` (plus typo fix)
- #1185 — `fix(ioutil): clean up temp file on Ingest error`

Closing in favor of those.

PR #1117, opened 2026-03-09
fix(cache): use resolve-first pattern in FetchReference to avoid unnecessary downloads

TerryHowe:
## Summary
- Fixes FetchReference in the cache layer to resolve the descriptor first before fetching content, avoiding unnecessary downloads when the content is already cached

## Test plan
- [ ] Unit tests pass (`make test`)


TerryHowe:
Superseded by #1086, which absorbed the resolve-first pattern in FetchReference. The fix is already in main — this branch's tree is identical to upstream/main after a rebase, so there's nothing left to merge. Closing.

PR #1118, opened 2026-03-09
refactor(remote): make SetReferrersCapability not return an error

TerryHowe:
## Summary
- Simplifies the `SetReferrersCapability` API by removing the error return value; the operation sets internal state and has no failure mode that warrants an error return

## Test plan
- [ ] Unit tests pass (`make test`)


TerryHowe:
Need a different approach

PR #1130, opened 2026-03-30
refactor(remote): move the referrers merge pool behind a lazy accessor

TerryHowe:
## What this is

This started as the umbrella branch holding the complete v3 change set. Everything on it has since landed on `main` through extracted PRs, so the branch has been rebased down to the one piece that was never extracted.

## Already merged (nothing left here)

| Area | PR |
|---|---|
| `builder.go`, `properties/attributes.go`, `config/*` | #1328 |
| `auth/{client,token,…}.go` | #1330 |
| `internal/configfile/`, `config/config.go`, `credentials/*` | #1331 |
| `referrers.go`, `ErrReferrersCapabilityAlreadySet` removal | #1333 |

## What remains

- `referrers_state.go` — give the referrers merge pool a named type wrapping `syncutil.Pool`.
- `repository.go` — swap the zero-value `syncutil.Pool` field for an `atomic.Pointer[referrerMergePool]` initialised on first use via CAS.
- Two comments in `blobStore.Exists` / `manifestStore.Exists` explaining why the context is marked as already policy-checked before `Resolve`.

`+147 / -3` across 3 files. Build, vet, gofmt and `go test ./registry/remote/...` are green.


TerryHowe:
Closing this out. Everything of substance on this branch has landed on `main` via the extracted PRs (#1328, #1330, #1331, #1333), so the umbrella has served its purpose.

The only piece never extracted was the referrers merge-pool refactor — moving the zero-value `syncutil.Pool` field to a lazily CAS-initialised `atomic.Pointer[referrerMergePool]`. That's optional polish rather than a fix: the zero value already works as a plain field, so the change buys indirection without a functional gain. Not worth carrying a branch for; happy to revisit as its own PR if anyone wants it.

PR #1151, opened 2026-04-25
fix: limit tag and referrer list page count

TerryHowe:
## Summary

- `Tags()` and `referrersByAPI()` followed server-supplied `Link` headers indefinitely with no page ceiling, allowing a malicious registry to drive unbounded round-trips and memory growth.
- Adds `TagListMaxPages` and `ReferrerListMaxPages` fields to both `Registry` (default) and `Repository` (override), mirroring the existing `TagListPageSize`/`ReferrerListPageSize` pattern.
- When a limit is set and exceeded, returns a wrapped `errdef.ErrTooManyPages` so callers can detect and handle the condition.
- Zero value (default) preserves the existing unlimited behavior for backward compatibility.

Closes GHSA-wv3q-vccw-6p3r

## Test plan

- [x] `TestRepository_Tags_MaxPages` — sets `TagListMaxPages=2` against a 3-page server, asserts `ErrTooManyPages`
- [x] `TestRepository_Referrers_MaxPages` — sets `ReferrerListMaxPages=2` against a 3-page server, asserts `ErrTooManyPages`
- [x] Existing `TestRepository_Tags` and `TestRepository_Referrers` pass unchanged (zero-value = unlimited)

### DIFF
--- a/errdef/errors.go
+++ b/errdef/errors.go
@@ -26,6 +26,7 @@ var (
 	ErrMissingReference   = errors.New("missing reference")
 	ErrNotFound           = errors.New("not found")
 	ErrSizeExceedsLimit   = errors.New("size exceeds limit")
+	ErrTooManyPages       = errors.New("too many pages")
 	ErrUnsupported        = errors.New("unsupported")
 	ErrUnsupportedVersion = errors.New("unsupported version")
 )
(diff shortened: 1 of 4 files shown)


TerryHowe:
Opening against fork instead.

PR #1152, opened 2026-04-25
fix: validate Location host before blob upload to prevent credential leak

TerryHowe:
## Summary

Fixes GHSA-jxpm-75mh-9fp7: credential forwarding via unvalidated `Location` header in blob upload.

During the monolithic blob upload flow, `completePushAfterInitialPost` follows the `Location` header from the registry's `POST` response and reuses the `Authorization` header for the subsequent `PUT`. A malicious registry can return a cross-host `Location`, causing oras-go to forward the caller's credentials to an attacker-controlled endpoint.

### Changes

- **`registry/remote/repository.go`**: After the existing port-443 workaround, validate that `location.Host` resolves to the same server as the original request before issuing the `PUT`. Also reject scheme downgrades (`https` → `http`). Adds `sameUploadHost()` helper that normalises implicit default ports (80/443) so that `example.com` and `example.com:443` compare equal over HTTPS — preserving the existing port-443 workaround behaviour.

- **`registry/remote/repository_test.go`**: Adds `Test_BlobStore_Push_CrossHostRedirect` — a malicious registry returns a cross-host `Location`; a second "attacker" server records whether it received an `Authorization` header. Asserts that `Push` returns an error and no credentials leak.

## Test plan

- [ ] `go test ./registry/remote/ -run Test_BlobStore_Push` — all four push tests pass (including the port-443 workaround tests)
- [ ] `go test ./...` — full suite passes with no regressions
- [ ] `Test_BlobStore_Push_CrossHostRedirect` explicitly verifies the fix

## References

- Advisory: https://github.com/oras-project/oras-go/security/advisories/GHSA-jxpm-75mh-9fp7
- Vulnerable function: `registry/remote/repository.go` → `blobStore.completePushAfterInitialPost`
- CWE-918 (SSRF), CVSS 7.5

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"mime"
 	"net/http"
+	"net/url"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

PR #1174, opened 2026-05-21
ci: fix v1 build workflow go version and pin action digests

TerryHowe:
Update actions/checkout, actions/setup-go, and actions/upload-artifact to current versions pinned by digest. Switch from go-version '1.23' to go-version-file to pick up the go 1.25.5 requirement from go.mod, which has been causing all v1 builds to fail since April 24.


TerryHowe:
Converted to draft because v1 is a low priority atm.

TerryHowe:
nvm

PR #1177, opened 2026-05-22
fix(remote): tolerate Content-Length mismatch in manifest fetch

### LINKED ISSUE #1176: fix(remote): manifestStore.Fetch fails with Content-Length mismatch on some registries (e.g. Harbor)
## Summary

`manifestStore.Fetch` in `registry/remote/repository.go` rejects manifest GET responses when the `Content-Length` header differs from `target.Size` (the descriptor's size, which was obtained from a prior HEAD/Resolve request). This causes `oras cp -r` to fail when copying to a **differently-named** destination repository on Harbor and potentially other registries.

## Root Cause

The check in `manifestStore.Fetch`:

```go
if size := resp.ContentLength; size != -1 && size != target.Size {
    return nil, fmt.Errorf("%s %q: mismatch Content-Length", resp.Request.Method, resp.Request.URL)
}
```

fails because some registries (notably Harbor) return different `Content-Length` values for HEAD (used by `Resolve`) and GET requests on the same manifest URL. This can occur due to transparent response compression applied only on GET, or other registry-side quirks.

The check is **redundant** for manifests: the `verifyContentDigest` call immediately after validates the `Docker-Content-Digest` response header against the expected digest, which is the authoritative integrity guarantee.

## Reproduction

See oras-project/oras#2054 for the user report. The failure mode:

```
oras cp -r --from-username ... --from-password ... --to-username ... --to-password ... \
  myharbor.mycorp.com/project_dev/repo/image:tag \
  myharbor.mycorp.com/project_prod/repo/different-name:tag
```

fails with:

```
Error from source registry for "...": GET ".../manifests/sha256:...": mismatch Content-Length
```

The same command succeeds if the destination **repository name** matches the source (only the tag differs), because in that case the manifest already exists at the destination and `copyGraph` skips the source fetch entirely.

## Fix

Remove the `Content-Length` check from `manifestStore.Fetch`. The `verifyContentDigest` check below it is sufficient for integrity.

> Note: the analogous check in `blobStore.Fetch` is kept — for blobs, size matters for range requests and progress tracking and the HEAD/GET divergence pattern is less common.

## Affected Version

Confirmed present on `feat/everything` (v3 branch) at `registry/remote/repository.go` lines 1417–1418. Also present in the v2 vendored copy used by oras-cli.

## Proposed Change

```diff
-	if size := resp.ContentLength; size != -1 && size != target.Size {
-		return nil, fmt.Errorf("%s %q: mismatch Content-Length", resp.Request.Method, resp.Request.URL)
-	}
 	if err := verifyContentDigest(resp, target.Digest); err != nil {
 		return nil, err
 	}
```

A fix with test update is ready on branch `fix/manifest-fetch-content-length-mismatch`.

TerryHowe:
## Summary

Some registries (e.g. Harbor) return different `Content-Length` values between HEAD (used by `Resolve`) and GET (used by `Fetch`) for the same manifest URL — typically due to transparent compression applied only on GET responses.

The existing check in `manifestStore.Fetch`:

```go
if size := resp.ContentLength; size != -1 && size != target.Size {
    return nil, fmt.Errorf("%s %q: mismatch Content-Length", ...)
}
```

was rejecting these valid responses. This causes `oras cp -r` to fail with `mismatch Content-Length` from the source registry when copying to a **differently-named** destination repository on Harbor. (With the same destination name the manifest already exists there, so the fetch is skipped and the error is never triggered.)

The check is redundant for manifests: `verifyContentDigest` immediately below validates the `Docker-Content-Digest` response header against the expected digest, which is the authoritative integrity guarantee.

## Changes

- Remove the two-line `Content-Length` check from `manifestStore.Fetch` in `registry/remote/repository.go`
- Update `Test_ManifestStore_Fetch` to verify the mismatch is now tolerated (renamed subtest from "fail with mismatching Content-Length" → "succeed with mismatching Content-Length")

The analogous check in `blobStore.Fetch` is intentionally kept — for blobs, size matters for range requests and progress tracking.

## Test plan

- [ ] `go test ./registry/remote/ -run Test_ManifestStore_Fetch` passes (all 5 subtests)
- [ ] Manual test: `oras cp -r` to a differently-named destination on a Harbor registry succeeds

Fixes #1176
Related: oras-project/oras#2054

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -1237,9 +1237,10 @@ func (s *manifestStore) Fetch(ctx context.Context, target ocispec.Descriptor) (r
 	if mediaType != target.MediaType {
 		return nil, fmt.Errorf("%s %q: mismatch response Content-Type %q: expect %q", resp.Request.Method, resp.Request.URL, mediaType, target.MediaType)
 	}
-	if size := resp.ContentLength; size != -1 && size != target.Size {
-		return nil, fmt.Errorf("%s %q: mismatch Content-Length", resp.Request.Method, resp.Request.URL)
-	}
+	// Content-Length is not validated here because some registries (e.g. Harbor)
+	// return different values between HEAD (used by Resolve) and GET responses,
+	// typically due to transparent compression. Integrity is guaranteed by
+	// verifyContentDigest below.
 	if err := verifyContentDigest(resp, target.Digest); err != nil {
 		return nil, err
 	}
--- a/registry/remote/repository_test.go
+++ b/registry/remote/repository_test.go
@@ -3930,7 +3930,10 @@ func Test_ManifestStore_Fetch(t *testing.T) {
 		}
 	})
 
-	t.Run("fail with mismatching Content-Length", func(t *testing.T) {
+	// Content-Length mismatches are tolerated for manifests because some
+	// registries (e.g. Harbor) return different values between HEAD (Resolve)
+	// and GET responses. Integrity is guaranteed by the digest check instead.
+	t.Run("succeed with mismatching Content-Length", func(t *testing.T) {
 		ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if r.Method != http.MethodGet {
 				t.Errorf("unexpected access: %s %s", r.Method, r.URL)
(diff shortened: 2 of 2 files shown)


sabre1041 (approved):
LGTM

PR #1141, opened 2026-04-07
feat(auth): add TokenFetcher interface and token abstraction

TerryHowe:
## Summary

- Adds `TokenFetcher` interface with `Distribution`, `OAuth2`, and `Composite` implementations extracted from the auth client monolith
- Replaces exported `ForceAttemptOAuth2` field with unexported `legacyMode` + public `SetLegacyMode()` setter for cleaner API
- Adds `ForceBasicAuth` field to force HTTP Basic auth regardless of registry-advertised scheme
- Allows callers to inject a custom `TokenFetcher` for advanced bearer token scenarios

## Details

The auth client previously had all token-fetching logic inlined. This PR extracts it into a `TokenFetcher` interface, making it composable and testable independently:

- **`TokenFetcher`** — interface with `FetchToken(ctx, TokenParams, Credential) (string, error)`
- **`DistributionTokenFetcher`** — fetches tokens via the distribution spec (basic auth to token endpoint)
- **`OAuth2TokenFetcher`** — fetches tokens via OAuth2 password/refresh grant
- **`CompositeTokenFetcher`** — selects between the above based on credential type

The `Client` struct gains a `TokenFetcher` field; if set, it is used instead of the built-in logic. The `legacyMode` field (replacing `ForceAttemptOAuth2`) preserves backward-compatible behavior for registries that don't support OAuth2.

## Depends on

- `internal/authtype` — already merged (oras-project/oras-go#1134)

## Test plan

- [x] `token.go` fully covered by `token_test.go`
- [x] `client_test.go` updated for `SetLegacyMode()` / `ForceBasicAuth` API
- [ ] Run `go test ./registry/remote/auth/...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1180, opened 2026-05-23
feat(auth)!: flip OAuth2 default and rename ForceAttemptOAuth2 to legacyMode

TerryHowe:
## Summary

Part 1 of 3 splitting #1141 into reviewable chunks.

Replace the exported `ForceAttemptOAuth2 bool` field on `auth.Client` with an unexported `legacyMode` field plus a public `SetLegacyMode(bool)` setter, and **flip the default behavior**:

| | Before | After |
|---|---|---|
| Default for username+password | distribution-spec token endpoint | OAuth2 password grant |
| Opt-in toggle | `client.ForceAttemptOAuth2 = true` | `client.SetLegacyMode(true)` |

The selection logic in `fetchBearerToken` is unchanged — only the polarity of the flag and the default are swapped.

## Breaking change

- Callers that relied on the **implicit distribution-spec default** (did not set `ForceAttemptOAuth2`, authenticated via username+password) must now call `client.SetLegacyMode(true)` to keep the old behavior.
- Callers that explicitly set `ForceAttemptOAuth2 = true` should drop that line — OAuth2 is the new default.
- `ForceAttemptOAuth2` is removed (no deprecation period). Happy to add a one-release deprecation shim if reviewers prefer.

## Test changes

- 8 distribution-spec tests (`TestClient_Do_Bearer_Auth*`, `TestClient_Do_Invalid_Credential_*`, `TestClient_Do_Scheme_Change`) gain `client.SetLegacyMode(true)`.
- 7 OAuth2 tests (`TestClient_Do_Bearer_OAuth2_Password*`, `TestClient_Do_Scope_Hint_Mismatch*`) drop `ForceAttemptOAuth2: true`.
- `example_test.go` drops the `ForceAttemptOAuth2` reference.
- Adds `TestClient_SetLegacyMode` for the new setter.

## Test plan

- [x] `go test -mod=mod ./registry/remote/auth/...` — passes
- [ ] CI on this PR

## Follow-ups in the split

- PR B: `feat(auth): add TokenFetcher interface and token abstraction` (based on this branch)
- PR C: `feat(auth): add ForceBasicAuth to override Bearer challenges` (independent, based on main)

Closes nothing on its own; replaces #1141 together with PRs B and C.

### DIFF
--- a/registry/remote/auth/client.go
+++ b/registry/remote/auth/client.go
@@ -99,13 +99,25 @@ type Client struct {
 	// Reference: https://distribution.github.io/distribution/spec/auth/oauth/#getting-a-token
 	ClientID string
 
-	// ForceAttemptOAuth2 controls whether to follow OAuth2 with password grant
-	// instead the distribution spec when authenticating using username and
-	// password.
+	// legacyMode controls whether to use the legacy distribution spec
+	// instead of OAuth2 with password grant when authenticating using
+	// username and password.
+	// Default is false (OAuth2 is used).
+	//
+	// When legacyMode is true, the client uses the legacy distribution spec for
+	// authentication. If the registry says it supports bearer authentication,
+	// basic authentication will be performed unless there are no credentials
+	// or a refresh token is provided. This is the approach used in oras-go
+	// v1 and v2.
+	//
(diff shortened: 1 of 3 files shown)


TerryHowe:
Another day on this

PR #1181, opened 2026-05-23
feat(auth): add TokenFetcher interface and token abstraction

TerryHowe:
## Summary

Part 2 of 3 splitting #1141 into reviewable chunks.

Extract bearer-token acquisition from `auth.Client` into a pluggable `TokenFetcher` interface:

```go
FetchToken(ctx, TokenParams, Credential) (string, error)
```

Three concrete implementations land in the same file:

| Type | Purpose |
|---|---|
| `DistributionTokenFetcher` | GET against the distribution-spec token endpoint with optional basic auth |
| `OAuth2TokenFetcher` | POST OAuth2 password / refresh_token grant |
| `CompositeTokenFetcher` | Selects between the above based on credential shape and a `LegacyMode` toggle — mirrors the existing in-client logic |

A new optional `Client.TokenFetcher` field lets callers inject a custom strategy (e.g. ECR/GCR-specific token exchange) without forking the client. When nil, the built-in fallback in `fetchBearerToken` preserves the existing behavior, so this is a **purely additive** change to the public API. `ForceAttemptOAuth2` is unchanged.

## Why split it out

Today the bearer-token strategy is hardcoded inside `Client.fetchBearerToken`. Callers who need a custom strategy have to fork or wrap the whole client. With `TokenFetcher`, they swap in one field.

The composite implementation is also easier to test in isolation than the inline logic — see `token_test.go` (805 lines) which exercises the strategies without spinning up an HTTP client at all.

This will replace https://github.com/oras-project/oras-go/blob/main/registry/remote/auth/client.go#L266:L400

## Test plan

- [x] `go test -mod=mod ./registry/remote/auth/...` — passes (existing tests + new TokenFetcher coverage)
- [x] `token_test.go` covers all three fetcher implementations independently
- [x] `client_test.go` adds `TestClient_Do_Bearer_CustomTokenFetcher` and `_Error` to verify wiring
- [ ] CI on this PR

## Follow-ups in the split

- PR A (#1180): `feat(auth)!: flip OAuth2 default and rename ForceAttemptOAuth2 to legacyMode` — separate breaking change, independent of this PR
- PR C (#1182): `feat(auth): add ForceBasicAuth to override Bearer challenges` — independent

Replaces #1141 together with PRs A and C.

### DIFF
--- a/registry/remote/auth/token.go
+++ b/registry/remote/auth/token.go
@@ -1,0 +1,253 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package auth
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/oras-project/oras-go/v3/registry/remote/credentials"
+	"github.com/oras-project/oras-go/v3/registry/remote/internal/errutil"
+)
+
+// TokenParams contains parameters for token acquisition.
+type TokenParams struct {
+	// Registry is the registry hostname (i.e. host:port).
+	Registry string
+	// Realm is the token endpoint URL.
+	Realm string
+	// Service is the service parameter from the WWW-Authenticate header.
+	Service string
+	// Scopes are the requested scopes for the token.
+	Scopes []string
+}
+
+// TokenFetcher abstracts the token acquisition strategy.
+type TokenFetcher interface {
+	// FetchToken fetches an access token for the given parameters and credential.
+	FetchToken(ctx context.Context, params TokenParams, cred credentials.Credential) (string, error)
+}
+
+// DistributionTokenFetcher implements distribution spec token endpoint (GET).
+// This fetches tokens using the legacy distribution specification.
+//
+// References:
+//   - https://distribution.github.io/distribution/spec/auth/jwt/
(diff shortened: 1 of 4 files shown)


registry/remote/auth/token.go
```diff
@@ -0,0 +1,253 @@
...
+		req.SetBasicAuth(cred.Username, cred.Password)
+	}
+	q := req.URL.Query()
+	if params.Service != "" {
+		q.Set("service", params.Service)
+	}
+	for _, scope := range params.Scopes {
+		q.Add("scope", scope)
+	}
+	req.URL.RawQuery = q.Encode()
```
gjenkins8, line 119: I suspect this code is copy/pasta'd from somewhere (?). But isn''t this unnecessary? (can remain if want to retain existing logic/code)

registry/remote/auth/token.go
```diff
@@ -0,0 +1,253 @@
...
+	if err := json.NewDecoder(lr).Decode(&result); err != nil {
+		return "", fmt.Errorf("%s %q: failed to decode response: %w", resp.Request.Method, resp.Request.URL, err)
+	}
+	if result.AccessToken != "" {
+		return result.AccessToken, nil
+	}
+	if result.Token != "" {
+		return result.Token, nil
+	}
+	return "", fmt.Errorf("%s %q: empty token returned", resp.Request.Method, resp.Request.URL)
```
gjenkins8: perhaps this response parsing logic could be a helper?

gjenkins8 (approved):
LGTM

Perhaps a common HTTP client that does the work of `DistributionTokenFetcher.send` and `OAuth2TokenFetcher.send` could be introduced (and would establish common baseline for HTTP request sending)

registry/remote/auth/token.go
TerryHowe: Yes, this is intentional and will be cleaned up in a follow-up. The client.go has some private methods that will be replaced with this, but I wanted to keep this PR smallish.

registry/remote/auth/token.go
TerryHowe: Done.

gjenkins8 (approved):
(Re) LGTM

In particular, the `TokenFetcher` interface w/ `TokenParams` representing the concrete registry details enables evolving how auth tokens are retrieved over time.

shizhMSFT (approved):
LGTM

PR #1182, opened 2026-05-23
feat(auth): add ForceBasicAuth to override Bearer challenges

TerryHowe:
## Summary

Part 3 of 3 splitting #1141 into reviewable chunks.

Add a `ForceBasicAuth bool` field to `auth.Client`. When true, a Bearer challenge from the registry is rewritten to a Basic challenge before credential handling. Useful for registries that advertise Bearer but also accept HTTP Basic for the same credential pair.

## Implementation

Three lines in `Client.Do()`, right after `parseChallenge()` and before the existing scheme switch:

\`\`\`go
if c.ForceBasicAuth && scheme == SchemeBearer {
    scheme = SchemeBasic
}
\`\`\`

The rest of the basic-auth path runs unchanged. The override bypasses the bearer flow entirely, so it does not interact with the token-fetching logic.

## Independence

Independent of PRs A (#1180) and B (#1181). Based directly on \`main\`. Can land in any order relative to the other two.

## Test plan

- [x] \`go test -mod=mod ./registry/remote/auth/...\` — passes
- [x] New \`TestClient_Do_ForceBasicAuth_OverridesBearer\` verifies: registry sends Bearer challenge → client retries with Basic header → 200
- [ ] CI on this PR

Replaces #1141 together with PRs A and B.

### DIFF
--- a/registry/remote/auth/client.go
+++ b/registry/remote/auth/client.go
@@ -106,6 +106,12 @@ type Client struct {
 	// - https://distribution.github.io/distribution/spec/auth/jwt/
 	// - https://distribution.github.io/distribution/spec/auth/oauth/
 	ForceAttemptOAuth2 bool
+
+	// ForceBasicAuth forces the use of HTTP Basic authentication regardless
+	// of what authentication scheme the registry advertises. When true, if
+	// the registry challenges with Bearer auth, Basic auth is used instead.
+	// This requires the registry to also accept Basic auth credentials.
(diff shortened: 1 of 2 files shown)


TerryHowe:
Closing: the auth direction on feat/everything is UseDistributionTokenAuth (the token-mechanism axis), which is where this work is going. ForceBasicAuth (challenge-scheme override) is a separate concept we're not pursuing. Not salvaging this.

PR #1184, opened 2026-05-23
fix(remote): guard GetCredentialFunc against nil store

TerryHowe:
## Summary

A `nil` `credentials.Store` passed to `GetCredentialFunc` would cause a deferred panic inside the returned closure at authentication time, making it hard to diagnose. This PR adds an early nil check that returns a no-op `CredentialFunc` yielding `EmptyCredential` when `store` is `nil`, matching the documented semantics.

Also fixes a small typo in the `buildRepositoryBlobMountURL` doc comment (missing space).

Split out from #1115 — the other half (Ingest temp-file cleanup) is in a separate PR.

## Test plan
- [x] Added `TestGetCredentialFunc_NilStore` to verify nil store returns `EmptyCredential` without error

### DIFF
--- a/registry/remote/auth.go
+++ b/registry/remote/auth.go
@@ -68,8 +68,15 @@ func Logout(ctx context.Context, store credentials.Store, registryName string) e
 	return nil
 }
 
-// GetCredentialFunc returns a GetCredentialFunc() function that can be used by auth.Client.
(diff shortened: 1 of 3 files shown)


sabre1041 (approved):
LGTM

PR #1185, opened 2026-05-23
fix(ioutil): clean up temp file on Ingest error

TerryHowe:
## Summary

`Ingest` in `registry/remote/internal/ioutil` creates a temp file and uses a deferred cleanup to remove it on error. The error paths for `Chmod` and `io.Copy` previously used bare `return "", err` returns, which bypassed the named return value `ingestErr` — so the deferred cleanup never triggered and temp files were leaked.

The fix switches both error paths to assign `ingestErr` and use a plain `return`, ensuring the defer always sees the error and removes the temp file.

A package-level `fileChmod` var is introduced so the chmod-failure branch can be exercised in tests.

Split out from #1115 — the other half (GetCredentialFunc nil guard) is in a separate PR.

## Test plan
- [x] Added comprehensive tests for `Ingest`: success, name prefix, invalid dir, reader error with cleanup, empty content, and chmod error with cleanup

### DIFF
--- a/registry/remote/internal/ioutil/ioutil.go
+++ b/registry/remote/internal/ioutil/ioutil.go
@@ -21,6 +21,9 @@ import (
 	"os"
 )
 
+// fileChmod is the function used to set file permissions, overridable in tests.
+var fileChmod = (*os.File).Chmod
+
 // Ingest writes content into a temporary ingest file with the file name format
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

PR #1186, opened 2026-05-23
feat(remote): add ReferrerCapability state machine and merge pool

TerryHowe:
## What

Adds a self-contained `ReferrerCapability` type and a `ReferrerMergePool` helper in `registry/remote`, plus a small doc-comment update on `ErrReferrersCapabilityAlreadySet`.

- `ReferrerCapability` — atomic state machine tracking per-registry referrers API support (unknown / supported / unsupported). Provides `IsSupported`/`IsUnsupported`/`IsUnknown`, strict `SetSupported`/`SetUnsupported` (returning `ErrReferrersCapabilityAlreadySet` on conflict), idempotent `TrySetSupported`/`TrySetUnsupported`, plus `State`/`Reset`.
- `ReferrerMergePool` — `syncutil.Pool[syncutil.Merge[referrerChange]]` wrapper that batches concurrent updates targeting the same referrers tag, exposing `Get` and `Do(referrersTag, change, prepare, update)` to coalesce read-modify-write cycles.
- Updates the `ErrReferrersCapabilityAlreadySet` doc comment to reference the new methods.

## Why

Part of the v3 PR-by-PR breakdown (PR 9: `feat/referrers-state`). Self-contained piece of the larger v3 work, reviewable independently before the repository integration PR plugs it in.

## Test plan

- [x] `go build -mod=mod ./registry/remote/...`
- [x] `go test -mod=mod ./registry/remote/` — passes (`ok ... 20.651s`)
- [x] New `referrers_state_test.go` covers state transitions, concurrent Set/TrySet semantics, and merge-pool batching

### DIFF
--- a/registry/remote/referrers_state.go
+++ b/registry/remote/referrers_state.go
@@ -1,0 +1,145 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package remote
+
+import (
+	"sync/atomic"
+
+	"github.com/oras-project/oras-go/v3/internal/syncutil"
+)
+
+// ReferrerCapability tracks a registry's Referrers API support.
+// It provides thread-safe state management for the referrers capability.
+type ReferrerCapability struct {
+	state int32
+}
+
+// NewReferrerCapability creates a new ReferrerCapability with unknown state.
+func NewReferrerCapability() *ReferrerCapability {
+	return &ReferrerCapability{
+		state: referrersStateUnknown,
+	}
+}
+
+// IsSupported returns true if the registry is known to support the Referrers API.
+func (c *ReferrerCapability) IsSupported() bool {
+	return atomic.LoadInt32(&c.state) == referrersStateSupported
+}
+
+// IsUnsupported returns true if the registry is known to NOT support the Referrers API.
+func (c *ReferrerCapability) IsUnsupported() bool {
+	return atomic.LoadInt32(&c.state) == referrersStateUnsupported
+}
+
+// IsUnknown returns true if the registry's Referrers API support is unknown.
+func (c *ReferrerCapability) IsUnknown() bool {
+	return atomic.LoadInt32(&c.state) == referrersStateUnknown
+}
+
+// State returns the current state of the referrers capability.
+func (c *ReferrerCapability) State() referrersState {
+	return atomic.LoadInt32(&c.state)
+}
+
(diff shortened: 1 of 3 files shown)


registry/remote/referrers_state.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"sync/atomic"
+
+	"github.com/oras-project/oras-go/v3/internal/syncutil"
+)
+
+// ReferrerCapability tracks a registry's Referrers API support.
+// It provides thread-safe state management for the referrers capability.
+type ReferrerCapability struct {
+	state int32
```
gjenkins8, line 27: curious as to why not?:


```suggestion
	state atomic.Int32
```

registry/remote/referrers_state.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// setState atomically sets the state, returning an error if the state
+// was already set to a different value.
+func (c *ReferrerCapability) setState(newState referrersState) error {
+	if swapped := atomic.CompareAndSwapInt32(&c.state, referrersStateUnknown, newState); swapped {
+		return nil
+	}
+	// Check if the current state matches what we're trying to set
+	if c.State() == newState {
```
gjenkins8, line 102: This `if` is IMHO unnecessary.

if `swapped` was true, I don't see how this would ever be false (presuming no racing (and it is a tiny race) modifications to an alternative state, which IMHO is a different problem).

registry/remote/referrers_state.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil
+	}
+	return ErrReferrersCapabilityAlreadySet
```
gjenkins8, line 105: I'm a little confused/uncertain about why we should/would return an error should the user attempt to set the same state.

Speculating, I assume it would be unusual to want to lookup and set the referrer state multiple times (And even more unusual/unlikely to want to set the referrer state to different values).

Unless there is a reason to detect (and error upon) setting the state multiple times(?),  the `CompareAndSwap` is unnecessary and it would be simple enough to simply do:

```
func (c *ReferrerCapability) setState(newState referrersState) { # no error return
    atomic.StoreInt32(&c.state, newState)
}
```

If it want wanted to detect attempts to set *different* states concurrently (which seems highly unusual), compare and swap would be used like:
```
func (c *ReferrerCapability) setState(newState referrersState) error {
    currentState := atomic.LoadInt32(&c.state, newState)
    if swapped := atomic.CompareAndSwapInt32(&c.state, referrersStateUnknown, newState); swapped {
       return nil
   }

    return ErrReferrersCapabilityConcurrentModification
}
```

Overall, I don't have enough context to know if setting the referrer state multiple times should be an error. Just picking up on an API oddity.

TerryHowe: This is for CLI to use to process command line args

gjenkins8 (approved):
Curiosity on the usage of atomic and `CompareAndSwap` - but Terry mentioned this code is being moved. LGTM

registry/remote/referrers_state.go
gjenkins8: Using concurrent-safe code (via atomic) for processing command line flags does seem like a mismatch.

It would seem like you would also want to support multiple sets of the value in this case too e.g. I would think a user would expect this to work (and not error):
```
oras --referrer unsupported --referrer supported # last one wins
```

TerryHowe:
Taking another approach on this

PR #1140, opened 2026-04-07
feat(properties): add remote properties package

TerryHowe:
## Summary

- Adds new `registry/remote/properties` package with structured types for configuring registry connections
- Provides `Registry`, `Reference`, `Transport`, `Mirror`, and `Attributes` structs
- No new inbound dependencies; foundational for builder, config, and registry reference changes

## Details

The `properties` package separates connection configuration from the `Repository`/`Registry` implementation, making it easier to compose and pass registry settings in a structured way without tight coupling.

**New types:**
- `Registry` — host, TLS, auth, retry, and credential settings for a registry
- `Reference` — structured image reference (registry + repository + tag/digest)
- `Transport` — custom `http.RoundTripper` configuration
- `Mirror` — mirror registry fallback settings
- `Attributes` — generic key/value metadata for registry objects

## Test plan

- [x] All new files have corresponding `_test.go` coverage
- [x] `example_test.go` provides runnable usage examples
- [ ] Run `go test ./registry/remote/properties/...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1188, opened 2026-05-23
feat(registry)!: add Tag and Digest fields to Reference

TerryHowe:
## What

Adds explicit `Tag` and `Digest` string fields to `registry.Reference`, fixes Form B parsing, and routes `ParseReference` through `properties.NewReference`.

- `Reference` gains `Tag` and `Digest` fields, populated by `ParseReference`. The legacy `Reference` string field is preserved (and still holds the digest for Form A/B / the tag for Form C) so existing `String()` output is unchanged.
- **Form B (`registry/repo:tag@digest`)**: the tag is now captured in `Reference.Tag` instead of being silently dropped. The digest still takes precedence in the `Reference` field.
- New `GetReference()` method returns the canonical reference string (digest if present, otherwise tag). Preferred over reading the now-deprecated `Reference` field directly.
- `ParseReference` now delegates to `properties.NewReference` (introduced in #1140).
- **Breaking**: the existing `Digest()` method is renamed to `GetDigest()` because the new `Digest` field would otherwise collide. In-tree callers in `registry/remote/repository.go` (3 call sites) and its tests are updated in the same commit.
- `Reference` (the type) and `Reference.Reference` (the field) are marked `Deprecated:` in favor of `registry/remote/properties.Reference`.

## Why

Part of the v3 PR-by-PR breakdown (**PR 14: `feat/registry-reference-updates`**). Depends on PR 4 — `feat(properties): add remote properties package` (#1140, merged) — for `properties.NewReference`.

The previous attempt at this work (#1042) was closed; this version makes the change minimal and self-contained.

## Test plan

- [x] `go build -mod=mod ./...`
- [x] `go test -mod=mod ./registry/... .` — all packages pass
- [x] `registry/reference_test.go` rewritten to cover Tag/Digest field population and GetReference()
- [x] `Repository.ParseReference` FQDN test cases updated to assert the new Tag/Digest fields

### DIFF
--- a/registry/example_reference_test.go
+++ b/registry/example_reference_test.go
@@ -34,7 +34,7 @@ func ExampleParseReference_digest() {
 	fmt.Println("Registry:", ref.Registry)
 	fmt.Println("Repository:", ref.Repository)
 
-	digest, err := ref.Digest()
+	digest, err := ref.GetDigest()
 	if err != nil {
 		panic(err)
 	}
--- a/registry/reference.go
+++ b/registry/reference.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/opencontainers/go-digest"
 	"github.com/oras-project/oras-go/v3/errdef"
+	"github.com/oras-project/oras-go/v3/registry/remote/properties"
 )
 
 // regular expressions for components.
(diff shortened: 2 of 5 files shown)


sabre1041 (approved):
LGTM

PR #1189, opened 2026-05-23
feat: add PolicyCheck callback to CopyOptions

TerryHowe:
## What

Adds an optional `PolicyCheck` callback to `CopyOptions`:

```go
PolicyCheck func(ctx context.Context, srcRef string) error
```

`Copy` invokes it once with the source reference before any I/O. A non-nil return aborts the copy and the error is wrapped via `%w` (`policy check failed for %s: %w`).

When `PolicyCheck` is nil the behavior is unchanged.

## Why

Lets callers plug in policy enforcement (e.g., containers-policy.json signature/identity checks) without the root `oras` package taking a build-time dependency on the `policy` package. Keeps the `oras` <-> `registry/remote/policy` boundary clean.

Part of the v3 PR-by-PR breakdown (**PR 17: `feat/copy-policy-check`**). Self-contained; no new package deps.

(Note: prs.md lists `content_test.go`, `example_test.go`, and `example_copy_test.go` under this PR, but those diffs in `feat/everything` are actually `repo.Registry.PlainHTTP` / `NewCredentialFunc` renames that belong to PR 13. They are intentionally excluded here so this PR stays scoped to PolicyCheck.)

## Test plan

- [x] `go build -mod=mod ./...`
- [x] `go test -mod=mod -run TestCopy_PolicyCheck -v .` — PASS for `rejected`, `allowed`, `no policy` subtests
- [x] `go test -mod=mod -run Copy .` — full Copy test family passes

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -54,6 +54,12 @@ type CopyOptions struct {
 	// reference will be passed to MapRoot, and the mapped descriptor will be
 	// used as the root node for copy.
 	MapRoot func(ctx context.Context, src content.ReadOnlyStorage, root ocispec.Descriptor) (ocispec.Descriptor, error)
+	// PolicyCheck is an optional callback invoked before copying begins.
+	// If set, it is called with the source reference string. If it returns
+	// a non-nil error, the copy is aborted. This allows decoupled policy
+	// enforcement (e.g., containers-policy.json) without importing the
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

PR #1192, opened 2026-05-28
fix: validate Location host before blob upload to prevent credential leak

TerryHowe:
## Summary

Backports the GHSA-jxpm-75mh-9fp7 fix from the `v2` line (originally #1152) to `main`.

Validates that the `Location` header returned by the registry POST response refers to the same host as the original request before issuing the PUT and forwarding the `Authorization` header. A malicious registry could otherwise return a cross-host `Location`, causing oras-go to send the caller's credentials to an attacker-controlled endpoint. Scheme downgrades (https → http) on the `Location` URL are also rejected.

A new `sameUploadHost()` helper normalizes implicit default ports (80/443) so that e.g. `example.com` and `example.com:443` compare equal over HTTPS, preserving the existing port-443 workaround behaviour.

Fixes: GHSA-jxpm-75mh-9fp7

Cherry-picked from commit b4437a640bf329efc16adfd995d2236f235da268.

## Test plan

- [x] `go build ./registry/remote/...`
- [x] `go test ./registry/remote/` passes

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"mime"
 	"net/http"
+	"net/url"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


TerryHowe:
nvm

PR #1203, opened 2026-06-05
enhance: add sync logic when save file

ningmingxiao:
ping @TerryHowe  can you take a look?


TerryHowe (approved):
/lgtm

I'm curious on what others think about the performance hit here over reliability. A large number of small blobs would be a lot slower I'd think.

ningmingxiao:
or we can add some benchmark test

TerryHowe (changes requested):
Going to need some tests and benchmarks would be useful

PR #1195, opened 2026-05-31
release: v2.6.1

TerryHowe:
This is a security patch release addressing five advisories in the authentication, remote, and content layers, plus accumulated bug fixes and maintenance since v2.6.0.

## Security Fixes

- Drop the `Authorization` header on cross-origin redirects to prevent origin credentials leaking to a redirect target on a different scheme/port of the same host ([GHSA-vh4v-2xq2-g5cg](https://github.com/oras-project/oras-go/security/advisories/GHSA-vh4v-2xq2-g5cg))
- Validate the bearer `realm` host before sending credentials to prevent credential exfiltration to an attacker-controlled token service, including TLS downgrades and IP-literal metadata endpoints; adds `TrustedRealmHosts` ([GHSA-28r5-37g7-p6mp](https://github.com/oras-project/oras-go/security/advisories/GHSA-28r5-37g7-p6mp), [GHSA-xf85-363p-868w](https://github.com/oras-project/oras-go/security/advisories/GHSA-xf85-363p-868w))
- Validate the `Location` host before blob upload to prevent credentials being forwarded to a cross-host upload endpoint (SSRF / CWE-918) (#1152, [GHSA-jxpm-75mh-9fp7](https://github.com/oras-project/oras-go/security/advisories/GHSA-jxpm-75mh-9fp7))
- Reject descriptor sizes exceeding 32 MiB in `content.ReadAll` to prevent a crafted OCI layout from triggering a `makeslice` panic and crashing the process (#1153, [GHSA-f36w-mj3v-6jqv](https://github.com/oras-project/oras-go/security/advisories/GHSA-f36w-mj3v-6jqv))
- Resolve symlinks when enforcing the `workingDir` write boundary in `content/file`, blocking writes that escape the boundary via a symlinked path component when `AllowPathTraversalOnWrite=false`

## Bug Fixes

- `graph.Memory` should use digest as map key (#1095)
- Fix credentials key for the Docker `registry-1` host (#966)
- Support an empty credentials file (#959)

## Other Changes

- Add GitOps release workflow with goreleaser (#1161)
- Shift the Go support window to [1.24, 1.25] (#991)
- Run `go modernize` (#1005)
- Sync `CODEOWNERS` and `OWNERS.md` from main to v2 (#1122)
- Remove scripts reference from the Makefile (#960)
- Bump `golang.org/x/sync` 0.14.0 → 0.20.0 (#971, #978, #1001, #1037, #1078, #1121)
[...]

PR #1209, opened 2026-06-09
ci: flag fork-based release PRs before and after merge

TerryHowe:
## Summary

The release workflow triggers only on merge and its `GITHUB_TOKEN` can push tags only to `oras-project/oras-go`. Because of that, the `release` job carried a job-level guard requiring the PR head to be in this repo — but when a release PR is opened from a **fork**, that guard makes the job **silently skip** (0s, no failure, no notification). That is exactly what happened to `v2.6.1` in #1195: the merge succeeded but no tag or release was produced, and nothing flagged it.

This PR makes that failure mode visible — both before and after merge.

## Changes

- **`.github/workflows/release-pr-check.yml` (new)** — a pre-merge guard that runs on open/reopen/synchronize of a `release/*` PR targeting `v2` and **fails as a status check in the merge box** when the PR comes from a fork. Runs in the safe fork-PR context (read-only token, no secrets). Can be made a required status check on `v2` branch protection to block the merge entirely.
- **`.github/workflows/release.yml`** — move the fork condition out of the job-level `if` and into an explicit failing step, so a fork PR that slips through produces a **red, notifying failure** instead of a silent skip.
- **`RELEASES.md`** — instruct pushing the release branch to `upstream` (not a personal fork), with a note explaining why a fork-based release PR cannot produce a release.

## Notes

- `v2.6.1` itself was already tagged and released manually after #1195 skipped.
- Consider adding `release-pr-check` as a required status check on the `v2` branch protection rule so fork-based release PRs cannot be merged.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -14,8 +14,16 @@ opened. Use an empty commit as a lightweight marker:
 git fetch upstream
 git checkout -b release/v2.7.0 upstream/v2
 git commit --allow-empty -s -m "chore: prepare release v2.7.0"
-git push origin release/v2.7.0
+git push upstream release/v2.7.0
 ```
+
+> **Important:** push the release branch to the upstream repository
+> (`oras-project/oras-go`), **not** a personal fork. The release workflow runs
+> only after merge and its `GITHUB_TOKEN` can only push tags to
+> `oras-project/oras-go`, so a release PR opened from a fork is rejected by the
+> `release-pr-check` workflow and — if merged anyway — produces no tag or
+> release. If your `origin` remote points at a fork, push to `upstream`
+> explicitly as shown above.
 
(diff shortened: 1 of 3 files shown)


RELEASES.md
```diff
@@ -14,9 +14,17 @@ opened. Use an empty commit as a lightweight marker:
 git fetch upstream
 git checkout -b release/v2.7.0 upstream/v2
 git commit --allow-empty -s -m "chore: prepare release v2.7.0"
-git push origin release/v2.7.0
+git push upstream release/v2.7.0
```
TerryHowe:
```suggestion
git push origin release/v2.7.0
```

TerryHowe:
Nice to have, but I have other fish to fry

PR #1212, opened 2026-06-10
fix(copy): re-push referenced manifests when the destination reports them missing

### LINKED ISSUE #1211: CopyGraph fails to self-heal a partially-copied multi-arch index (MANIFEST_BLOB_UNKNOWN on retry)
## Summary
`CopyGraph` (and therefore `oras cp --recursive` in the ORAS CLI) cannot recover a multi-arch image **index** whose copy was previously interrupted. Re-running the copy fails repeatedly with `MANIFEST_BLOB_UNKNOWN` on the index push and never self-heals.

## Affected versions
Reproduced on both `main` (v3) and the `v2` branch.

## What happens
Copying a multi-arch image index (for example `mongo:7.0.34`, an `application/vnd.oci.image.index.v1+json` referencing per-platform child manifests) fails on the index push:

```
Error response from registry: manifest blob unknown: blob unknown to registry: sha256:<child-digest>
```

Re-running the identical copy reports the child manifests as already present (`✓ Exists`), skips them, and fails the index push again — every time. The only manual recovery is to copy each child manifest by digest and then re-push the index.

## Root cause
In `copyGraph` (`copy.go`), each node is skipped when `dst.Exists(desc)` reports it present:

```go
// skip if a rooted sub-DAG exists
exists, err := dst.Exists(ctx, desc)
// ...
if exists {
    // ...
    return nil
}
```

If a previous copy was interrupted after a child manifest's **blobs** were uploaded but before that child **manifest** was durably committed, the destination can report the child manifest as existing (HEAD/`Exists` succeeds) while still rejecting any parent (the index) that references it. Because the child is skipped on every run, the index push fails identically forever.

The registry's rejection is itself spec-consistent — an index that references absent child manifests must be rejected (see the [distribution-spec error codes](https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#error-codes)). The gap is that `CopyGraph` does not reconcile or retry when this happens.

## Proposed fix
When a node push fails because the destination is missing referenced content (`MANIFEST_BLOB_UNKNOWN` / `BLOB_UNKNOWN` / `MANIFEST_UNKNOWN`), re-copy the node's manifest successors unconditionally (bypassing the existence short-circuit) and retry the push once, so an interrupted copy can self-heal. The retry is a no-op on the success path.

I have a fix plus a deterministic reproduction test ready for both `main` and `v2` and will open PRs that reference this issue.

## Design question for maintainers
Detection can be done either:
1. directly via the `registry/remote/errcode` codes from `copyGraph` (self-contained; what the PR currently does), or
2. by introducing a sentinel in `errdef` that the remote `manifestStore` maps `MANIFEST_BLOB_UNKNOWN` / `BLOB_UNKNOWN` onto, keeping `copyGraph` transport-agnostic.

Happy to take direction on the preferred layering.

Orexii:
## What this does
Makes `CopyGraph` self-heal a partially-copied graph. When the destination rejects a manifest push because it is missing referenced content (`MANIFEST_BLOB_UNKNOWN` / `BLOB_UNKNOWN` / `MANIFEST_UNKNOWN`), the copy now re-pushes the node's manifest successors and retries the push once, instead of returning the error.

## Why
`CopyGraph` skips a node whose sub-DAG it believes already exists ("skip if a rooted sub-DAG exists"). If a prior copy was interrupted after a child manifest's blobs were uploaded but before the child manifest was committed, the destination can report the child as present (`Exists`/HEAD) yet reject the parent that references it. For a multi-arch image index this means the index push fails with `MANIFEST_BLOB_UNKNOWN`, and since the child is skipped on every run, re-running the same `oras cp --recursive` never recovers.

This was hit in production mirroring a multi-arch image (`mongo:7.0.34`) between two registries; the only manual workaround was copying each child manifest by digest and re-pushing the index.

## How
- `copyGraph`: wrap the node push; on a "missing referenced content" error, re-copy the manifest successors unconditionally (bypassing the existence short-circuit) and retry the push once. No-op on the success path; blob successors are not re-pushed.
- `isMissingReferencedContentError`: typed classification via `errcode` (handles both a single `errcode.Error` and a multi-error `errcode.Errors`); no string matching.

## Testing
- `TestCopyGraph_SelfHealsInterruptedIndexCopy`: end-to-end reproduction against a destination that emulates the interrupted state — fails without this change (`MANIFEST_BLOB_UNKNOWN`), passes with it.
- `TestIsMissingReferencedContentError`: unit test for the classifier (100% covered).
- `go test -race ./...` passes; `gofmt` / `go vet` clean; new files carry the Apache license header.

## Notes for reviewers
- Design choice: detection currently uses the registry-specific `errcode` package from transport-agnostic `copyGraph`. An alternative is a sentinel in `errdef` that the remote `manifestStore` maps `MANIFEST_BLOB_UNKNOWN` / `BLOB_UNKNOWN` onto. Happy to switch if you prefer that layering.
- Scope: recovers the immediate manifest successors (the index -> child-manifest case). No API or signature changes; no breaking changes.

Fixes #1211

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -31,6 +31,7 @@ import (
 	"github.com/oras-project/oras-go/v3/internal/status"
 	"github.com/oras-project/oras-go/v3/internal/syncutil"
 	"github.com/oras-project/oras-go/v3/registry"
+	"github.com/oras-project/oras-go/v3/registry/remote/errcode"
 	"golang.org/x/sync/semaphore"
 )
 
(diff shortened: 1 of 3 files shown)


TerryHowe (approved):
/lgtm

Orexii:
@TerryHowe anything needed to do on my end for this? Please let me know. Thanks

PR #1213, opened 2026-06-10
fix(copy): re-push referenced manifests when the destination reports them missing

Orexii:
Backport of #1212 to the `v2` branch.

## What this does
Makes `CopyGraph` self-heal a partially-copied graph. When the destination rejects a manifest push because it is missing referenced content (`MANIFEST_BLOB_UNKNOWN` / `BLOB_UNKNOWN` / `MANIFEST_UNKNOWN`), the copy now re-pushes the node's manifest successors and retries the push once, instead of returning the error.

## Why
`CopyGraph` skips a node whose sub-DAG it believes already exists ("skip if a rooted sub-DAG exists"). If a prior copy was interrupted after a child manifest's blobs were uploaded but before the child manifest was committed, the destination can report the child as present (`Exists`/HEAD) yet reject the parent that references it. For a multi-arch image index this means the index push fails with `MANIFEST_BLOB_UNKNOWN`, and since the child is skipped on every run, re-running the same `oras cp --recursive` never recovers. This is the line that reaches the ORAS CLI.

This was hit in production mirroring a multi-arch image (`mongo:7.0.34`) between two registries; the only manual workaround was copying each child manifest by digest and re-pushing the index.

## How
- `copyGraph`: wrap the node push; on a "missing referenced content" error, re-copy the manifest successors unconditionally (bypassing the existence short-circuit) and retry the push once. No-op on the success path; blob successors are not re-pushed.
- `isMissingReferencedContentError`: typed classification via `errcode` (handles both a single `errcode.Error` and a multi-error `errcode.Errors`); no string matching.

## Testing
- `TestCopyGraph_SelfHealsInterruptedIndexCopy`: end-to-end reproduction against a destination that emulates the interrupted state — fails without this change (`MANIFEST_BLOB_UNKNOWN`), passes with it.
- `TestIsMissingReferencedContentError`: unit test for the classifier (100% covered).
- `go test -race ./...` passes; `gofmt` / `go vet` clean; new files carry the Apache license header.

The diff is identical to #1212 apart from the module import paths (`github.com/oras-project/oras-go/v3` -> `oras.land/oras-go/v2`).

Relates to #1211.

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -32,6 +32,7 @@ import (
 	"oras.land/oras-go/v2/internal/status"
 	"oras.land/oras-go/v2/internal/syncutil"
 	"oras.land/oras-go/v2/registry"
+	"oras.land/oras-go/v2/registry/remote/errcode"
 )
 
 // defaultConcurrency is the default value of CopyGraphOptions.Concurrency.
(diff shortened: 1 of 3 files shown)


Orexii:
@TerryHowe anything needed to do on my end for this? Please let me know. Thanks

TerryHowe (approved):
/lgtm

PR #1214, opened 2026-06-11
fix: bound tag and referrer list pagination to prevent unbounded requests

TerryHowe:
## Summary

Tag and referrer listing in `Repository` follow the server-provided `Link: rel="next"` header until the registry stops sending one, with **no cap on the number of pages**. A malicious or misbehaving registry can advertise an endless chain of pages — including a self-referential `next` link pointing back at the same endpoint — and force the client into unbounded HTTP requests and resource consumption (a client-side denial of service).

This is the library-side mitigation for the pagination issue described in **GHSA-298f-872v-2rcx**.

## Changes

- Add `TagListMaxPages` and `ReferrerListMaxPages` fields to `Repository` (propagated via `clone()`), bounding the number of pages followed during tag and referrer listing.
- Enforce the caps in `Tags()` and `referrersByAPI()`; when the limit is exceeded, listing returns the new `errdef.ErrTooManyPages`.
- Zero — the default — means unlimited, preserving existing behavior; callers opt in by setting a limit.

## Tests

- `TestRepository_Tags_MaxPages` and `TestRepository_Referrers_MaxPages` drive an `httptest` server that returns an endless `Link` chain and assert the listing terminates with `errdef.ErrTooManyPages`.

## Notes

This addresses the **oras-go (pagination)** half of GHSA-298f-872v-2rcx. The advisory also describes a recursive `oras discover` graph traversal issue, which lives in the oras CLI rather than this library.

### DIFF
--- a/errdef/errors.go
+++ b/errdef/errors.go
@@ -26,6 +26,7 @@ var (
 	ErrMissingReference   = errors.New("missing reference")
 	ErrNotFound           = errors.New("not found")
 	ErrSizeExceedsLimit   = errors.New("size exceeds limit")
+	ErrTooManyPages       = errors.New("too many pages")
 	ErrUnsupported        = errors.New("unsupported")
 	ErrUnsupportedVersion = errors.New("unsupported version")
 )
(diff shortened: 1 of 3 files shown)


sabre1041 (approved):
LGTM

PR #1217, opened 2026-06-16
feat: add untag support for remote registries

### LINKED ISSUE #1028: Implement optional tag deletion API
Implement optional tag deletion API.

> Just a note: The [tag deletion API](https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#content-management) is an optional feature in the distribution-spec, which states,
> 
> A registry MAY implement tag deletion, while others MAY allow deletion only by manifest.
> 
> To support this feature, it would need to be implemented on the underlying `oras-go` library side.

 _Originally posted by @Wwwsylvia in [#1752](https://github.com/oras-project/oras/issues/1752#issuecomment-2961881761)_

matthiasbruns:
**Description**

This PR adds tag deletion support for remote registries as per https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#deleting-tags.

For this, it extends `registry/remote/repository.go` and implements  `Untagger` from `content/resolver.go` introduced in https://github.com/oras-project/oras-go/pull/647.
For our downstream project, we implemented this by ourselves and thought this would be a good addition to bring to upstream.

It adds a guard by using `ValidateReferenceAsTag` to not allow deletion by digests, only tags are allowed.

Closes: https://github.com/oras-project/oras-go/issues/1028

**Testing**
- Added `ExampleRepository_Untag` to `example_test.go`
- Added `TestRepository_Untag` to `repository_test.go`

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -440,9 +440,6 @@ func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference
 //
 // Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#deleting-tags
 func (r *Repository) Untag(ctx context.Context, reference string) error {
-	if reference == "" {
-		return errdef.ErrMissingReference
-	}
 	if err := r.checkPolicy(ctx, reference); err != nil {
 		return err
 	}
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -428,6 +428,49 @@ func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference
 	return r.Manifests().Tag(withPolicyChecked(ctx), desc, reference)
 }
 
+// Untag removes the association between the given tag and the manifest it
+// currently points to. The underlying content is NOT deleted.
+//
+// Tag deletion is an optional capability per the OCI Distribution Spec
+// and is not supported by all registries.
+//
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#deleting-manifests
+func (r *Repository) Untag(ctx context.Context, reference string) error {
+	if reference == "" {
+		return errdef.ErrMissingReference
+	}
+	if err := r.checkPolicy(ctx, reference); err != nil {
+		return err
+	}
+	ref, err := r.ParseReference(reference)
+	if err != nil {
+		return err
+	}
+	if err := ref.ValidateReferenceAsTag(); err != nil {
+		return err
+	}
+	ctx = auth.AppendRepositoryScope(ctx, ref, auth.ActionDelete)
+	url := buildRepositoryManifestURL(r.PlainHTTP, ref)
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, url, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := r.do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusOK, http.StatusAccepted, http.StatusNoContent:
+		return nil
+	case http.StatusNotFound:
+		return fmt.Errorf("%s: %w", ref.GetReference(), errdef.ErrNotFound)
+	default:
+		return errutil.ParseErrorResponse(resp)
+	}
+}
+
 // PushReference pushes the manifest with a reference tag.
 func (r *Repository) PushReference(ctx context.Context, expected ocispec.Descriptor, content io.Reader, reference string) error {
 	if err := r.checkPolicy(ctx, reference); err != nil {
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -434,7 +434,7 @@ func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference
 // Tag deletion is an optional capability per the OCI Distribution Spec
 // and is not supported by all registries.
 //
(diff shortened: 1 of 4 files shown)


registry/remote/repository.go
```diff
@@ -428,6 +428,49 @@ func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference
 	return r.Manifests().Tag(withPolicyChecked(ctx), desc, reference)
 }
 
+// Untag removes the association between the given tag and the manifest it
+// currently points to. The underlying content is NOT deleted.
```
TerryHowe:
```suggestion
// currently points to.

File: registry/remote/repository.go [hunk 2]
```diff
@@ -428,6 +428,49 @@ func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference
...
+	if err := ref.ValidateReferenceAsTag(); err != nil {
+		return err
+	}
+	ctx = auth.AppendRepositoryScope(ctx, ref, auth.ActionDelete)
+	url := buildRepositoryManifestURL(r.PlainHTTP, ref)
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, url, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := r.do(req)
```
TerryHowe: Every other operation here like `Tag` and others delegates to the manifestStore and for consistency, this should too.

TerryHowe (changes requested):
Thanks, couple comments

registry/remote/repository.go
matthiasbruns: My bad - moved it into the `manifestStore` implementation.
One question though - 
https://github.com/oras-project/oras-go/blob/0932d3634a4d4a28ef7cc2038e2f65fa6f0b15de/registry/interfaces.go#L66

```go
type ManifestStore interface {
	BlobStore
	content.Tagger
	ReferencePusher
}
```
does not include `content.Untagger` yet. Adding it there would break implementors.

What I did is treat is as optional:
https://github.com/oras-project/oras-go/pull/1217/changes#diff-eb396d2d573ffcc5766c387d68ae4c0cdc0e11be3ab0c4eafe45d8673345ae43R447

technically, in this code at the current state we wouldn't need to be safe about, since the `Repository` owns `Manifests()`, but I wouldn't want a `panic` if something is being changes in the future and this would be missed.

So two options:
- add `content.Untagger` to ManifestStore -> breaking, and maybe not desired by design
- go the route with the optional interface handling as proposed in this PR

registry/remote/repository.go
```diff
@@ -428,6 +428,31 @@ func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference
 	return r.Manifests().Tag(withPolicyChecked(ctx), desc, reference)
 }
 
+// Untag removes the association between the given tag and the manifest it
+// currently points to.
+//
+// Tag deletion is an optional capability per the OCI Distribution Spec
+// and is not supported by all registries.
```
TerryHowe:
```suggestion
// Untag removes the association between the given tag and the manifest it
// currently points to. Only tags are accepted; to delete a manifest by
// digest, use Delete.
//
// Tag deletion is an optional, loosely specified capability. Some registries
// do not support it (returning 400 or 405, surfaced here as an error), and
// some delete the underlying manifest — and thus every tag pointing to it —
// rather than only the given tag. This differs from the local content/oci
// store's Untag, which never deletes the underlying content.
```

TerryHowe (dismissed):

TerryHowe:
Any chance you could increase test coverage on this. Also, I committed that comment change, so you'll have to git pull that branch. Thanks

matthiasbruns:
Will take care of this tomorrow - so I assume the design is fine :D

registry/remote/repository.go
```diff
@@ -428,6 +428,35 @@ func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference
...
+// do not support it (returning 400 or 405, surfaced here as an error), and
+// some delete the underlying manifest — and thus every tag pointing to it —
+// rather than only the given tag. This differs from the local content/oci
+// store's Untag, which never deletes the underlying content.
+//
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#deleting-tags
+func (r *Repository) Untag(ctx context.Context, reference string) error {
+	if reference == "" {
+		return errdef.ErrMissingReference
+	}
```
TerryHowe:
```suggestion
```

Seems to be more consistent with this package.

TerryHowe (changes requested):
I would think more test coverage would be possible.

registry/remote/repository_test.go
```diff
@@ -1006,6 +1006,81 @@ func TestRepository_Tag(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Repository.Untag() error = %v", err)
+	}
+	if !untagged {
+		t.Errorf("Repository.Untag() did not send DELETE request")
+	}
+
+	// closing the server makes the underlying request fail
+	ts.Close()
+	err = repo.Untag(ctx, ref)
```
TerryHowe: Please use a cancelled context instead of closing the server will make the tests run faster. Something like
```
canceledCtx, cancel := context.WithCancel(ctx)
cancel()
err = repo.Untag(canceledCtx, ref)
```

TerryHowe (changes requested):
Looking very close, thanks! One comment to speed up tests.

TerryHowe:
conflict

matthiasbruns:
Was on vacation - will take care of this today

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1223, opened 2026-06-21
fix(content): bound ReadAll allocation by content, not descriptor size

TerryHowe:
## What this fixes

`content.ReadAll` currently rejects any descriptor with `Size > 32 MiB`
(`maxDescriptorSize`, added in #1153 for GHSA-f36w-mj3v-6jqv). That cap
over-corrected: `ReadAll` is also used by the in-memory store's `Push`
(`internal/cas/memory.go`) and by `FetchAll`/`FetchBytes`, so it now
rejects legitimate, digest-verified blobs. This broke OCI pulls of large
artifacts — e.g. installing the helm-unittest plugin (~160 MiB layer):

```
Error: failed to pull plugin from oci://ghcr.io/helm-unittest/helm-unittest/unittest:1.1.1: failed to perform "Push" on destination: invalid descriptor size
```

See helm/helm#32247.

## Root cause

The underlying vulnerability (#1153) was eagerly pre-allocating
`make([]byte, desc.Size)` from an attacker-controllable size, which panics
(`makeslice: len out of range`) on a forged value such as `2^62`. The
32 MiB rejection treated the symptom (size) rather than the cause
(trusting the size for allocation).

## Fix

Cap only the **initial** buffer allocation and grow the buffer as content
is read. The `VerifyReader` (LimitedReader + digest verifier) continues to
enforce the declared size and digest, so:

- a forged size now fails verification (`io.ErrUnexpectedEOF`) instead of
  panicking, and
- legitimately large, digest-verified content is read in full.

The manifest/metadata read paths remain independently bounded by
`limitSize(..., MaxMetadataBytes)` before calling `ReadAll`
(`registry/remote/repository.go`), so removing the blanket cap does not
reopen the metadata DoS that #1153 addressed.

## Tests

- `TestReadAll_OversizedDescriptor` updated: a `2^62` descriptor now fails
  with `io.ErrUnexpectedEOF` (no panic), still guarding GHSA-f36w-mj3v-6jqv.
- `TestReadAll_LargeContent` added: a 33 MiB blob (over the old cap)
  round-trips through `ReadAll`.

### DIFF
--- a/content/reader.go
+++ b/content/reader.go
@@ -16,6 +16,7 @@ limitations under the License.
 package content
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

PR #1224, opened 2026-06-21
fix(content): bound ReadAll allocation by content, not descriptor size

TerryHowe:
## What this fixes

`content.ReadAll` allocates `make([]byte, desc.Size)` up front, trusting a
size that is attacker-controllable. A crafted OCI layout `index.json` can
declare an enormous `Size` (e.g. `2^62`), panicking the process
(`makeslice: len out of range`) before any content is read
(GHSA-f36w-mj3v-6jqv).

Note: the `v2` branch received a 32 MiB rejection for this in #1153, but
that fix was never ported to `main`, so `main` is still exposed to the
panic. This PR fixes it directly.

## Fix

Cap only the **initial** buffer allocation and grow the buffer as content
is read. The `VerifyReader` (LimitedReader + digest verifier) continues to
enforce the declared size and digest, so a forged size now fails
verification (`io.ErrUnexpectedEOF`) instead of panicking, while
legitimately large, digest-verified content is read in full.

This deliberately avoids the blanket 32 MiB rejection used on `v2`, which
over-corrected and broke OCI pulls of large artifacts such as plugin and
chart layers (helm/helm#32247). The manifest/metadata read paths remain
independently bounded by `limitSize(..., MaxMetadataBytes)` before calling
`ReadAll`, so this does not reopen the metadata DoS.

The companion `v2` PR is #1223; both branches converge on the same
`ReadAll` implementation.

## Tests

- `TestReadAll_OversizedDescriptor` added: a `2^62` descriptor fails with
  `io.ErrUnexpectedEOF` (no panic), guarding GHSA-f36w-mj3v-6jqv.
- `TestReadAll_LargeContent` added: a 33 MiB blob round-trips through
  `ReadAll`.

### DIFF
--- a/content/reader.go
+++ b/content/reader.go
@@ -16,6 +16,7 @@ limitations under the License.
 package content
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


shizhMSFT (approved):
LGTM. Nice catch.

PR #1187, opened 2026-05-23
feat(policy): tighten scope matching and refine validation errors

TerryHowe:
## What

Small, self-contained updates to the existing `registry/remote/policy` package.

**`policy.go`**
- Add `ErrNoPolicyFound` sentinel error; `GetDefaultPolicyPath` now wraps it with `%w` so callers can branch on `errors.Is(err, policy.ErrNoPolicyFound)`.
- Restrict longest-prefix and wildcard subdomain scope matching in `GetRequirementsForImage` to the docker transport. Other transports now go straight from exact-match → transport default → global default, matching the containers-policy.json spec more precisely.
- Refactor `isPathPrefix` to `(prefix, s)` and simplify the body.

**`requirement.go`**
- Split the "exactly one of `keyPath`/`keyPaths`/`keyData`" check in `PRSignedBy.Validate` into distinct messages for the zero-provided and multiple-provided cases.
- Tighten `Fulcio` validation error messages.
- Document `KeyDatas` as base64-encoded.

## Why

Part of the v3 PR-by-PR breakdown (PR 10: `feat/policy-updates`). Self-contained updates to the existing policy package; later PRs (mirror middleware, builder, signature) build on this.

## Test plan

- [x] `go build -mod=mod ./registry/remote/policy/...`
- [x] `go vet -mod=mod ./registry/remote/policy/...`
- [x] `go test -mod=mod ./registry/remote/policy/...` — passes (`ok ... 0.397s`)
- [x] `policy_test.go` / `requirement_test.go` updated to cover the new docker-only matching scope, the wrapped `ErrNoPolicyFound`, and the split validation messages

PR #1226, opened 2026-06-23
feat(remote): add repository middleware with policy enforcement

TerryHowe:
## What

Adds a repository middleware mechanism to `registry/remote`:

- `RepositoryMiddleware` — a `func(registry.Repository) registry.Repository`
  decorator type
- `Compose(...)` — chains multiple middlewares into one
- `WithPolicyEnforcement(evaluator, transport, scope)` — wraps a repository so
  `Fetch`/`Resolve`/`FetchReference` (and the `Blobs()` / `Manifests()`
  sub-stores) are gated by a `policy.Evaluator`

## Why standalone

Self-contained: depends only on the already-merged `registry` interfaces and
`registry/remote/policy` (#1187). No dependency on the in-flight config,
builder, or repository PRs.

Part of the **feat/everything (#1130)** v3 breakdown. The companion mirror
fallback logic (originally bundled with middleware) lands separately, since it
depends on the client builder and repository-struct updates.

## Testing

`go build ./...` and `go test ./registry/remote/ -run 'Middleware|Compose|Policy'`
pass.

### DIFF
--- a/registry/remote/middleware.go
+++ b/registry/remote/middleware.go
@@ -1,0 +1,225 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package remote
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/oras-project/oras-go/v3/registry"
+	"github.com/oras-project/oras-go/v3/registry/remote/policy"
+)
+
+// RepositoryMiddleware wraps a registry.Repository to add cross-cutting concerns.
+type RepositoryMiddleware func(registry.Repository) registry.Repository
+
+// Compose chains multiple middlewares together.
+// The first middleware is the outermost (executed first for requests,
+// last for responses).
+func Compose(middlewares ...RepositoryMiddleware) RepositoryMiddleware {
+	return func(repo registry.Repository) registry.Repository {
+		for i := len(middlewares) - 1; i >= 0; i-- {
+			repo = middlewares[i](repo)
+		}
+		return repo
+	}
+}
+
+// WithPolicyEnforcement creates a middleware that adds policy checks to all operations.
+// The transport and scope parameters are used for constructing image references
+// for policy evaluation.
+func WithPolicyEnforcement(evaluator *policy.Evaluator, transport policy.TransportName, scope string) RepositoryMiddleware {
+	return func(repo registry.Repository) registry.Repository {
+		return &policyEnforcingRepository{
+			Repository: repo,
+			evaluator:  evaluator,
+			transport:  transport,
+			scope:      scope,
+		}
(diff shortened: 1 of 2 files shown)


registry/remote/middleware.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		return repo
+	}
+}
+
+// WithPolicyEnforcement creates a middleware that adds policy checks to all operations.
+// The transport and scope parameters are used for constructing image references
+// for policy evaluation.
+func WithPolicyEnforcement(evaluator *policy.Evaluator, transport policy.TransportName, scope string) RepositoryMiddleware {
+	return func(repo registry.Repository) registry.Repository {
```
shizhMSFT, line 56: Should we check `evaluator` nil here? If `evaluator == nil`, should we just return a no-op middleware so that we don't need to check every time?

registry/remote/middleware.go
```diff
@@ -0,0 +1,225 @@
...
+			evaluator:  evaluator,
+			transport:  transport,
+			scope:      scope,
+		}
+	}
+}
+
+// policyEnforcingRepository wraps a Repository and enforces policy on all operations.
+type policyEnforcingRepository struct {
+	registry.Repository
```
shizhMSFT: `registry.Repository` or named field? I'm worried if there is a new method for Repository someday and this policyEnforcer does not notice that. It is unlikely in a published major version, but it is possible for a new major version.

shizhMSFT: Same for other types.

registry/remote/middleware.go
TerryHowe: Done. `WithPolicyEnforcement` now returns a no-op middleware (the identity function) when `evaluator == nil`, and the per-call nil guard in `checkPolicy` is removed.

registry/remote/middleware.go
TerryHowe: Done. Switched all three wrappers (`policyEnforcingRepository`, `policyEnforcingBlobStore`, `policyEnforcingManifestStore`) from embedding the interface to a named field, and implemented every method explicitly. Now a new method added to `registry.Repository`/`BlobStore`/`ManifestStore` fails to compile here instead of silently passing through unenforced.

registry/remote/middleware.go
```diff
@@ -0,0 +1,299 @@
...
+
+// Exists returns true if the described content exists.
+func (r *policyEnforcingRepository) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+	return r.repo.Exists(ctx, target)
+}
+
+// Delete removes the content identified by the descriptor.
+func (r *policyEnforcingRepository) Delete(ctx context.Context, target ocispec.Descriptor) error {
+	return r.repo.Delete(ctx, target)
+}
```
shizhMSFT, line 122: Question: Do those operations need policy check?

shizhMSFT: so as other newly added functions.

shizhMSFT (approved):

shizhMSFT (approved):
LGTM with questions above

registry/remote/middleware.go
TerryHowe: Created https://github.com/oras-project/oras-go/issues/1265 to track this

TerryHowe:
Filed #1265 to track @shizhMSFT's question about the ungated operations (`Delete`/`Exists`/`Tags`/`Referrers`).

Proposed resolution there is the **admission-over-content** model: gate `Exists` (same content-read shape as `Fetch`), and keep `Delete`, `Referrers`, and `Tags` as intentional, documented carve-outs — since policy here is a content-trust check, not authorization. `Referrers` in particular shouldn't be gated because signature/attestation discovery (used by verification itself) relies on it. The overstated "no operation silently bypasses policy enforcement" doc comment is corrected there too.

PR #1232, opened 2026-07-02
fix: prevent hardlink path traversal via process CWD (GHSA-fxhp-mv3v-67qp)

TerryHowe:
## Summary

Backport of the `main`-branch fix (b11f777) to the **v2** release line for **GHSA-fxhp-mv3v-67qp / CVE-2026-50163**.

`ensureLinkPath` validated `TypeLink` (hardlink) targets by resolving relative paths against the link file's directory, but returned the *original, unresolved* target string. The caller passed this to `os.Link`, which resolves relative paths against the process **CWD** via `link(2)` — not the link file's directory. A crafted OCI artifact could exploit this to hardlink a CWD file into the extract tree, creating a shared inode with an arbitrary file outside the extraction directory.

The fix landed on `main` (now the v3 module) but was never carried to the v2 line, so v2.6.1 and earlier remain vulnerable. This PR closes that gap.

## Fix

Explicitly resolve relative hardlink targets against `filepath.Dir` of the link file before calling `os.Link`. Symlink handling is unaffected since `os.Symlink` stores the path verbatim and resolves at access time.

## Test

Adds a regression test that sets the process CWD to a directory containing a sentinel file and verifies that a tar `TypeLink` entry with a relative `Linkname` cannot hardlink that file into the extract tree. The test fails without the fix ("hardlink escaped extract dir") and passes with it.

Suggested release: **v2.6.2**.

### DIFF
--- a/content/file/utils.go
+++ b/content/file/utils.go
@@ -188,6 +188,11 @@ func extractTarDirectory(dirPath, dirName string, r io.Reader, buf []byte, prese
 			// This is a known limitation and will not be addressed.
 			var target string
 			if target, err = ensureLinkPath(dirPath, dirName, filePath, header.Linkname); err == nil {
+				if !filepath.IsAbs(target) {
+					// link(2) resolves relative paths against the process CWD, not
+					// the link file's directory. Resolve explicitly to prevent escape.
+					target = filepath.Join(filepath.Dir(filePath), target)
+				}
 				err = os.Link(target, filePath)
 			}
 		case tar.TypeSymlink:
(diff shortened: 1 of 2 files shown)


shizhMSFT (approved):
LGTM

buzzlighty:
@TerryHowe Is there a timeline for v2.7.0 release? I know you're mostly maintaining this alone, so no pressure 😄

TerryHowe:
> @TerryHowe Is there a timeline for v2.7.0 release? I know you're mostly maintaining this alone, so no pressure 😄

As soon as https://github.com/oras-project/oras-go/pull/1235 is approved

PR #1237, opened 2026-07-08
fix: validate caller-supplied config descriptor in Pack/PackManifest

### LINKED ISSUE #1236: Push to github container registry missing `config` annotations
Very well might be doing something wrong / misinterpreting the API, but when I push an OCI artifact to github container registry using this for the config descriptor

```go
ConfigDescriptor: &ocispec.Descriptor{
			MediaType: "application/vnd.custom.config.v1+json",
			Annotations: map[string]string{
				"git-commit-hash": gitHash,
			},
		},
```

the media type shows up properly, but not the annotations. I expected to see an annotation for `"git-commit-hash": gitHash`. 

```
"config": {
    "digest": "",
    "mediaType": "application/vnd.custom.config.v1+json",
    "size": 0
  }
```

The full `PackManifest` call is as follows. These are mainly dummy values for testing.

```go
	desc, err := oras.PackManifest(ctx, repo, oras.PackManifestVersion1_1, "application/octet-stream", oras.PackManifestOptions{
		Layers: layers,
		ManifestAnnotations: map[string]string{
			"org.opencontainers.image.source": gitRemote,
		},
		ConfigDescriptor: &ocispec.Descriptor{
			MediaType: "application/vnd.custom.config.v1+json",
			Annotations: map[string]string{
				"sal.git-commit-hash": gitHash,
			},
		},
	})
```

I am using the following dependency and pushing to the standard github container registry (no special enterprise / self hosted registry)

```
oras.land/oras-go/v2 v2.6.1
```

From researching the docs it seemed like maybe Github container registry just wants all annotations in the block like this? These ManifestAnnotations definitely show up fine without issue

```go
		ManifestAnnotations: map[string]string{
			"org.opencontainers.image.source": gitRemote,
		},
```

TerryHowe:
## What this PR does

When a `ConfigDescriptor` is supplied to `Pack` / `PackManifest`, the caller is responsible for pushing the config blob and populating the descriptor's `Digest` and `Size`. The library uses the descriptor verbatim and does **not** push the config blob in that path.

Previously, a partially-populated descriptor (e.g. only `MediaType` and `Annotations` set, with an empty `Digest` and zero `Size`) was accepted and produced a manifest referencing a non-existent config blob:

```json
"config": {
  "digest": "",
  "mediaType": "application/vnd.custom.config.v1+json",
  "size": 0
}
```

Some registries (e.g. GHCR) accept such manifests while silently dropping the config descriptor's annotations, which is confusing to debug.

## Change

- Add `validateConfigDescriptor`, which validates both the media type format and the digest of a caller-supplied config descriptor.
- Apply it in `packManifestV1_0`, `packManifestV1_1`, and the deprecated `packManifestV1_1_RC2` paths. Callers now get an error wrapping `errdef.ErrInvalidDigest` instead of a silently-malformed manifest.
- Clarify the `ConfigDescriptor` godoc to state that the caller must push the config blob and populate `Digest`/`Size`.

Because this only rejects manifests that were already malformed, and oras-go v3 has not been released yet, this is not a breaking change.

## Tests

Added `Test_PackManifest_ImageV1_0_InvalidConfigDigest`, `Test_PackManifest_ImageV1_1_InvalidConfigDigest`, and `Test_Pack_ImageV1_1_RC2_InvalidConfigDigest`. Full suite (`go test ./...`) passes.

Fixes #1236

### DIFF
--- a/pack.go
+++ b/pack.go
@@ -101,6 +101,12 @@ type PackManifestOptions struct {
 
 	// ConfigDescriptor is a pointer to the descriptor of the config blob.
 	// If not nil, ConfigAnnotations will be ignored.
+	//
+	// When set, the caller is responsible for pushing the referenced config
+	// blob to the target and for populating the descriptor's Digest and Size.
+	// PackManifest uses the descriptor as-is and does NOT push the config blob.
+	// The descriptor's Digest MUST be valid, otherwise PackManifest returns an
+	// error wrapping [errdef.ErrInvalidDigest].
 	ConfigDescriptor *ocispec.Descriptor
 
 	// ConfigAnnotations is the annotation map of the config descriptor.
@@ -225,8 +231,8 @@ func packManifestV1_0(ctx context.Context, pusher content.Pusher, artifactType s
 	// prepare config
 	var configDesc ocispec.Descriptor
 	if opts.ConfigDescriptor != nil {
-		if err := validateMediaType(opts.ConfigDescriptor.MediaType); err != nil {
-			return ocispec.Descriptor{}, fmt.Errorf("invalid config mediaType format: %w", err)
+		if err := validateConfigDescriptor(*opts.ConfigDescriptor); err != nil {
+			return ocispec.Descriptor{}, err
 		}
 		configDesc = *opts.ConfigDescriptor
 	} else {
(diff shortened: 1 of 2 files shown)


TerryHowe:
Fixed by #1269

PR #1095, opened 2026-01-13
fix: graph.Memory should use digest as map key

wangxiaoxuan273:
Should fix #1093

PR #1260, opened 2026-07-31
fix: graph.Memory should use digest as map key

### LINKED ISSUE #1093: Local OCI store garbage collection can misidentify dangling blobs and remove them erroneously
### Description
Background: I currently work on the [KitOps project](https://github.com/kitops-ml/kitops), which uses the oras-go library to manage storing OCI artifacts. In our use case, we use different mediatypes for blob layers to distinguish between types of data in an artifact, though most blobs are packaged in the same way (tarballs of files).

The in-memory graph of a local OCI store stores predecessor and successor maps of descriptors to represent the graph of references. This graph is used in garbage collection in order to prune dangling blobs when e.g. a manifest is removed. 

However, the graph stores a map of descriptors while the underlying CAS store on disk only uses digests. If two manifests refer to the same blob digest with a different media type, garbage collection will find the blob to be dangling when either manifest is deleted, rendering the other manifest invalid.

### Reproducer
This reproducer uses OCI artifacts I generated with the Kit CLI, tagged `sample-one` and `sample-two` here: https://quay.io/repository/amisevsk0/oras-cli?tab=tags. The artifact tagged `sample-one` has layers

```json
  "layers": [
    {
      "mediaType": "application/vnd.kitops.modelkit.docs.v1.tar",
      "digest": "sha256:65c705d810aa0ec82d1ea886c170462c9f17203aa313bfc2cf01d099d4e5ff96",
      "size": 2048,
    }
  ],
```
whereas the artifact tagged `sample-two` has layers
```json
 "layers": [
    {
      "mediaType": "application/vnd.kitops.modelkit.model.v1.tar",
      "digest": "sha256:65c705d810aa0ec82d1ea886c170462c9f17203aa313bfc2cf01d099d4e5ff96",
      "size": 2048,
    }
  ],
```
These two artifacts refer to the same blob (with digest `sha256:65c705d810aa0ec82d1ea886c170462c9f17203aa313bfc2cf01d099d4e5ff96`) but use a different mediaType. The underlying file is just a text file containing the text `testfile`. Below is a Go program that uses the oras-go library to fetch these artifacts to a local OCI store, deletes one of them, and then attempts to copy the the other artifact to another OCI store, which fails (as an example -- anything that requires reading the second artifacts blobs will similarly fail).

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"path/filepath"

	"oras.land/oras-go/v2"
	"oras.land/oras-go/v2/content/oci"
	"oras.land/oras-go/v2/registry/remote"
)

func dieIfErr(err error) {
	if err != nil {
		log.Fatalf("Unexpected error: %s\n", err)
	}
}

func main() {
	ctx := context.Background()
	tmpDir, err := os.MkdirTemp("", "oras-remove-reproducer-*")
	dieIfErr(err)
	storePathA := filepath.Join(tmpDir, "storage-a")
	storePathB := filepath.Join(tmpDir, "storage-b")

	log.Printf("Setting up local OCI stores in %s and %s\n", storePathA, storePathB)
	localStoreA, err := oci.New(storePathA)
	dieIfErr(err)
	localStoreB, err := oci.New(storePathB)
	dieIfErr(err)

	log.Println("Setting up remote repository to fetch reproducer artifacts")
	remoteRegistry, err := remote.NewRegistry("quay.io")
	dieIfErr(err)
	remoteRepo, err := remoteRegistry.Repository(ctx, "amisevsk0/oras-cli")
	dieIfErr(err)

	log.Println("Copying reproducer artifacts to local OCI store A")
	sampleOneDesc, err := oras.Copy(ctx, remoteRepo, "sample-one", localStoreA, "sample-one", oras.DefaultCopyOptions)
	dieIfErr(err)
	_, err = oras.Copy(ctx, remoteRepo, "sample-two", localStoreA, "sample-two", oras.DefaultCopyOptions)
	dieIfErr(err)

	log.Println("Removing sample-one tagged manifest from local store A")
	err = localStoreA.Delete(ctx, sampleOneDesc)
	dieIfErr(err)

	log.Println("Copying sample-two tagged manifest from local store A to local store B")
	_, err = oras.Copy(ctx, localStoreA, "sample-two", localStoreB, "sample-two", oras.DefaultCopyOptions)
	if err != nil {
		fmt.Printf("Unable to copy 'sample-two' to another OCI store: %s\n", err)
		os.Exit(1)
	}

	log.Println("Error did not reproduce")
}
```

The above file fails with 

```
Unable to copy 'sample-two' to another OCI store: failed to perform "Fetch" on source: sha256:65c705d810aa0ec82d1ea886c170462c9f17203aa313bfc2cf01d099d4e5ff96: application/vnd.kitops.modelkit.model.v1.tar: not found
```

as the blob (digest `sha256:65c705d8...`) was deleted by garbage collection in the earlier `Delete` call.

### LINKED ISSUE #1258: Forward-port #1095 to main: graph.Memory should use digest as map key (GC can prune shared blobs)
### Description

The garbage-collection correctness bug fixed by #1095 (which resolved #1093) on the `v2` branch is **still present on `main` (v3)** and needs to be forward-ported.

In `internal/graph/memory.go`, the in-memory graph keys its `nodes`, `predecessors`, and `successors` maps on `descriptor.Descriptor`:

```go
nodes        map[descriptor.Descriptor]ocispec.Descriptor
predecessors map[descriptor.Descriptor]set.Set[descriptor.Descriptor]
successors   map[descriptor.Descriptor]set.Set[descriptor.Descriptor]
```

Because `internal/descriptor.Descriptor` includes `MediaType` and `Size` in addition to `Digest`, two manifests that reference the same blob **digest** with a different **media type** are treated as distinct graph nodes. The underlying CAS store on disk is keyed purely by digest, so when one manifest is deleted, garbage collection misidentifies the shared blob as dangling and prunes it — invalidating the other manifest.

This is the same defect described in #1093 (reproducer there uses KitOps artifacts sharing a blob digest under different media types).

### Fix

Port #1095 to `main`: key the graph maps on `digest.Digest` instead of `descriptor.Descriptor`, matching what the on-disk CAS uses.

```go
nodes        map[digest.Digest]ocispec.Descriptor
predecessors map[digest.Digest]set.Set[digest.Digest]
successors   map[digest.Digest]set.Set[digest.Digest]
```

### References

- v2 fix: #1095 (merged into `v2`, released in v2.6.1 / v2.6.2)
- Original bug report: #1093
- Affected file: `internal/graph/memory.go`

### Status on branches

| branch | map key | fixed? |
| --- | --- | --- |
| `v2` | `digest.Digest` | ✅ (via #1095) |
| `main` (v3) | `descriptor.Descriptor` | ❌ not yet ported |

DoTuanAnh2k1:
## Description

Forward-port of #1095 (which fixed #1093) from `v2` to `main` (v3). Fixes #1258.

`internal/graph/memory.go` keyed its `nodes`, `predecessors`, and `successors` maps on `internal/descriptor.Descriptor`, which includes `MediaType` and `Size` in addition to `Digest`. Two manifests that reference the same blob **digest** under a different **media type** were therefore treated as distinct graph nodes, while the on-disk CAS is keyed purely by digest. When one manifest is removed, garbage collection misidentifies the shared blob as dangling and prunes it — invalidating the other manifest (the KitOps reproducer in #1093).

## Fix

Key the graph maps on `digest.Digest` to match the on-disk CAS, mirroring the merged v2 fix (#1095):

- Drop the `internal/descriptor` import; use `node.Digest` / `successor.Digest` instead of `descriptor.FromOCI(...)`.
- `nodes`, `predecessors`, `successors` are now `map[digest.Digest]...`.
- `memory_test.go` adapted to key on `desc*.Digest`.

## Regression test

Beyond the faithful forward-port, this adds `TestMemory_SharedBlobDigestUnderDifferentMediaType`: it builds two real manifests that reference the same blob digest under different layer media types and asserts the shared blob stays a single digest-keyed node — in particular that removing one manifest does **not** report the still-referenced blob as dangling. The test **fails** on the pre-fix (descriptor-keyed) code and passes on the fix, so it directly guards the #1093 GC regression. (Happy to drop it if you'd prefer strict parity with #1095.)

## Testing

- `go test ./...` passes.
- `make check-encoding` passes.
- Verified the new regression test fails on the pre-fix commit and passes on `HEAD`.

No public API change.

---
<sub>Disclosure: prepared with AI assistance (Claude Code); I reviewed the diff and verified the tests (including fail-on-pre-fix) locally.</sub>

### DIFF
--- a/internal/graph/memory.go
+++ b/internal/graph/memory.go
@@ -25,7 +25,6 @@ import (
 	"github.com/oras-project/oras-go/v3/content"
 	"github.com/oras-project/oras-go/v3/errdef"
 	"github.com/oras-project/oras-go/v3/internal/container/set"
-	"github.com/oras-project/oras-go/v3/internal/descriptor"
 	"github.com/oras-project/oras-go/v3/internal/status"
 	"github.com/oras-project/oras-go/v3/internal/syncutil"
 )
@@ -34,9 +33,9 @@ import (
 type Memory struct {
 	// nodes has the following properties and behaviors:
 	//  1. a node exists in Memory.nodes if and only if it exists in the memory
-	//  2. Memory.nodes saves the ocispec.Descriptor map keys, which are used by
+	//  2. Memory.nodes saves the ocispec.Descriptor indexed by digest, which are used by
 	//    the other fields.
-	nodes map[descriptor.Descriptor]ocispec.Descriptor
+	nodes map[digest.Digest]ocispec.Descriptor
 
 	// predecessors has the following properties and behaviors:
 	//  1. a node exists in Memory.predecessors if it has at least one predecessor
@@ -44,14 +43,14 @@ type Memory struct {
 	//    the memory.
 	//  2. a node does not exist in Memory.predecessors, if it doesn't have any predecessors
 	//    in the memory.
-	predecessors map[descriptor.Descriptor]set.Set[descriptor.Descriptor]
+	predecessors map[digest.Digest]set.Set[digest.Digest]
 
 	// successors has the following properties and behaviors:
 	//  1. a node exists in Memory.successors if and only if it exists in the memory.
 	//  2. a node's entry in Memory.successors is always consistent with the actual
 	//    content of the node, regardless of whether or not each successor exists
 	//    in the memory.
-	successors map[descriptor.Descriptor]set.Set[descriptor.Descriptor]
+	successors map[digest.Digest]set.Set[digest.Digest]
 
 	lock sync.RWMutex
 }
@@ -59,9 +58,9 @@ type Memory struct {
 // NewMemory creates a new memory PredecessorFinder.
 func NewMemory() *Memory {
 	return &Memory{
-		nodes:        make(map[descriptor.Descriptor]ocispec.Descriptor),
-		predecessors: make(map[descriptor.Descriptor]set.Set[descriptor.Descriptor]),
-		successors:   make(map[descriptor.Descriptor]set.Set[descriptor.Descriptor]),
+		nodes:        make(map[digest.Digest]ocispec.Descriptor),
+		predecessors: make(map[digest.Digest]set.Set[digest.Digest]),
+		successors:   make(map[digest.Digest]set.Set[digest.Digest]),
 	}
 }
 
@@ -108,14 +107,13 @@ func (m *Memory) Predecessors(_ context.Context, node ocispec.Descriptor) ([]oci
 	m.lock.RLock()
 	defer m.lock.RUnlock()
 
-	key := descriptor.FromOCI(node)
-	set, exists := m.predecessors[key]
+	set, exists := m.predecessors[node.Digest]
 	if !exists {
 		return nil, nil
 	}
 	var res []ocispec.Descriptor
-	for k := range set {
-		res = append(res, m.nodes[k])
+	for digest := range set {
+		res = append(res, m.nodes[digest])
 	}
 	return res, nil
 }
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

PR #1264, opened 2026-07-31
ci: harden release version handling

### LINKED ISSUE #1261: Release workflow: unvalidated version interpolated into shell before semver check (command injection)
## Summary

The `release.yml` workflow added in #1259 (forward-porting #1161) interpolates an attacker-influenced value directly into a shell `run:` block in the **Validate semantic version** step, *before* that value is validated. This is a GitHub Actions script-injection sink in a job that holds `contents: write`.

## Location

`.github/workflows/release.yml`, "Validate semantic version" step:

```yaml
- name: Validate semantic version
  run: |
    VERSION="${{ steps.version.outputs.version }}"
    if ! [[ "$VERSION" =~ ^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(alpha|beta|rc)\.[0-9]+)?$ ]]; then
      ...
```

`steps.version.outputs.version` derives from `github.head_ref` (the release branch name). GitHub Actions substitutes `${{ }}` expressions into the script text at render time, so the regex gate runs *after* substitution — it cannot protect the very step that first interpolates the value.

## Impact

A branch name such as `release/v0.0.0";id;"` renders to:

```bash
VERSION="v0.0.0";id;""
```

and `id` (or any command) executes before the semver regex ever runs, in a job with `contents: write` (tag push + release publication).

Git ref names permit `"`, `;`, `$`, backtick, `(`, `)`, and `|`. Spaces are disallowed, but `${IFS}` substitutes for them, so a working payload is constructible.

## Mitigating factors (why not critical)

- The job `if:` guard restricts triggering to same-repo branches (`github.event.pull_request.head.repo.full_name == github.repository`), so forks cannot trigger it.
- Merging a `release/*` PR into the release branch requires branch-protection approval (3 of 4 owners).

So exploitation requires repo write access plus a merged release PR. It is defense-in-depth, but the fix is trivial and this contradicts the stated hardening goal of #1259 ("pass the release branch through the step environment to avoid direct untrusted-context interpolation") — that hardening was applied to the *Extract version* step but not to this one.

## Suggested fix

Pass the value through `env:` so it never enters the shell as literal script text:

```yaml
- name: Validate semantic version
  env:
    VERSION: ${{ steps.version.outputs.version }}
  run: |
    if ! [[ "$VERSION" =~ ^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(alpha|beta|rc)\.[0-9]+)?$ ]]; then
      echo "Error: '$VERSION' does not follow semantic versioning"
      exit 1
    fi
```

For consistency and defense-in-depth, the "Create and push tag" step should do the same (it interpolates `${{ steps.version.outputs.version }}` directly too, though it is currently gated by the validation step running first).

## Notes

The same pattern exists on the `v2` branch (introduced by #1161), so a fix there is likely warranted as well.

lntutor:
## Summary

- pass the branch-derived release version through step environment variables instead of rendering it into shell source
- quote the version and merge commit when creating and pushing the release tag
- quote `GITHUB_OUTPUT` so the hardened workflow passes Actionlint

Fixes #1261

## Test plan

- `actionlint .github/workflows/release.yml`
- valid semantic-version shell regression check
- malicious branch-version shell regression check confirming no command execution
- `make test`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,11 +44,12 @@ jobs:
           BRANCH: ${{ github.head_ref }}
         run: |
           VERSION="${BRANCH#release/}"
-          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
 
       - name: Validate semantic version
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
         run: |
-          VERSION="${{ steps.version.outputs.version }}"
           if ! [[ "$VERSION" =~ ^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(alpha|beta|rc)\.[0-9]+)?$ ]]; then
             echo "Error: '$VERSION' does not follow semantic versioning"
             echo "Expected: vX.Y.Z or vX.Y.Z-(alpha|beta|rc).N"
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

PR #1267, opened 2026-07-31
fix: enforce policy for exists operations

### LINKED ISSUE #1265: policyEnforcingRepository: gate Exists and document Delete/Tags/Referrers carve-outs (admission-over-content)
## Summary

`policyEnforcingRepository` (added in #1226, `registry/remote/middleware.go`) gates
`Fetch`, `Push`, `Resolve`, `Tag`, `FetchReference`, and `PushReference` through the
`policy.Evaluator`, but leaves **`Delete`, `Exists`, `Tags`, and `Referrers`** ungated.

This was raised in review by @shizhMSFT:
https://github.com/oras-project/oras-go/pull/1226#discussion_r3687113946
("Question: Do those operations need policy check?", and the follow-up
https://github.com/oras-project/oras-go/pull/1226#discussion_r3687115548).

The type's doc comment also currently claims *"no operation silently bypasses policy
enforcement"*, which is not accurate while those four methods pass through unchecked.

## Decision: admission-over-content model

`policy.Evaluator.IsImageAllowed` is a **content-admission / trust** check
(containers/image-style accept/reject + signature verification). It answers *"should I
trust this image content?"* — it is **not** an authorization/RBAC layer answering *"may
this caller touch this repo?"*. Under that model the four operations are not equivalent,
and gating all of them would be wrong for two of them:

| Operation | Gate? | Rationale |
|-----------|-------|-----------|
| `Exists`    | **Yes** | Same shape as `Fetch` (has `target.Digest`); cheap; closes a presence-leak under a deny-all policy. Belongs on the content-read path. |
| `Referrers` | No  | Referrer discovery is how signatures/attestations are located — including by policy verification itself. Gating it on the same policy risks a chicken-and-egg that breaks verification. |
| `Tags`      | No  | Enumerates tag *names*, returns no image content — there is no image reference for an admission policy to evaluate. |
| `Delete`    | No  | Destructive removal. Gating on *trust* is backwards: operators must be able to remove exactly the content that **fails** policy. |

(If a stricter "fail-closed over all access" threat model is preferred instead, the
alternative is to also gate `Tags`/`Referrers` and keep only `Delete` carved out. This
issue proposes the admission-over-content model above.)

## Proposed change

1. **Gate `Exists`** on `policyEnforcingRepository` and both sub-stores
   (`policyEnforcingBlobStore`, `policyEnforcingManifestStore`), mirroring `Fetch`:

   ```go
   func (r *policyEnforcingRepository) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
       if err := r.checkPolicy(ctx, target.Digest.String()); err != nil {
           return false, err
       }
       return r.repo.Exists(ctx, target)
   }
   ```

2. **Document the intentional carve-outs** on `Delete`, `Referrers`, and `Tags` so they
   read as decisions rather than omissions (one-line rationale each, per the table above).

3. **Fix the type doc comment** to state the accurate guarantee: the named-field design
   forces a policy *decision* per method at compile time — not that every method is gated.

## Tests

- Add `Exists` (repository + blob store + manifest store) to
  `TestWithPolicyEnforcement_PolicyDenied_AllOperations`.
- Add a positive test asserting `Delete`, `Tags`, and `Referrers` still succeed **under a
  deny-all policy**, locking the carve-out decisions so a future change can't silently
  gate them.

## Notes

- A separate, related concern from the #1226 review is that `checkPolicy` passes an
  incomplete `Reference` (bare digest / bare tag, no repository) to the evaluator, which
  breaks real `SignedBy`/`Sigstore` verifiers. That is tracked/fixed independently and is
  out of scope for this issue.

lntutor:
## Summary

- enforce policy checks for repository, blob-store, and manifest-store `Exists` calls
- document why `Delete`, `Tags`, and `Referrers` intentionally bypass admission checks
- add deny-all regression coverage for gated operations and the intentional carve-outs

## Tests

```
go test ./registry/remote -count=1
```

322 tests passed.

Fixes #1265

### DIFF
--- a/registry/remote/middleware.go
+++ b/registry/remote/middleware.go
@@ -63,12 +63,13 @@ func WithPolicyEnforcement(evaluator *policy.Evaluator, transport policy.Transpo
 	}
 }
 
-// policyEnforcingRepository wraps a Repository and enforces policy on all operations.
+// policyEnforcingRepository wraps a Repository and makes an explicit policy
+// decision for every operation.
 //
 // The wrapped repository is held in a named field rather than embedded so that
 // every method of registry.Repository must be implemented explicitly. If a new
 // method is added to the interface, this type fails to compile until the method
-// is handled here, ensuring no operation silently bypasses policy enforcement.
+// is handled here, ensuring policy behavior is considered for every operation.
 type policyEnforcingRepository struct {
 	repo      registry.Repository
 	evaluator *policy.Evaluator
@@ -111,12 +112,15 @@ func (r *policyEnforcingRepository) Push(ctx context.Context, expected ocispec.D
 	return r.repo.Push(ctx, expected, content)
 }
 
-// Exists returns true if the described content exists.
+// Exists returns true if the described content exists with policy enforcement.
 func (r *policyEnforcingRepository) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
+	if err := r.checkPolicy(ctx, target.Digest.String()); err != nil {
+		return false, err
+	}
 	return r.repo.Exists(ctx, target)
 }
 
-// Delete removes the content identified by the descriptor.
+// Delete removes content without a policy check so denied content remains removable.
 func (r *policyEnforcingRepository) Delete(ctx context.Context, target ocispec.Descriptor) error {
 	return r.repo.Delete(ctx, target)
 }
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

PR #1272, opened 2026-08-01
fix: redact sensitive request data from debug logs

### LINKED ISSUE #1271: LoggingTransport leaks Cookie, Proxy-Authorization, and URL query strings in debug logs
### Summary

`LoggingTransport` (`registry/remote/logging_transport.go`, added in #1190) redacts only two headers before debug-logging HTTP requests/responses. When a user opts into this transport, several other sensitive values are still written in clear text. This is the source of code-scanning alert [#11](https://github.com/oras-project/oras-go/security/code-scanning/11) (`go/clear-text-logging`, high).

The alert itself is *partly* a false positive — `Authorization` and `Set-Cookie` **are** masked — but there are real residual gaps worth closing so the alert can be resolved legitimately rather than dismissed.

### Current behavior

```go
var sensitiveHeaders = []string{
    "Authorization",
    "Set-Cookie",
}
```

`formatHeaders` replaces the values of those keys with `*****`. `RoundTrip` also logs `req.URL` in full.

### Gaps

1. **`Cookie` request header is not redacted.** Only `Set-Cookie` (response) is masked; the request-side `Cookie` header can carry session credentials and is logged in clear text.
2. **`Proxy-Authorization` is not redacted.** Carries proxy credentials.
3. **Full URL is logged unredacted** (`"url", req.URL`). Some registries convey tokens in query parameters (e.g. pre-signed blob URLs), which would be leaked.

### Impact

Low severity and opt-in: only triggered when a consumer explicitly wraps their transport with `NewLoggingTransport` and enables `Debug`-level logging. No effect on default usage. But for anyone who enables debug logging against a registry that uses cookies, a proxy, or query-string tokens, credentials can end up in logs.

### Suggested fix

- Add `Cookie` and `Proxy-Authorization` to `sensitiveHeaders`.
- Redact query strings (or at least known credential-bearing params) from the logged URL, or log `req.URL.Redacted()` which masks userinfo.
- Add tests asserting these values are masked.

Resolving these in code will let code-scanning alert #11 auto-close as *fixed* on the next scan.

### References

- File: `registry/remote/logging_transport.go`
- Introduced in: #1190
- Code scanning alert: https://github.com/oras-project/oras-go/security/code-scanning/11

nightcityblade:
Fixes #1271

## Summary

- redact `Cookie` and `Proxy-Authorization` alongside the existing sensitive headers
- omit query strings and fragments from the URL written to debug logs without mutating the outgoing request
- add end-to-end regression coverage for header and query credential leakage

## Testing

- `go test ./registry/remote`
- `go test ./...`
- `go test -race ./registry/remote`

`go vet ./...` still reports the existing lock-copy warning in `registry/remote/auth.go:37`.

### DIFF
--- a/registry/remote/logging_transport.go
+++ b/registry/remote/logging_transport.go
@@ -29,6 +29,8 @@ import (
 // sensitiveHeaders is the set of headers whose values are scrubbed from logs.
 var sensitiveHeaders = []string{
 	"Authorization",
+	"Cookie",
+	"Proxy-Authorization",
 	"Set-Cookie",
 }
 
@@ -73,10 +75,15 @@ func NewLoggingTransport(inner http.RoundTripper, logger *slog.Logger) *LoggingT
 // RoundTrip implements http.RoundTripper, logging the request and response.
 func (t *LoggingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	id := requestCounter.Add(1) - 1
+	loggedURL := *req.URL
+	loggedURL.RawQuery = ""
+	loggedURL.ForceQuery = false
+	loggedURL.Fragment = ""
+	loggedURL.RawFragment = ""
 
 	t.logger.Debug(req.Method,
 		"id", id,
-		"url", req.URL,
+		"url", loggedURL.Redacted(),
 		"header", formatHeaders(req.Header),
 	)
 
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

PR #1280, opened 2026-08-02
feat(remote): add mirror fallback for read operations

TerryHowe:
## What

Adds registry mirror support to `remote.Repository`. When a repository has
one or more mirrors configured, read operations try each applicable mirror in
order before falling back to the primary registry:

- `Fetch`, `Exists` (blob reads, digest-based)
- `Resolve`, `FetchReference` (manifest reads, tag or digest)

Write operations (`Push`, `Delete`, `Tag`, `PushReference`, `Mount`) always
target the primary — mirrors are read-only.

## Details

- **`mirror.go`**
  - `mirrorRepository` — pairs a `*Repository` with its pull policy.
  - `PullFromMirror` matching — `all` / `digest-only` / `tag-only`, so a
    mirror can be restricted to (e.g.) digest pulls only.
  - `mirrorReference` — normalizes a possibly fully-qualified reference so it
    resolves against a mirror whose registry/repository differs from the
    primary's.
  - `withMirrorFallback{Fetch,Exists,Resolve,FetchReference}` — wrap each read,
    skipping non-applicable mirrors and treating context cancellation /
    deadline as non-retryable (no fallthrough).
- **`repository.go`** — an unexported `mirrors` field on `Repository` (copied
  in `clone()`) plus the fallback wiring in the four read methods.

## Scope / follow-up

The `mirrors` field is unexported and currently only settable internally. The
public constructor that populates it from a `properties.Registry`
(`NewRepositoryWithProperties` / the `ClientBuilder`) lands in a later PR in
the split, along with `buildMirrorRepositories` — the tests for those pieces
travel with that PR, not this one.

This change is **behavior-preserving** for existing callers: with no mirrors
configured, `len(r.mirrors) == 0` and every read goes straight to the primary,
exactly as today.

## Testing

- `go build ./...`
- `go test -race ./registry/remote/...` — passes, including the new
  `mirror_test.go` (fallback ordering, pull-policy matching, reference
  normalization, non-retryable context errors).
- `gofmt` and `go vet` clean (no new findings).

Part of the **feat/everything (#1130)** v3 breakdown.

### DIFF
--- a/registry/remote/mirror.go
+++ b/registry/remote/mirror.go
@@ -1,0 +1,206 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package remote
+
+import (
+	"context"
+	"io"
+	"strings"
+
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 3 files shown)


shizhMSFT (approved):
LGTM

PR #1282, opened 2026-08-02
fix: validate blob upload Location host

### LINKED ISSUE #1273: Forward-port GHSA-jxpm-75mh-9fp7 blob-upload credential-leak fix (#1152) to main (v3)
## Summary

The fix for **GHSA-jxpm-75mh-9fp7** (credential forwarding via an unvalidated `Location` header during blob upload, CWE-918 SSRF) landed on the `v2` branch in #1152 but was **never forward-ported to `main` (v3)**. The `main` branch is still vulnerable.

## Details

During the monolithic blob upload flow, `blobStore.completePushAfterInitialPost` (`registry/remote/repository.go`) follows the `Location` header returned by the registry's `POST` response and reuses the caller's `Authorization` header for the subsequent `PUT`. A malicious or compromised registry can return a cross-host `Location`, causing oras-go to forward the caller's credentials to an attacker-controlled endpoint. A scheme downgrade (`https` → `http`) is likewise not rejected.

On `main`, `completePushAfterInitialPost` proceeds straight from the port-443 workaround to `url := location.String()` with no host validation:

```go
// if location port 443 is missing, add it back
if reqPort == "443" && locationHostname == reqHostname && locationPort == "" {
    location.Host = locationHostname + ":" + reqPort
}
url := location.String()   // <-- no host/scheme validation; credentials follow the Location
```

## Fix already on v2

`v2` commit `4683c46` (PR #1152) adds a `sameUploadHost()` helper and validates the `Location` host/scheme before issuing the `PUT`:

```go
// Validate the Location stays on the same host to prevent credentials from
// being forwarded to an attacker-controlled endpoint.
if !sameUploadHost(location, req.URL) {
    return fmt.Errorf("blob upload Location %q is on a different host than the registry %q", location.Host, req.URL.Host)
}
if req.URL.Scheme == "https" && location.Scheme != "https" {
    return fmt.Errorf("blob upload Location %q downgrades scheme from https", location.Host)
}
```

`sameUploadHost` normalizes implicit default ports (80/443) so the existing port-443 workaround behavior is preserved.

## Proposed action

Forward-port #1152 to `main` (v3). The change cherry-picks cleanly onto current `main`; `go build ./...` and `go test ./registry/remote/` pass, including the added `Test_BlobStore_Push_CrossHostRedirect` regression test that asserts no `Authorization` header leaks to a cross-host `Location`.

## References

- Advisory: GHSA-jxpm-75mh-9fp7 (CWE-918, CVSS 7.5)
- v2 fix: #1152 (commit `4683c46`)
- Vulnerable function: `registry/remote/repository.go` → `blobStore.completePushAfterInitialPost`

iamhaseebn:
## Summary

- Forward-port the established v2 validation for blob upload Location headers to v3.
- Require the upload target to remain on the registry host after normalizing default ports.
- Reject HTTPS-to-HTTP downgrades before the follow-up PUT request is created.
- Add regression coverage proving a cross-host upload Location is rejected before the target endpoint is reached.

Fixes #1273

## Testing

- go test ./registry/remote/ -run Test_BlobStore_Push -count=1
- make test

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"mime"
 	"net/http"
+	"net/url"
 	"slices"
 	"strconv"
 	"strings"
@@ -998,6 +999,25 @@ func (s *blobStore) Push(ctx context.Context, expected ocispec.Descriptor, conte
 	return s.completePushAfterInitialPost(ctx, req, resp, expected, content)
 }
 
+// sameUploadHost reports whether location and reqURL refer to the same host,
+// normalizing implicit default ports (80 for http, 443 for https) so that
+// e.g. "example.com" and "example.com:443" compare equal over HTTPS.
+func sameUploadHost(location, reqURL *url.URL) bool {
+	if location.Hostname() != reqURL.Hostname() {
+		return false
+	}
+	canonicalPort := func(u *url.URL) string {
+		if p := u.Port(); p != "" {
+			return p
+		}
+		if u.Scheme == "https" {
+			return "443"
+		}
+		return "80"
+	}
+	return canonicalPort(location) == canonicalPort(reqURL)
+}
+
 // completePushAfterInitialPost implements step 2 of the push protocol. This can be invoked either by
 // Push or by Mount when the receiving repository does not implement the
 // mount endpoint.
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

PR #1279, opened 2026-08-01
feat(signature): add simple signing verification package

TerryHowe:
## What

Adds `registry/remote/signature` — the OpenPGP simple-signing verification
layer for oras-go v3.

- **`verifier.go`** — `DefaultSignedByVerifier` implements
  `policy.SignedByVerifier`: fetches signatures from a `SignatureStore`,
  verifies the OpenPGP signature, validates the payload digest, and applies
  identity matching.
- **`lookaside.go`** — lookaside signature storage (sigstore attachments),
  resolving signature locations from the registries.d config.
- **`simplesigning.go`** — simple-signing payload parsing.
- **`identity.go`** — identity / reference matching rules.
- **`openpgp.go`** — OpenPGP keyring loading and signature verification.
- **`storage.go`** — the `SignatureStore` interface.

Adds `github.com/ProtonMail/go-crypto` (pulls in `cloudflare/circl` as an
indirect dep).

Closes: https://github.com/oras-project/oras-go/issues/1029

## Stacked on #1225

> [!IMPORTANT]
> This PR depends on **#1225** (`registry/remote/config`) and is stacked on
> its branch. Until #1225 merges, the diff here also shows the config commit;
> GitHub will narrow it to just the signature commit once #1225 lands. Please
> review/merge #1225 first. Kept as a **draft** until then.

`signature` uses `config.RegistriesDConfig` for lookaside resolution
(`config.NewSignedByVerifierFromConfig`) and the already-merged
`registry/remote/policy` package. This is the package #1225's description
calls out as unblocked by it.

## Notes

- Part of the **feat/everything (#1130)** v3 breakdown.

## Testing

`go build ./...`, `go vet ./registry/remote/signature/...`, and
`go test ./registry/remote/signature/...` all pass (coverage 83.5%).

PR #1225, opened 2026-06-22
feat(config): add remote config package with credentials refactor

TerryHowe:
## What

Adds the `registry/remote/config` package — the foundational configuration
layer for oras-go v3:

- Docker `config.json` loading (auths, credHelpers, credsStore)
- `registries.conf` / `registries.d` (YAML) loading
- credential-helper and certificate discovery
- conversion to `properties.Registry` via `LoadConfigs` / `LoadConfigsWithOptions`

## Why this bundles credentials + configpaths

These three pieces are mutually coupled and cannot be split without breaking
compilation or tests, so they land together:

- **`registry/remote/credentials` refactor** — adds `AuthConfig` and reworks the
  stores around `internal/authtype`. `config` imports `credentials`, while the
  `credentials` tests import `config/configtest`, so the two are mutually
  dependent.
- **`registry/remote/internal/configpaths` refinements** — extends `PathResolver`
  with `DockerConfigPath`, `AuthPrimaryPath`, `PolicyPaths`, `CertsDirPaths` and
  `RegistriesDDirs`, which the config loaders require (the merged registries.conf
  strategy in #1131 only landed a subset).

Adds `gopkg.in/yaml.v3` for registries.d parsing.

## Notes

- Part of the **feat/everything (#1130)** v3 breakdown. Unblocks the signature
  package (depends on `config.RegistriesDConfig`).
- One adaptation vs. the source branch: `credentials/example_test.go` uses the
  current `remote.GetCredentialFunc` name; the rename to `NewCredentialFunc`
  lands with the later repository/auth PR.

## Testing

`go build ./...`, `go vet`, and `go test` pass for `config`, `credentials`, and
`configpaths`.

PR #1284, opened 2026-08-02
test(functional): add functional-tagged suite (config/mirror/policy/signature)

TerryHowe:
## What

Adds the second half of the `test/functional` module — the tests behind the
**`functional`** build tag: `config_test.go`, `mirror_test.go`,
`policy_test.go`, `signature_test.go`, and their own `functional_test.go`
harness. Plus the registry-deploy scripts they use (`setup.sh`, `teardown.sh`,
`deploy/`) and a second CI job.

These cover config-driven repository construction, mirror fallback, policy
enforcement, and simple-signing verification, end-to-end.

## Stacked on #1283 — and BLOCKED

> [!IMPORTANT]
> **Draft. Does not build against `main` yet.** The `functional`-tagged suite
> depends on packages still landing via the v3 breakdown:
> - `registry/remote/config` — #1225
> - `registry/remote/signature` — #1279
> - the `ClientBuilder` / `NewRepositoryWithProperties` API — not yet on main
>
> Because those packages are absent, the module cannot compile or
> `go mod tidy` against `main` today, so CI here will be red until they merge.
> `go.mod`/`go.sum` carry the full dependency set for the unblocked end state.
> Mark ready once the dependencies land (and rebase onto #1283).

This is stacked on **#1283** (the `k8sfunctional` suite). Until #1283 merges,
the diff here also shows its commits.

## Why a separate tag / harness

The two suites deploy differently and always did (in the source branch):

| Tag | Deploy path | Registries |
|---|---|---|
| `k8sfunctional` (#1283) | `TestMain` applies `testdata/k8s` | single zot |
| `functional` (this PR) | `setup.sh` + `deploy/` | primary + mirror registry (mirror needs TLS) |

The added CI `functional` job provisions kind, runs `setup.sh`, then
`go test -tags functional`. It's wired for the end state and documented as
blocked in the workflow.

## Testing

Blocked as above — cannot run against `main`. Verified only that the files are
the complete `functional`-tagged set and that the `k8sfunctional` suite
(inherited from #1283) is unaffected.

Part of the **feat/everything (#1130)** v3 breakdown.


TerryHowe:
Refreshed: rebased onto current `main` and re-tidied `test/functional/go.mod`.

Note on the base: this branch is now stacked on #1328. The `functional`-tagged suite calls `remote.NewClientBuilder` in `mirror_test.go`, which does not exist on `main` yet — rebased onto plain `main`, `go test -tags functional` fails to compile and this PR's CI goes red. Stacking on #1328 keeps it green.

The two builder commits at the base of this branch will drop out of the diff once #1328 merges; this PR should be reviewed for the `test/functional/` changes only.

TerryHowe:
Folded into #1283 and closing.

This branch was a strict superset of #1283 — the same two base commits plus the `functional`-tagged suite — so keeping them apart meant rebasing the same files twice for no review benefit. #1283 now carries both suites, is stacked on #1328 the same way this branch was, and includes the fix for `TestSignature_GPG_WrongKeyRejected`.

Nothing is dropped.

Issue #1294, opened 2026-08-06
`oci.Store.GC` hangs forever when an untagged referrer's subject is not reachable from a tag


aimd54:
I hit this while garbage-collecting an OCI layout store that holds signatures as referrers. After removing a tag from a signed manifest, `GC` simply stopped returning. No error, no progress, one core at 100% until I killed it. It took a while to find because everything about the store looks healthy.

The cause is a shadowed variable in `gcIndex`, in `content/oci/oci.go`:

```go
subject := &desc
for {
    subject, err := manifestutil.Subject(ctx, s.storage, *subject)
    if err != nil {
        return err
    }
    if subject == nil {
        break
    }
    if graph.Exists(*subject) {
        ...
        break
    }
}
```

The `subject, err := ...` declares a new `subject` scoped to the loop body, so the `*subject` passed into the call keeps reading the outer one. Every iteration recomputes the same first hop, and the loop can only end if that hop is `nil` or is already in the graph. If it is neither, this spins forever.

There are really two problems in that one line. The obvious one is the hang. The quieter one is that the chain is never followed past a single hop, so a referrer that would reach the graph two or three hops up can never be recognised, which is what the loop looks like it was written to do.

## Getting into that state

You need a manifest still in `index.json` under its own digest, whose subject exists in storage but is not reachable from any tag. Removing a tag from a manifest that has referrers does it: the referrer stays indexed, the subject is still on disk, and nothing tagged reaches either of them. Every `GC` after that hangs.

I think `TestStore_GC` misses it by a hair. The referrers with unreachable subjects are dropped from the index with `tagResolver.Untag` before `GC` runs, so the second loop never sees them, and the one untagged manifest that does survive has no subject at all, so the walk exits on its first check.

## Reproducing it

Two untagged manifests, one pointing at the other, nothing tagged:

```go
func TestGCHangs(t *testing.T) {
	s, err := oci.New(t.TempDir())
	if err != nil {
		t.Fatal(err)
	}
	ctx := context.Background()

	empty := content.NewDescriptorFromBytes(ocispec.MediaTypeEmptyJSON, ocispec.DescriptorEmptyJSON.Data)
	if err := s.Push(ctx, empty, bytes.NewReader(ocispec.DescriptorEmptyJSON.Data)); err != nil {
		t.Fatal(err)
	}
	push := func(m ocispec.Manifest) ocispec.Descriptor {
[...]

PR #1295, opened 2026-08-06
fix(oci): advance the subject walk in gcIndex

aimd54:
Refs #1294

`gcIndex` walks the subject chain of an untagged manifest to decide whether it reaches the reachable graph, but the loop re-declared `subject` inside its own body. The dereference passed into the call therefore read the outer variable on every pass, so each iteration recomputed the first hop.

Two things follow from that. `GC` never returns when the first hop is neither `nil` nor already in the graph, spinning with no error and no progress. And the chain is never followed past one hop, so a referrer that would reach the graph further up cannot be recognised, which is what the loop looks like it was meant to do.

An untagged referrer whose subject is no longer reachable from a tag is enough to trigger it, which is the state a store is left in once a tag is removed from a manifest that has referrers.

This assigns the next hop to its own variable and moves the walk along. No cycle guard: subject chains are content-addressed, so a manifest would have to contain its own digest to loop.

## Tests

`TestStore_GCWithUnreachableReferrer` covers both halves. The first subtest fails on a timeout without this change, since the point is that `GC` never returns; the second asserts that a referrer two hops from a tagged manifest survives, which the single-hop walk could not express.

The existing `TestStore_GC` does not reach this path: the referrers with unreachable subjects are dropped from the index with `tagResolver.Untag` before `GC` runs, and the untagged manifest it keeps has no subject, so the walk exits on its first check.

`make test` passes. `content/oci` is at 85.8% coverage, `gcIndex` at 83.3%.

## Notes

- No public API change and no behaviour change beyond `GC` terminating and classifying multi-hop referrers correctly.
- The same line is present on the `v2` branch and in released `v2.6.2`. Happy to open the equivalent PR there if you want it backported.


TerryHowe (approved):
/lgtm

PR #1297, opened 2026-08-07
refactor(remote)!: hoist shared config to Registry, add parent pointer to Repository

TerryHowe:
## What

Restructures `remote.Registry` / `remote.Repository` so that shared
configuration lives on the parent `Registry` and each `Repository` references
it via a new `Registry *Registry` pointer:

```go
repo.Registry.PlainHTTP = true
```

`Client`, `Reference`, `PlainHTTP`, `Policy`, the warning handler, and the
metadata limits move from `Repository` to `Registry`. `Repository` keeps
per-repository overrides (media types, page sizes, max pages, `SkipReferrersGC`)
and reads shared settings through unexported accessors (`client()`,
`plainHTTP()`, `policy()`, `handleWarning()`, `maxMetadataBytes()`).

`Repository.Reference` is now a **method** (`Reference()`), not a field.

## Why

This is the foundational layer of the v3 work in #1130, carved out as a
standalone, reviewable base. The mirror fallback, config-driven builder, and
signature verification features all build on this parent-`Registry` model, so
landing it on its own first keeps those follow-up PRs small and focused.

## Scope

Foundation only — deliberately **excludes** the pieces from #1130 that depend
on other in-flight PRs:

- `builder.go` (config-driven `ClientBuilder`) — depends on the config/properties work
- `mirror.go` and read-path fallback hooks — tracked in #1280
- signature verification — tracked in #1279

## Breaking change

`Repository.Client`, `Repository.PlainHTTP`, and related config fields are
removed; `Repository.Reference` becomes a method. Configure shared settings on
`Repository.Registry` instead.

## Testing

`go build ./...` and `go test ./...` pass (33 packages, no failures);
`gofmt`/`go vet` clean on all changed files.

### DIFF
--- a/content/cache/target_test.go
+++ b/content/cache/target_test.go
@@ -322,7 +322,7 @@ func TestReferenceTarget_FetchReference(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create repository:", err)
 	}
-	repo.PlainHTTP = true
+	repo.Registry.PlainHTTP = true
 
 	// create cache
 	cache := memory.New()
@@ -441,7 +441,7 @@ func TestNew_WithReferenceFetcher(t *testing.T) {
 	if err != nil {
 		t.Fatal("failed to create repository:", err)
 	}
-	repo.PlainHTTP = true
+	repo.Registry.PlainHTTP = true
 
 	// create cache
 	cache := memory.New()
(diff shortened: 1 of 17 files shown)


sabre1041 (approved):
LGTM. Simplifies the configuration holistically

shizhMSFT (approved):
LGTM. The "registry" now becomes the entry-class concept. Multi-tenant registries require multiple registry objects to operation on different repositories at the same time with different settings (including auth).

PR #1301, opened 2026-08-14
fix(content/oci): make the local OCI store survive a partial write

cwedgwood:
`index.json` and `oci-layout` are written with `os.WriteFile`, which opens the file
`O_TRUNC`. The truncation succeeds even when the file system is full, so a write that fails
part-way leaves the file empty — and `loadIndexFile` treats a zero-length file as a decode
error rather than the missing-file state it already recovers from. The store then fails to
open for as long as the file is there, and nothing rebuilds it:

```
invalid OCI Image Index: failed to decode index file: EOF
```

This is an edge case, but we have hit it several times in production. A node fills up,
`index.json` is left at 0 bytes, and because the cache directory outlives the container,
restarting does not help. `blobs/` is unaffected, since blobs already go through
ingest-then-rename.

```go
os.WriteFile(filepath.Join(dir, "oci-layout"), []byte(`{"imageLayoutVersion":"1.0.0"}`), 0666)
os.WriteFile(filepath.Join(dir, "index.json"), nil, 0666)
_, err := oci.New(dir) // invalid OCI Image Index: failed to decode index file: EOF
```

**Commit 1** writes both metadata files to a temporary file in the same directory, flushes
it, and renames it over the target — the same thing `content/oci/storage.go` already does
for blobs and `registry/remote/config` does for the config file. The flush is load-bearing:
with delayed allocation a write is accepted against space that is never allocated and
`Close` does not report it, so rename alone would not fix the `ENOSPC` case. A zero-length
metadata file is now treated as missing and rewritten; malformed-but-not-empty still fails,
since that may be real corruption whose tags are worth keeping for a human to look at.

**Commit 2** fixes `ingest` returning `return "", err`, which clears the named return `path`
before the deferred cleanup runs, so it calls `os.Remove("")` and leaks the ingest file —
the same bug #1185 fixed in `registry/remote/internal/ioutil`. `GC` now also reclaims that
residue, which nothing did before. Happy to split this commit into its own PR.

Worth knowing, all documented on `writeFileAtomic`: replacing a file rather than writing
through it means symlinks and hard links are replaced, and a `0444` or `0000` target is
replaced where `os.WriteFile` returned `EACCES`, though its mode is preserved. Ownership,
ACLs and xattrs are not carried over. Permissions are otherwise unchanged — `os.CreateTemp`
ignores the umask, so the temporary file is created with `O_EXCL` at 0666, or at the
target's own mode when replacing, and set exactly before the rename. The directory flush
after the rename is best-effort and returns nothing, since by then the rename has already
taken effect.

Five of the new tests fail against current `main`, including `TestStore_EmptyIndexFile` with
the exact error above and `TestStorage_BadPush_NoIngestFileLeftBehind` with
`len(ingest entries) = 1, want 0`. `make test` passes under both matrix Go versions (1.25
and 1.26), coverage 85.3%, and `TestStore_BadIndex` and `TestStore_BadLayout` still pass.

Not in scope: the blob write is not flushed either, and its directory entry is not synced
after the rename. Both are worth fixing, but separately from this.

We still depend on the `v2` line in a number of places, so we would value a backport if this
is accepted. The two commits cherry-pick onto `v2` cleanly and the suite passes there; happy
to open that PR whenever you want it.


content/oci/oci.go
```diff
@@ -521,6 +582,50 @@ func (s *Store) GC(ctx context.Context) error {
...
+	// every file in the ingest directory is the residue of a blob write.
+	if err := removeFiles(s.storage.ingestRoot, func(string) bool { return true }); err != nil {
+		return err
+	}
+	// the temporary files of metadata writes are created next to the file that
+	// they replace, in the root of the store.
+	return removeFiles(s.root, func(name string) bool {
+		return isTempFileOf(name, ocispec.ImageIndexFile) || isTempFileOf(name, ocispec.ImageLayoutFile)
+	})
+}
```
TerryHowe, line 624:
```suggestion
// leftoverExpiry is how long a temporary file must have been untouched before
// GC reclaims it, so that a write in progress -- possibly by another Store on
// the same directory -- is never reclaimed from underneath.
const leftoverExpiry = time.Hour

func (s *Store) gcLeftovers() error {
      // A write removes its own temporary file when it fails, so anything still
      // here is the residue of a process that died mid-write -- or a write that
      // another Store on the same directory is performing right now, which the
      // lock of this Store does not serialize against. Only entries that have
      // not been touched for leftoverExpiry are reclaimed, since the modification
      // time of a write in progress keeps advancing.
      stale := olderThan(leftoverExpiry)

      if err := removeFiles(s.storage.ingestRoot, stale); err != nil {
              return err
      }
      return removeFiles(s.root, func(entry fs.DirEntry) bool {
              name := entry.Name()
              if !isTempFileOf(name, ocispec.ImageIndexFile) && !isTempFileOf(name, ocispec.ImageLayoutFile) {
                      return false
              }
              return stale(entry)
      })
}

// olderThan returns a matcher accepting the directory entries that have not
// been modified for at least d. An entry whose information cannot be read is
// not matched: it has either just been removed by someone else, or it is not
// ours to reason about.
func olderThan(d time.Duration) func(entry fs.DirEntry) bool {
      return func(entry fs.DirEntry) bool {
              fi, err := entry.Info()
              if err != nil {
                      return false
              }
              return time.Since(fi.ModTime()) >= d
      }
}
```

TerryHowe (changes requested):

content/oci/oci.go
cwedgwood: Thanks — you are right, and this is a real hole rather than a style point. The
sweep assumed that a temporary file still present belongs to a write that will
never complete, which holds for a process that died but not for a second `Store`
opened on the same directory: that write is live, and the lock of one `Store`
does not serialize against it.

Applied your suggestion as written. Also updated the commit message, which
asserted the assumption you just corrected ("a temporary file present when `GC`
runs belongs to a write that will never complete") — it now describes the expiry
instead.

Two things I added on top of the suggestion:

- `TestStore_GC_Leftovers` now covers both directions: entries backdated past
  `leftoverExpiry` are reclaimed, and a freshly written ingest file and metadata
  temporary file survive. Against the unconditional sweep the new case fails
  with the in-flight ingest and `index.json_*` reported missing, so it holds the
  behaviour you asked for rather than just describing it.
- `Test_olderThan` covers the matcher directly, including that an entry whose
  `Info()` cannot be read is not matched.

`removeFiles` now takes `func(fs.DirEntry) bool` rather than a name, as your
suggestion needs.

Rebased onto current `main` while I was here. `make test` passes under both
matrix Go versions, and `lint`, `build` and the license check pass on a fork run
of these exact commits.

cwedgwood:
Suggestion applied and pushed — this is ready for another look.

- `leftoverExpiry`, `olderThan()` and the reworked `gcLeftovers` are in as written; `removeFiles` now takes `func(fs.DirEntry) bool`.
- `TestStore_GC_Leftovers` covers both directions: entries backdated past the expiry are reclaimed, a freshly written ingest file and metadata temporary file survive. Against the previous unconditional sweep that case fails, so it holds the behaviour rather than describing it. `Test_olderThan` covers the matcher, including an entry whose `Info()` cannot be read.
- The commit message asserted the assumption you corrected; it now describes the expiry instead.
- Rebased onto `main`. `make test` passes on both matrix Go versions, and `lint`, `build` and the license check pass on a fork run of these commits — the workflow approval here is still pending, so the checks on this PR have not run yet.

One note in case it saves you a broken build: the inline suggestion still shows as applicable, but it is anchored to what is now the closing brace of `gcLeftovers`, so committing it would splice a second copy of all three declarations into the function. It is already applied by hand.

cwedgwood:
@TerryHowe — this has your suggestion applied and is rebased on `main`; the checks here still need a workflow approval to run. Anything else you'd like changed?

TerryHowe:
I'm seeing some performance issues in testing, but I need to investigate the finding

cwedgwood:
Confirmed, and thank you for catching it. I reproduced it, and it is worse than I would have
guessed. I benchmarked the blob path when I was looking at #1203 and never benchmarked my own
change, which is the thing I should have done before opening this.

The cause is that `AutoSaveIndex` is on by default, so `index.json` is rewritten on every push and
every tag, and I made each of those rewrites pay an `fsync` on the file plus an `fsync` on the
directory.

Measured on ext4 on an enterprise SSD (`fsync` ~0.09 ms), 30 iterations x 3 runs, medians. "50
pushes" is 50 config+manifest pairs into one store; `Tag` is one tag on an existing manifest:

| variant                         | 50 pushes | vs base | `Tag`    | vs base |
|---------------------------------|-----------|---------|----------|---------|
| `main` today                    | 7.80 ms   | 1.00x   | 0.041 ms | 1.00x   |
| temp file + rename, no fsync    | 8.15 ms   | 1.05x   | 0.065 ms | 1.58x   |
| + fsync the file                | 13.93 ms  | 1.79x   | 0.152 ms | 3.74x   |
| + fsync the directory (this PR) | 18.44 ms  | 2.36x   | 0.228 ms | 5.61x   |

The ratio gets *worse* on faster storage, because the baseline shrinks faster than the `fsync` does
— on a slower machine the same comparison was 1.93x / 1.73x. The numbers above are not a worst case.

**The two fsync calls are not equally worth their cost**

I would like to drop the directory `fsync` and keep the one on the file, if that seems reasonable
to you.

The directory `fsync` only makes the *rename* durable. Losing it means a crash leaves the previous
`index.json` in place, which is still a valid index, so the failure is benign. It costs 2.36x ->
1.79x on pushes and 5.61x -> 3.74x on tags, which I do not think it earns.

The one on the file is doing more work. Without it there is no guarantee that the contents reach
the disk before the rename does, so a crash can leave the file present and empty — the same
unusable store this PR is about, arriving by a different route. That is the part I would rather
not give up, because it would leave the symptom reachable while the code reads as though it were
handled.

**The frequency may matter more than the per-call cost**

The overhead scales with how often `index.json` is rewritten, and the store already has a knob for
that. The same 50-manifest workload with `AutoSaveIndex = false` and one `SaveIndex()` at the end:

|                                    | 50 pushes | vs base |
|------------------------------------|-----------|---------|
| `main` today                       | 4.90 ms   | 1.00x   |
| this PR, directory fsync dropped   | 5.19 ms   | 1.06x   |
| this PR as it stands               | 5.41 ms   | 1.10x   |

So a caller doing bulk work can already pay one `fsync` instead of fifty for about 6%. If that
seems useful I will add a sentence to the `AutoSaveIndex` docs pointing at it.

That leaves 1.79x on the default path. My instinct is that the safe default is the right one here,
partly because it is not symmetric: a caller who wants throughput has `AutoSaveIndex` today,
whereas a caller who wants the flush cannot add it from outside. But you have a much better view
of how people actually use this than I do, and if 1.79x is too much for the common case I am happy
to put it behind an option on `Store`, or to take a different approach entirely if you have one in
mind — I may well be missing a cheaper way to get the same guarantee.

Happy to push whichever shape you prefer. Worth knowing that the checks here have never run — they
are still waiting on a workflow approval — so the numbers above are from local runs and a run on
my fork.

cwedgwood:
Hi Terry — just checking back when you have a chance. We’ve rebased onto current `main`. We’re happy to drop the directory fsync and keep the file fsync, put the durability behavior behind an option, or follow another approach you prefer. Thanks again for looking into the performance tradeoff.

TerryHowe (dismissed):
/lgtm

TerryHowe:
Apologies for the slow reply here, and thanks for the benchmark table — producing it was the right call.

I've dismissed my approval on this while we settle the tradeoff. That approval was submitted in error on my side, before I'd caught up with this thread.

I reproduced your results and then ran the same comparison on macOS, and I think it changes the shape of the decision.

Go's `os.File.Sync()` on darwin issues `fcntl(F_FULLFSYNC)` rather than `fsync(2)` — a full drive-cache barrier, ~5 ms on the machine below, against the ~0.09 ms you measured on ext4. This PR performs two of them per index write.

Apple M1, APFS, medians of 3 x 100 iterations. `Tag` is one tag on an existing manifest in a store holding 100 tags; `SaveIndex` is the write on its own:

| variant                          | `Tag`     | vs base | `SaveIndex` | vs base |
|----------------------------------|-----------|---------|-------------|---------|
| `main` today                     | 0.114 ms  | 1.00x   | 0.046 ms    | 1.00x   |
| temp file + rename, no fsync     | 0.257 ms  | 2.3x    | 0.19 ms     | 4.2x    |
| + fsync the file                 | 5.67 ms   | 50x     | —           | —       |
| + fsync the directory (this PR)  | 10.38 ms  | 91x     | 10.09 ms    | 221x    |

One machine, `-benchtime=100x`, not a benchstat run — but the gap is two orders of magnitude, so the precision isn't what's carrying the argument.

The part that bears on your proposal: dropping the directory fsync halves the cost exactly, because it removes one of the two barriers, and still leaves ~50x. It's aimed at the cheaper of the two problems, so I don't think it's sufficient on its own.

**What I'd like instead**

Keep the temporary file and the rename unconditionally, and put the fsync behind a field on `Store`, defaulting off.

The rename is the part that fixes the bug you opened this for, and it costs 2.3x. For a plain ENOSPC the write error is returned before the rename happens, so the existing `index.json` survives untouched — which is the production failure in your description. The fsync buys something real but different: cover for delayed allocation, where the error surfaces at writeback instead of at `write()`, and for crash or power loss. That's worth having available and worth paying for deliberately. I don't think it should be the price of every `Tag()` and every manifest push by default.

On the asymmetry you raised — that a caller who wants the flush can't add it from outside — that's true, and it's a good argument for exposing the option. It's a weaker argument for the default, since a caller who doesn't want a 10 ms tag can't opt out from outside either.

Your point about `AutoSaveIndex` is well taken, and the docs sentence pointing at it would be welcome regardless.

Happy to look at a different shape if you see one, but this is the one I'd merge.

TerryHowe:
I just wonder if there is any way this could be made optional

content/oci/oci.go
```diff
@@ -81,11 +83,19 @@ type Store struct {
 }
 
 // New creates a new OCI store with context.Background().
+//
+// If `index.json` is present but empty, which is what a write interrupted by a
+// full file system leaves behind, it is reinitialized rather than reported as
+// an error. The blobs of the store are kept, but the tags recorded in the lost
+// index are not recoverable, and the blobs they referenced become unreferenced
+// and are removed by the next call to GC.
```
TerryHowe, line 91: **A partially written `index.json` still fails to open — the same failure this PR is about.**

The recovery keys on `Size() == 0`, but `os.WriteFile` loops on `Write`, so an interrupted write doesn't only produce zero-length files: a short write leaves however many bytes fit before the failure. Against this branch, with a truncated index:

```
New() with partially-written index.json -> err = invalid OCI Image Index: failed to decode index file: unexpected EOF
```

`writeFileAtomic` stops new occurrences of both variants, so this doesn't undermine the fix. But the recovery half is narrower than the failure it exists to recover from, and an already-damaged store in the field has roughly even odds of being the variant that's still bricked.

I'm not asking you to reset a malformed-but-non-empty file — your reasoning for keeping it for inspection is sound. I'd just like the limit stated where a caller will find it.

```suggestion
// If `index.json` is present but empty, which is what a write interrupted by a
// full file system leaves behind, it is reinitialized rather than reported as
// an error. The blobs of the store are kept, but the tags recorded in the lost
// index are not recoverable, and the blobs they referenced become unreferenced
// and are removed by the next call to GC.
//
// Only a zero-length file is recovered this way. A write interrupted part-way
// can also leave a short, undecodable `index.json` behind, and that is still
// reported as an error, since it may be corruption whose surviving tags are
// worth keeping for inspection rather than discarding.
```

content/oci/oci.go
```diff
@@ -521,6 +583,80 @@ func (s *Store) GC(ctx context.Context) error {
 			}
 		}
 	}
+
+	// clean up the temporary files left behind by interrupted writes
+	if err := s.gcLeftovers(); err != nil {
+		return fmt.Errorf("unable to remove leftover temporary files: %w", err)
+	}
```
TerryHowe, line 590: **A single unremovable file turns an otherwise-successful `GC` into a failed one.**

`gcLeftovers` runs after the blob collection has already finished, and `removeFiles` returns on the first `os.Remove` failure that isn't `ErrNotExist`. So `GC` reports an error for work that did succeed. The concrete case is Windows, where removing a file another process holds open fails with a sharing violation — and tolerating a concurrent store is exactly what `leftoverExpiry` is there for.

Since the sweep is opportunistic by design, I'd rather it couldn't fail the call at all:

```suggestion
	// Reclaiming leftovers is opportunistic: the collection above has already
	// completed, so a temporary file that cannot be removed -- one held open by
	// another process on Windows, say -- must not turn a successful GC into a
	// failed one.
	_ = s.gcLeftovers()
```

Dropping the error return from `gcLeftovers` and `removeFiles` altogether, the way `syncDir` already does for the same reason, would read better than discarding it at the call site. Either is fine by me.

TerryHowe (commented):
Two items from reading the current code, separate from the performance thread above. Both are small, and neither is new in the last push — I simply never posted them.

cwedgwood:
Thanks!
Done as suggested, rebased, and tests pass locally.

PR #1302, opened 2026-08-14
refactor(credentials): decouple tests from the config package

### LINKED ISSUE #1290: Split credentials refactor out of #1225 into an independent PR
## Summary

Split the `credentials` refactor (and the `internal/configpaths` refinements) out
of #1225 so they can land ahead of the new `registry/remote/config` package.

## Background

#1225 ("feat(config): add remote config package with credentials refactor")
bundles three things:

1. the new `registry/remote/config` package,
2. a refactor of `registry/remote/credentials`, and
3. refinements to `registry/remote/internal/configpaths`.

The PR description states these "cannot be split without breaking compilation or
tests." In practice the coupling is weaker than that: the **production** code is
cleanly decoupled, and only two `_test.go` files tie `credentials` back to
`config`.

## Why a split is feasible

**Production dependency direction is one-way: `config` → `credentials`.**

- `credentials/store.go` exposes a registration hook — the `ConfigFileLoader`
  type, a package-level `defaultConfigLoader`, and the exported
  `SetDefaultConfigLoader()`. It imports nothing from `config`.
- `config/config.go` registers itself in `init()` via
  `credentials.SetDefaultConfigLoader(...)`.
- `internal/configpaths` imports neither `config` nor `credentials`, so it can
  ship first with zero coupling.

**The only `credentials` → `config` references are test-only:**

- `credentials/file_store_test.go` and `credentials/store_test.go` use
  `config/configtest.Config` as a fixture struct. `configtest` is a standalone
  ~39-line package of JSON-shaped structs that imports nothing (not even
  `config`).
- `credentials/file_store_test.go` additionally blank-imports `config`
  (`_ ".../config" // registers config loader`) because those tests exercise
  `NewStore` / `NewStoreFromDocker`, which require a registered loader.

Nothing here breaks *compilation* without `config`.

## Proposed work

1. Land `internal/configpaths` refinements on their own (independent, no coupling).
2. Land the `credentials` refactor without the `config` package by:
   - inlining or relocating the `configtest` fixture struct into the credentials
     test package (or a shared internal test helper), and
   - registering a stub loader via `SetDefaultConfigLoader` (or switching to
     `NewStoreFromConfig`) in `file_store_test.go` instead of blank-importing
     `config`.
3. Land the `registry/remote/config` package as a follow-up, swapping the stub
   test loader for the real `config` registration.

## Benefit

Smaller, independently reviewable PRs; `configpaths` and the `credentials`
refactor stop being blocked on review of the much larger `config` package.

## Trade-off

The credentials tests get touched twice (stub first, real `config` later). If the
priority is simply landing the `config` layer to unblock the signature package,
the existing bundle in #1225 is reasonable — this issue tracks the split as the
alternative for reviewability.

Refs #1225. Part of the feat/everything (#1130) v3 breakdown.

srikary12:
## What this is

Fixes #1290 — though not the way the issue proposed, because the situation changed.

#1290 asked to split the `credentials` refactor out of #1225 so it could land ahead of the new `registry/remote/config` package. **#1225 merged on 2026-08-05**, so the pre-merge split is no longer possible. What still applies is the issue's underlying goal: keep the dependency direction one-way, so `credentials` does not reach into `config`.

The production code was already clean — `go list -deps ./registry/remote/credentials` has never contained `config`. The registration-hook design (`ConfigFileLoader` / `SetDefaultConfigLoader`) works as intended. The remaining coupling was entirely test-only, in three forms, all addressed here.

## Changes

**1. `configtest` moves to a neutral internal package.** `registry/remote/config/configtest` → `registry/remote/internal/configtest`. Both `config` and `credentials` live under `registry/remote`, so both can import it without `credentials` depending on `config`. Only import lines change; the `configtest.` selectors are untouched.

This is not a user-facing break: the package was introduced by #1225 itself and v3 has only `-dev` tags, so it never shipped in a stable release. No `MIGRATION_GUIDE.md` entry seemed warranted — happy to add one if you'd prefer.

**2. `credentials` gets its own `testdata/`.** Three test files read their fixtures from `../config/testdata`, reaching across a package boundary. They now own the eight files they actually use. `config/testdata` keeps its copies plus the fixtures only it uses.

**3. The loader registration is made explicit.** The blank import of `config` in `file_store_test.go` registers the default loader for the *whole test binary*, including the in-package tests in `store_test.go`. That dependency spans two packages and was undocumented — removing the import made **44 tests** fail with an unexplained `ErrNoConfigLoader`.

It can't be moved into `package credentials`: `config` imports `credentials`, so that would be an import cycle, and only an external test package can close it. Replacing it with a stub loader (as #1290 floated, as a *temporary* device) would now cost real coverage — 19 `store_test.go` tests round-trip actual Docker config JSON through the real parser.

So the import stays, but is now documented, and a `TestMain` fails once with an explanation if the loader ever goes missing. The 44 opaque failures become one legible message.

**4. Removes `registry/remote/internal/configuration`.** Nothing imported it but its own test. It is the pre-refactor config implementation, superseded by `registry/remote/config` (a strict API superset), left behind when #1225 merged. Its testdata files were byte-identical to the copies the credentials tests now own, so git records them as renames and the repo's file count comes out neutral.

Flagging this one explicitly since it's an unrelated ~1,700-line deletion riding along — happy to split it into its own PR if you'd rather review it separately.

## Result

The only remaining `credentials` → `config` reference is the single deliberate loader registration, confined to the external test package:

| Edge | Before | After |
|---|---|---|
| Production deps | clean | clean |
| In-package test imports | `config/configtest` | **none** |
| External test imports | `config` + `config/configtest` | `config` only |

## Verification

- `go build ./...`, `go vet ./...`, and the full `go test ./...` suite all pass.
- Dependency edges confirmed with `go list -deps` and `go list -f '{{.TestImports}} {{.XTestImports}}'`.
- Guard behaviour checked by removing the blank import and confirming a single explanatory failure rather than 44.

Pre-existing `gofmt` drift elsewhere in the tree (including unrelated alignment in `store_test.go`) is left alone to keep the diff scoped.


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1316, opened 2026-08-19
fix(signature): bound untrusted reads and tighten identity matching

TerryHowe:
The signature package reads data from a lookaside store named by configuration, so those reads are attacker-influenced. Several were unbounded, and one identity match was wider than intended.

## Unbounded reads

- `GetSignatures` and `PutSignature` scanned the signature index until the store answered "not found". A store that always returns 200 keeps the loop running forever with the accumulated slice growing. Capped at 32 with a new `ErrTooManySignatures`, failing loudly rather than verifying against an arbitrary prefix.
- Signature bodies were read with `io.ReadAll`. Both the HTTP and file paths are now bounded at 4 MiB.
- `VerifyOpenPGPSignature` read the OpenPGP message body unbounded — before the signature was checked, and with the literal packet possibly compressed. That is a decompression bomb. Bounded at 1 MiB, keeping the drain that populates `md.SignatureError`.
- The lookaside HTTP client had no timeout.

## Identity matching

`remapIdentity` matched its prefix as a plain string, so a configured prefix of `registry.io/team` also matched `registry.io/team-evil/app:v1` and remapped it into the signer's namespace. Whoever controls that adjacent repository at the signer could then produce a signature satisfying the policy. The prefix must now end at a component boundary (`/`, `:`, `@`, or end of string).

## Error reporting

`Verify` discarded every per-signature rejection and returned `(false, nil)`, making a misconfigured keyring indistinguishable from unsigned content. Rejections are now collected and returned when nothing validates.

This still fails closed. `Evaluator.IsImageAllowed` turns a requirement error into `(false, err)` and `checkPolicy` denies, so the operation is blocked either way — now with a reason attached. Genuinely unsigned content (`len(sigs) == 0`) still returns a clean `(false, nil)`, so "unsigned" stays distinguishable from "rejected".

Six tests encoded the old swallow-everything contract and now assert a reason is reported.

## Also

- `LoadKeyRing` no longer appends into its caller's slice.
- `isNotFound` uses `errors.As` with an `Unwrap`.
- `signaturePath`'s comment described stripping the registry host when the code deliberately keeps it, matching the containers/image layout.

### DIFF
--- a/registry/remote/signature/hardening_test.go
+++ b/registry/remote/signature/hardening_test.go
@@ -1,0 +1,218 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package signature
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/opencontainers/go-digest"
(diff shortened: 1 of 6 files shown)


sabre1041 (approved):
LGTM

PR #1317, opened 2026-08-19
fix(remote): correct mirror fallback semantics

TerryHowe:
Two bugs in the mirror read-fallback path.

## `Exists` reports missing content that exists

`withMirrorFallbackExists` treated a mirror's `(false, nil)` as authoritative:

```go
ok, err := exists(ctx, mirrors[i].Repository, target)
if err == nil {
    return ok, nil   // returns false without consulting the primary
}
```

A mirror answering "I don't have it" is not an error, so this returned `false` and the primary was never asked. Content present on the primary but absent from a mirror was reported as non-existent.

Only a positive result now ends the search; a miss continues to the next mirror and then the primary.

## Cancelled contexts walk every mirror

`isMirrorFallbackError` compared against the sentinels with `==`:

```go
return err != context.Canceled && err != context.DeadlineExceeded
```

The transport wraps context errors, so these comparisons never matched in practice and a cancelled context still tried every remaining mirror before failing. Now uses `errors.Is`.

The existing test only covered bare sentinels, which is why this survived; the new test covers wrapped and doubly-wrapped errors.

### DIFF
--- a/registry/remote/mirror.go
+++ b/registry/remote/mirror.go
@@ -17,6 +17,7 @@ package remote
 
 import (
 	"context"
+	"errors"
 	"io"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

PR #1318, opened 2026-08-19
fix(credentials)!: stop persisting the detected credentials store

TerryHowe:
On the first successful `Put`, `DynamicStore` wrote the detected native helper into the user's `~/.docker/config.json` and saved the file:

```go
return ds.setCredsStoreOnce.Do(func() error {
    if ds.detectedCredsStore != "" {
        ds.config.SetCredentialsStore(ds.detectedCredsStore)
        if err := ds.config.Save(); err != nil { ... }
    }
    return nil
})
```

That edits a file this library does not own — shared with the Docker CLI and every other tool that reads it — as a side effect of saving one credential. And because `credsStore` is a global setting, it changed *their* behaviour too, not just ours.

The write also bought nothing locally: `detectedCredsStore` is consulted directly by `getStore`, so detection works without persisting anything. The write only exported the decision to other tools.

**Detection itself is unchanged and stays on by default.** The `!cfg.IsAuthConfigured()` guard already limits it to config files with no auth at all, so users with existing credentials there are unaffected.

## Breaking change

`SetCredentialsStore` and `Save` are removed from the `credentials.ConfigFile` interface. They were used only by the removed block — `PutAuthConfig` and `DeleteAuthConfig` call `saveFile()` internally, so the plaintext file store never needed them. The interface goes from 10 methods to 8.

`Test_DynamicStore_noAuthConfigured` asserted the old behaviour and now asserts the opposite, reading the config file back from disk to confirm no `credsStore` key appears — the file being the thing that was previously modified.

### DIFF
--- a/registry/remote/credentials/store.go
+++ b/registry/remote/credentials/store.go
@@ -25,8 +25,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-
-	"github.com/oras-project/oras-go/v3/internal/syncutil"
 )
 
 // ConfigFile is the interface for a Docker configuration file that provides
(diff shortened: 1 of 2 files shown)


sabre1041 (approved):
LGTM

PR #1319, opened 2026-08-19
refactor(config)!: remove the dead force-basic-auth option

TerryHowe:
`force-basic-auth` is parsed from registries.conf into `properties.Attributes.ForceBasicAuth`, and nothing reads it:

```
$ git grep -n ForceBasicAuth -- '*.go' | grep -v _test
registry/remote/config/properties.go:78:  if origReg.ForceBasicAuth {
registry/remote/config/properties.go:79:      props.Attributes.ForceBasicAuth = true
registry/remote/config/registries.go:61:  ForceBasicAuth bool `toml:"force-basic-auth"`
registry/remote/properties/attributes.go:55: ForceBasicAuth bool
```

Parse, store, and no consumer. Setting `force-basic-auth = true` today succeeds and silently does nothing, which is worse than not offering the option: it promises an authentication behaviour the library does not implement, and a user relying on it has no way to notice.

Beyond being unimplemented, it is the wrong thing to configure. The authentication scheme follows the registry's `WWW-Authenticate` challenge; a client overriding a Bearer challenge with Basic is working against the registry rather than with it.

What *is* worth configuring is how a bearer token is acquired for username/password credentials — the distribution-spec `GET` versus the OAuth2 password grant, which is a Docker extension rather than part of the OCI spec, and which not every registry implements. That option arrives together with the `ClientBuilder` that will consume it, rather than landing here as another unread field.

## Breaking change

The `force-basic-auth` registries.conf field and `properties.Attributes.ForceBasicAuth` are removed. Since nothing consumed either, no behaviour changes — configurations setting it were already no-ops.

### DIFF
--- a/registry/remote/config/properties.go
+++ b/registry/remote/config/properties.go
@@ -75,10 +75,6 @@ func NewRegistryProperties(ref string, regConf *RegistriesConfig) (*properties.R
 			props.Transport.Insecure = true
 		}
 
-		if origReg.ForceBasicAuth {
-			props.Attributes.ForceBasicAuth = true
-		}
-
 		switch origReg.ReferrersAPI {
 		case "supported":
 			props.Attributes.ReferrersAPI = properties.ReferrersAPISupported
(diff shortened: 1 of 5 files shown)


sabre1041 (approved):
Trim the fat. LGTM

PR #1107, opened 2026-03-03
docs: add usage scenarios for oras-go library

TerryHowe:
## Summary
- Adds `docs/SCENARIOS.md` documenting 10 usage scenarios for the oras-go library
- Covers: full config stack, CLI tool with flag overrides, policy enforcement, artifact distribution, object-oriented artifacts, registry mirroring, OCI local storage, credential management, signature verification, and library middleware
- Includes configuration loading options (LoadConfigs, LoadConfigsWithOptions, direct construction)
- Includes summary matrix mapping scenarios to packages and features

## Test plan
- [ ] Verify code examples are consistent with current API surface
- [ ] Review scenario descriptions for accuracy and completeness

PR #1322, opened 2026-08-19
docs: add design notes for signedBy resolution and config/credentials coupling

TerryHowe:
Two design notes for open v3 questions, written up so the decisions can be made deliberately rather than settling by default. Docs only — no code changes.

## `docs/design/signedby-tag-resolution.md`

A `signedBy` policy requirement can never be satisfied when the image is referenced by tag. The verifier needs a manifest digest (that is what a simple-signing payload binds to), but `checkPolicy` runs *before* the reference is resolved, so `parseImageDigest` gets a tag and errors out. Every tag pull under a `signedBy` policy is denied, however the image is signed.

It fails closed, so this is a functional gap rather than a security hole — but it makes the feature unusable for the most common way people name images.

The note works through why resolving first is not a one-line reordering (policy is deliberately pre-flight, and resolving inside the verifier re-enters the policy path), summarises how containers/image splits the decision, and compares three options: two-phase evaluation, resolving inside the verifier, or documenting the limitation. Recommends two-phase as the target with the documented limitation as an interim.

## `docs/design/config-credentials-coupling.md`

`credentials.NewStore` depends on a package-level mutable variable that the `config` package populates from its `init()`. Three consequences: whether `NewStore` works depends on whether an otherwise-unused package was imported; the exported `SetDefaultConfigLoader` writes an unsynchronised global that any caller can rewrite process-wide; and a ten-method `ConfigFile` interface mirrors `config.Config` purely to break the import cycle, so every change has to be made in two places.

Proposes moving the shared types to an internal package both can import, which removes the cycle and lets all three of `SetDefaultConfigLoader`, `ConfigFileLoader` and `ErrNoConfigLoader` go away, with the public API preserved via type aliases. Notes this is cheapest before the v3 API freeze.

### DIFF
--- a/docs/design/config-credentials-coupling.md
+++ b/docs/design/config-credentials-coupling.md
@@ -1,0 +1,130 @@
+# Design: decoupling `credentials` from `config` without a global
+
+Status: **proposal** — recommend settling before the v3 API freeze.
+
+## Problem
+
+`credentials.NewStore` depends on a package-level mutable variable that a
+*different* package populates as an import side effect.
+
+`registry/remote/credentials/store.go`:
+
+```go
+// defaultConfigLoader is set by the config package during init.
+var defaultConfigLoader ConfigFileLoader
+
+var ErrNoConfigLoader = fmt.Errorf("no config loader registered; import the config package or use NewStoreFromConfig")
+
+func SetDefaultConfigLoader(loader ConfigFileLoader) {
+	defaultConfigLoader = loader
+}
+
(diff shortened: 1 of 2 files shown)


TerryHowe:
Superseded by #1107, which now carries both design notes along with `SCENARIOS.md` and `docs/PLAN.md`. Consolidating the documentation-only changes so they can be reviewed and pruned in one place.

Closing in favour of that PR — nothing here is dropped, both files land unchanged.

PR #1326, opened 2026-08-22
avoid redundant policy evaluation in blobStore/manifestStore Exists

### LINKED ISSUE #1325: Skip redundant policy evaluation in blobStore.Exists and manifestStore.Exists
## What

`blobStore.Exists` and `manifestStore.Exists` in `registry/remote/repository.go` evaluate the access policy, then call `Resolve`, which evaluates it again for the same operation:

```go
func (s *blobStore) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
    if err := s.repo.checkPolicy(ctx, ""); err != nil {   // first evaluation
        return false, err
    }
    _, err := s.Resolve(ctx, target.Digest.String())      // Resolve checks again
```

`Resolve` opens with its own `checkPolicy`, and the context carries no marker, so every existence check runs the policy twice.

That is exactly the redundancy `policyCheckedKey` exists to prevent — the other operations (`Fetch`, `Push`, `Delete`, `Tag`, `PushReference`, `FetchReference`) all pass `withPolicyChecked` down. These two were missed.

## Proposed change

Pass `withPolicyChecked(ctx)` into the inner `Resolve` call in both `blobStore.Exists` and `manifestStore.Exists`, so the already-performed evaluation is not repeated.

## Impact

No behavioural change: a denial is still returned by the first check, before any request is made. With a `signedBy` requirement the second evaluation can mean a duplicated signature fetch and verification on every `Exists` call.

## Notes for a first-time contributor

- Two small edits in `registry/remote/repository.go` (`blobStore.Exists` and `manifestStore.Exists`).
- Verify with `go build ./...`, `go vet ./...`, `go test ./...`, and `golangci-lint run`.

A reference implementation exists in #1321.

srikary12:
## What this PR does

`blobStore.Exists` and `manifestStore.Exists` both call `checkPolicy` and then
delegate to their own `Resolve`, which calls `checkPolicy` again. For a
configured `Policy`, that means two `IsImageAllowed` evaluations per `Exists`
call — and for signature-based requirements (`signedBy` / `sigstoreSigned`)
that can mean doing the verification work twice.

This threads `withPolicyChecked(ctx)` into the inner `Resolve` call in both
methods, so the evaluation already performed by `Exists` is not repeated. This
is the same pattern already used by `Repository.Fetch`, `Repository.Push`,
`Repository.Resolve`, and friends when they delegate to a sub-store.

Fixes #1325

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -1227,7 +1227,7 @@ func (s *blobStore) Exists(ctx context.Context, target ocispec.Descriptor) (bool
 	if err := s.repo.checkPolicy(ctx, ""); err != nil {
 		return false, err
 	}
-	_, err := s.Resolve(ctx, target.Digest.String())
+	_, err := s.Resolve(withPolicyChecked(ctx), target.Digest.String())
 	if err == nil {
 		return true, nil
 	}
@@ -1431,7 +1431,7 @@ func (s *manifestStore) Exists(ctx context.Context, target ocispec.Descriptor) (
 	if err := s.repo.checkPolicy(ctx, ""); err != nil {
 		return false, err
 	}
-	_, err := s.Resolve(ctx, target.Digest.String())
+	_, err := s.Resolve(withPolicyChecked(ctx), target.Digest.String())
 	if err == nil {
 		return true, nil
 	}


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

Issue #1327, opened 2026-08-23
main is not gofmt-clean (17 files) and CI does not enforce formatting


TerryHowe:
## What

17 files on `main` are not `gofmt`-clean. `v2` is unaffected.

```console
$ gofmt -l .
content/cache/cache_test.go
registry/interfaces.go
registry/interfaces_test.go
registry/remote/config/properties_test.go
registry/remote/config/registries_test.go
registry/remote/credentials/store_test.go
registry/remote/internal/configpaths/current_darwin.go
registry/remote/internal/configpaths/current_freebsd.go
registry/remote/internal/configpaths/current_linux.go
registry/remote/internal/configpaths/current_windows.go
registry/remote/internal/configpaths/uapi_darwin.go
registry/remote/internal/configpaths/uapi_freebsd.go
registry/remote/internal/configpaths/uapi_linux.go
registry/remote/internal/configpaths/uapi_windows.go
registry/remote/policy/policy_other.go
registry/remote/policy/policy_test.go
registry/remote/policy/requirement_test.go
```

Reproduced identically under both Go 1.25.0 and Go 1.26.2, so this is not a
toolchain-version artifact.

The findings fall into three groups:

1. **Build-constraint placement (9 files)** — the 8 `configpaths/` files plus
   `policy/policy_other.go`. `gofmt` wants `//go:build` hoisted above the
   Apache license block:

   ```diff
   +//go:build linux
   +
    /*
    Copyright The ORAS Authors.
    ...
    */
   
   -//go:build linux
   -
    package configpaths
   ```

2. **Stray blank lines (4 files)** — a trailing blank line at EOF (e.g.
   `registry/interfaces.go`), plus one doubled blank line between functions in
   `policy/requirement_test.go`.

3. **Composite-literal field alignment (4 files)** — e.g. in
   `policy/policy_test.go`, `KeyPath`/`KeyData` need an extra space to align
   with the adjacent `KeyDatas` key.

## This is cosmetic only — the build tags do work today

Worth stating up front, since group 1 looks alarming: the constraints **are**
currently honored despite sitting below the license block. `go/build` skips
leading comments when scanning for the preamble, so the files are correctly
excluded per-platform:

```console
$ go list -f '{{.GoFiles}}' ./registry/remote/internal/configpaths   # on darwin
[current.go current_darwin.go resolver.go uapi.go uapi_darwin.go]
```

So there is no platform-selection bug to fix here, only formatting.

## Why it went unnoticed

[...]

PR #1328, opened 2026-08-23
feat(remote): add ClientBuilder for constructing configured registries

TerryHowe:
## What

Extracted from #1130. Adds `ClientBuilder`, which turns a
`*properties.Registry` into a fully configured `*auth.Client`, plus
`NewRegistryWithProperties` / `NewRepositoryWithProperties` built on top of it.

Today, assembling a `Repository` from a config file means wiring the TLS
config, CA pool, transport, retry policy, credential func and headers together
by hand at each call site — and then doing it again for every mirror client.
`ClientBuilder` does that once.

`main` already refers to this type: the doc comment at
`registry/remote/repository.go:156` says a Repository "is constructed via the
ClientBuilder", which does not exist yet. This PR makes that true.

## Commits

1. **`feat(properties): add TokenFlow to select the bearer token flow`** —
   prerequisite. Registries implementing only the distribution-spec token
   endpoint cannot serve the OAuth2 password grant (a Docker extension), and
   there was no way to request the spec'd flow short of supplying a custom
   `TokenFetcher`. Adds `properties.TokenFlow` (`Default`/`OAuth2`/
   `Distribution`) and the ORAS-specific `token-flow` key in `registries.conf`.

2. **`feat(remote): add ClientBuilder for constructing configured registries`**
   — the builder itself.

## Notes for reviewers

**Precedence.** A caller-supplied `ClientBuilder.TokenFetcher` wins over the
`registries.conf` `token-flow` setting: the former is the caller speaking
directly, the latter is a config file.

**An unrecognized `token-flow` is rejected, not ignored.** This intentionally
differs from the neighbouring `referrers-api` key, which falls back to
auto-detection. Silently defaulting a typo'd flow would leave the user
authenticating a way they did not ask for, and the failure surfaces as bad
credentials — a confusing place to start debugging.

**Credential scoping.** The credential func matches the requested host against
the registry the credential was configured for and returns `EmptyCredential`
for anything else, so a credential is never handed to a mirror or to a host
that merely shares a suffix with the configured one
(`registry.example.com.evil.test`). The `docker.io` alias is normalized to
`registry-1.docker.io`, since that is the host the transport actually dials.
`mirror_isolation_test.go` covers all three cases.

## Testing

- `go build ./...`, `go vet ./...`, and `go test ./...` all pass.
- 8 files, **+1478, zero deletions** — purely additive, no existing behaviour
  changed.
- All added files are `gofmt`-clean (relevant given #1327).

## Follow-ups

Unblocks #1284, whose `test/functional/mirror_test.go` uses `ClientBuilder`.

### DIFF
--- a/registry/remote/builder.go
+++ b/registry/remote/builder.go
@@ -1,0 +1,397 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package remote
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/oras-project/oras-go/v3/registry"
+	"github.com/oras-project/oras-go/v3/registry/remote/auth"
+	"github.com/oras-project/oras-go/v3/registry/remote/credentials"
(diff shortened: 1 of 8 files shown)


shizhMSFT (approved):
LGTM

PR #1330, opened 2026-08-24
refactor(auth)!: route bearer token fetching through TokenFetcher

TerryHowe:
## What

Extracted from #1130. Finishes the `TokenFetcher` work started in #1181 by
making it the single path for bearer token acquisition, and closes a gap where
the new path missed the redirect protection the old one had.

Independent of #1328 — that PR touches `builder.go`/`properties`, this one only
`registry/remote/auth/`. `NewCompositeTokenFetcher`'s signature is unchanged,
so the two do not interact.

### 1. `fix(auth): strip credentials on token endpoint redirects`

`Client.send` drops the `Authorization` header when a redirect crosses an HTTP
origin, but token endpoint requests issued through `TokenFetcher` did not —
`sendRequest` called `client.Do` directly. The distribution spec flow sends the
credential to the realm as HTTP Basic, so a redirect away from the realm could
carry it to an unintended endpoint. Same class of leak as
GHSA-vh4v-2xq2-g5cg, which is already fixed for registry requests.

Extracts the wrapper as `redirectSafeClient` and routes both `Client.send` and
`sendRequest` through it.

**No released version is affected.** The `TokenFetcher` path arrived in #1181
on 2026-08-14 and is in no tagged release (the only v3 tags are `v3.0.x-dev`,
all of which predate it), so this never reached users and needs no advisory.
Flagging the reasoning explicitly so it can be double-checked rather than
taken on trust.

### 2. `refactor(auth)!: route bearer token fetching through TokenFetcher`

`Client` carried its own copies of the distribution and OAuth2 flows alongside
the `TokenFetcher` abstraction, so the same two flows existed twice:
`ForceAttemptOAuth2` selected between the built-in pair, while a configured
`TokenFetcher` bypassed both. This deletes `Client.fetchDistributionToken` and
`Client.fetchOAuth2Token` and routes `fetchBearerToken` through
`TokenFetcher`, defaulting to a `CompositeTokenFetcher`.

**Behaviour is unchanged**: anonymous access uses the distribution spec token
endpoint, credentialed access uses OAuth2 with password grant.

## Breaking change

`Client.ForceAttemptOAuth2` is removed. To follow the legacy distribution spec
flow used in v1 and v2:

```go
client.TokenFetcher = auth.NewCompositeTokenFetcher(
	client.Client, client.Header, client.ClientID, true)
```

The removal is contained to the `auth` package — I grepped the tree and no
other package referenced the field.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` all pass (33 packages).
- Net **−168 lines** of duplicated token-fetching logic, with `client_test.go`
  gaining coverage for the routed paths.
- `gofmt`-clean (relevant given #1327).

### DIFF
--- a/registry/remote/auth/client.go
+++ b/registry/remote/auth/client.go
@@ -128,13 +128,24 @@ func (c *Client) send(req *http.Request) (*http.Response, error) {
 	for key, values := range c.Header {
 		req.Header[key] = append(req.Header[key], values...)
 	}
-	// Drop the Authorization header when a redirect crosses an HTTP origin
-	// (scheme, host, or port). The standard library only strips sensitive
-	// headers when the hostname changes, so a redirect to a different port on
-	// the same host would otherwise forward credentials to an unintended
-	// endpoint. Any caller-provided CheckRedirect is preserved.
-	// Reference: https://github.com/oras-project/oras-go/security/advisories/GHSA-vh4v-2xq2-g5cg
-	client := c.client()
+	return redirectSafeClient(c.client()).Do(req)
+}
+
+// redirectSafeClient returns a shallow copy of client whose CheckRedirect drops
+// the Authorization header when a redirect crosses an HTTP origin (scheme,
+// host, or port). The standard library only strips sensitive headers when the
+// hostname changes, so a redirect to a different port on the same host would
+// otherwise forward credentials to an unintended endpoint. Any caller-provided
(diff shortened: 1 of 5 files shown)


sabre1041 (commented):
LGTM

sabre1041 (approved):

PR #1331, opened 2026-08-24
refactor(credentials)!: replace the config loader global with a shared package

TerryHowe:
## What

Extracted from #1130 — the last code change on that branch that did not yet
have a PR of its own.

`config` and `credentials` both need to read the Docker config file, but
`credentials` cannot import `config` without a cycle. That was worked around
with a process-wide registration hook:

- `config` called `SetDefaultConfigLoader` from an `init`,
- `credentials` read that global,
- and anyone who used `credentials` without also importing `config` got
  `ErrNoConfigLoader`.

So whether the library worked depended on which packages the caller happened to
import. `credentials/main_test.go` carried a `TestMain` whose only purpose was
to produce a readable message when that registration had not happened — a guard
against the design rather than a test of behaviour.

This moves the file format into `registry/remote/internal/configfile`, holding
`Config`, `Load` and the `ConfigFile` interface. Both packages import it and
neither imports the other for this purpose, so the dependency lives in the
import graph instead of in `init` ordering.

## Breaking change

`SetDefaultConfigLoader`, `ConfigFileLoader` and `ErrNoConfigLoader` are gone.

**Nothing needs to replace them at a call site.** They existed only to perform
the registration; `credentials` now reads config files directly. If you were
calling `SetDefaultConfigLoader`, delete the call.

`config.Config` remains a type alias for `configfile.Config`, so the exported
surface of `config` is unchanged for anyone using it to read a config file.

## Notes for reviewers

- The bulk of the diff is a move: `config/config_test.go` and the whole
  `config/testdata/` directory relocate next to the code they exercise. Git
  records them as renames, so the real change is smaller than the file count
  suggests — 22 files, +520/−514, and most of that is `config.go` shrinking
  from 428 lines to 49 as its contents move to `configfile.go`.
- Incidentally fixes 3 of the 17 files listed in #1327: `config/properties_test.go`,
  `config/registries_test.go` and `credentials/store_test.go` come out
  `gofmt`-clean, taking the repo from 17 offenders to 14.
- Independent of #1328 and #1330; those touch `builder.go`/`properties` and
  `auth/` respectively, and this touches neither.

## Follow-up

The design note behind this change, `docs/design/config-credentials-coupling.md`,
is in #1107 and describes the design as a *proposal*. It is deliberately not
touched here, since the file does not exist on `main` yet. Once both land, its
status line wants flipping to *implemented*.

## Testing

`go build ./...`, `go vet ./...` and `go test ./...` all pass.


TerryHowe:
Recording the follow-up wording here so it is not lost.

`docs/design/config-credentials-coupling.md` is carried by #1107 with its original *proposal* status, which is accurate against `main` today. Once this PR lands, that note's header wants replacing with:

```markdown
Status: **implemented**.

The shared internal package is `registry/remote/internal/configfile`. It holds
the Docker config file format (`Config`, `Load`) and the `ConfigFile` interface;
`config` and `credentials` both import it and neither imports the other for this
purpose. `SetDefaultConfigLoader`, `ConfigFileLoader` and `ErrNoConfigLoader`
are gone, and the `TestMain` guard in `credentials/main_test.go` that existed to
diagnose the missing registration went with them.

The rest of this note is the original proposal, kept as the rationale.
```

This wording previously lived on the v3 working branch (#1130), which is now code-only.

sabre1041 (approved):
LGTM

PR #1333, opened 2026-08-24
refactor(remote)!: remove unused ErrReferrersCapabilityAlreadySet

### LINKED ISSUE #1324: Remove unreachable ErrReferrersCapabilityAlreadySet from registry/remote
## What

`ErrReferrersCapabilityAlreadySet` in `registry/remote/referrers.go` is exported but never returned by anything in the tree:

```
$ git grep -n ErrReferrersCapabilityAlreadySet -- '*.go'
registry/remote/referrers.go:52: ErrReferrersCapabilityAlreadySet = errors.New(...)
```

`SetReferrersCapability` silently ignores a conflicting set rather than reporting one, and the error's own doc comment describes it as "reserved to signal" — dead public surface. v3 is the moment to drop it rather than carry it for the life of the major release.

## Proposed change

1. Delete the `ErrReferrersCapabilityAlreadySet` var from `registry/remote/referrers.go`.
2. Reword the `SetReferrersCapability` doc comment in `registry/remote/repository.go` so the no-op-on-conflict behaviour reads as deliberate instead of as an oversight, and point callers at `ReferrersCapability` to read back the effective value.

## Why the signature should stay as it is

The obvious companion change would be restoring an `error` return on `SetReferrersCapability` so the value becomes reachable again. That would be the wrong fix: the same setter is also the recorder for auto-detection, called from six internal sites:

```
repository.go:751, 757   Referrers fallback
repository.go:942, 949   pingReferrers
repository.go:1700, 1796 push/delete with indexing
```

Those race by design — concurrent operations each observe what the registry did and try to record it. "First value wins" is correct there, and an error return would be meaningless noise at every one of those call sites.

## Notes for a first-time contributor

- This is a breaking change to exported API, so it targets v3 (`main`) only — do not backport to v2.
- No behaviour change and no new tests needed; verify with `go build ./...`, `go vet ./...`, `go test ./...`, and `golangci-lint run`.

A reference implementation exists in #1320.

aswinsreeraj:
## Description
Resolves #1324 

`ErrReferrersCapabilityAlreadySet` in `registry/remote/referrers.go` is exported but never returned anywhere in the codebase. `SetReferrersCapability` uses `atomic.CompareAndSwapInt32` where the first value set wins and subsequent conflicting calls are deliberately ignored to accommodate concurrent discovery across call sites. 

This PR removes the dead exported variable for the v3 release and clarifies the deliberate "first value set wins" behavior in `SetReferrersCapability`'s doc comment.

## What Changes Were Made?
- Deleted the unused exported variable `ErrReferrersCapabilityAlreadySet` from `registry/remote/referrers.go`.
- Reworded `SetReferrersCapability` doc comment in `registry/remote/repository.go` to explicitly document that:
  - The capability is fixed once set; the first value wins and conflicting later calls are ignored.
  - Callers can invoke `ReferrersCapability` to read back the effective capability.

## How Was This Verified?
- `go build ./...`
- `go vet ./...`
- `go test -v ./registry/remote -run TestRepository_SetReferrersCapability` (passes and verifies first-value-wins / conflicting sets ignored behavior).
- Code formatted with `gofmt`.

## Checklist
- [x] Unused exported variable `ErrReferrersCapabilityAlreadySet` removed from `registry/remote/referrers.go`
- [x] Doc comment for `SetReferrersCapability` updated to clarify deliberate behavior and mention `ReferrersCapability`
- [x] Existing tests pass (`go test -v ./registry/remote -run TestRepository_SetReferrersCapability`)
- [x] Code passes `go build ./...` and `go vet ./...`

### DIFF
--- a/registry/remote/referrers.go
+++ b/registry/remote/referrers.go
@@ -46,11 +46,6 @@ type referrerChange struct {
 }
 
 var (
-	// ErrReferrersCapabilityAlreadySet is reserved to signal that the referrers
-	// capability of a repository has already been set to a conflicting value.
-	// The capability is fixed once set; the first value wins.
-	ErrReferrersCapabilityAlreadySet = errors.New("referrers capability cannot be changed once set")
-
 	// errNoReferrerUpdate is returned by applyReferrerChanges() when there
 	// is no any referrer update.
 	errNoReferrerUpdate = errors.New("no referrer update")
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -336,9 +336,9 @@ func (r *Repository) skipReferrersGC() bool {
 // SetReferrersCapability indicates the Referrers API capability of the remote
 // repository. true: capable; false: not capable.
 //
-// SetReferrersCapability has no effect if the capability has already been set
-// to the same value. If the capability has been set to a conflicting value it
-// is silently ignored; the first value set wins.
+// The capability is fixed once set; the first value set wins and conflicting
+// later calls are ignored. Callers can use ReferrersCapability to read back
+// the effective capability.
 //   - When the capability is set to true, the Referrers() function will always
 //     request the Referrers API. Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#listing-referrers
 //   - When the capability is set to false, the Referrers() function will always


TerryHowe (approved):
/lgtm

PR #1340, opened 2026-08-25
test: add coverage for programmatic policy construction API

### LINKED ISSUE #1339: Add test coverage for the programmatic policy construction API
## Summary

Five exported functions in `registry/remote/policy` — the whole programmatic policy-construction API — have no test coverage at all. `policy_test.go` is otherwise thorough (20+ tests covering loading, validation, evaluation and JSON round-tripping), so this is a clean gap rather than a neglected package.

| Function | `policy.go` | Coverage |
|---|---|---|
| `NewPolicy` | 66 | 0.0% |
| `NewInsecureAcceptAnythingPolicy` | 75 | 0.0% |
| `NewRejectAllPolicy` | 83 | 0.0% |
| `(*Policy) SetDefault` | 90 | 0.0% |
| `(*Policy) SetTransportScope` | 96 | 0.0% |

Measured against `725b6d8`, two ways, to rule out a per-package artifact:

```
go test ./registry/remote/policy/ -coverprofile=p1.out
go test ./... -coverpkg=./registry/remote/policy/ -coverprofile=p2.out
```

Both report 0.0% for all five — nothing anywhere in the repo's test suite calls them, including the examples in `example_test.go`, which build `Policy` values with struct literals instead.

## Why it is worth covering

These are the fluent builder API a caller reaches for when constructing a policy without a `policy.json` file, and two of them have behavior that is easy to get wrong and currently unpinned:

**1. `NewPolicy()` returns a policy that fails `Validate()`.** Its `Default` is `make(PolicyRequirements, 0)`, and `Validate` rejects an empty default:

```go
if len(p.Default) == 0 {
	return fmt.Errorf("default policy requirements must not be empty")
}
```

That is correct per the containers-policy.json spec — the global default is mandatory — but the doc comment ("Use this for programmatic policy construction without a file") does not say the result is unusable until `SetDefault` is called. A test should pin both halves: `NewPolicy()` alone is invalid, `NewPolicy().SetDefault(...)` is valid.

**2. `SetDefault()` with no arguments sets `Default` to nil**, since a variadic with zero args yields a nil slice rather than an empty one. That also fails `Validate`, and silently clears a previously-set default. Worth pinning so nobody "fixes" it into a no-op.

**3. `SetTransportScope` has a nil-map guard** that is only reachable from the constructors which leave `Transports` nil:

```go
if p.Transports == nil {
	p.Transports = make(map[TransportName]TransportScopes)
}
```

`NewPolicy()` pre-populates `Transports`, but `NewInsecureAcceptAnythingPolicy()` and `NewRejectAllPolicy()` do not — so `NewRejectAllPolicy().SetTransportScope(...)` is the case that exercises it.

## Suggested test cases

In `registry/remote/policy/policy_test.go`, matching the existing table-driven style:

- `NewPolicy()` — `Default` and `Transports` both non-nil and empty; `Validate()` returns an error.
- `NewInsecureAcceptAnythingPolicy()` — `Default` holds exactly one `*InsecureAcceptAnything`; `Validate()` passes; `GetRequirementsForImage` returns it for an arbitrary scope.
- `NewRejectAllPolicy()` — same shape with `*Reject`.
- `SetDefault` — replaces existing requirements rather than appending; returns the receiver so chaining works; zero-arg call leaves the policy invalid.
- `SetTransportScope` — creates the `Transports` map when nil (start from `NewRejectAllPolicy()`), creates the inner `TransportScopes` when absent, overwrites an existing scope, and returns the receiver.
- A round-trip: build a policy with the fluent API, `Save` it, `LoadPolicy` it back, and assert it matches — this ties the builders to the existing `TestPolicy_SaveAndLoadPolicy` path.

## Getting started

```sh
go test ./registry/remote/policy/ -coverprofile=cover.out
go tool cover -func=cover.out | grep policy.go
```

Existing conventions to follow are in `policy_test.go`; `TestPolicy_Validate` and `TestPolicy_SaveAndLoadPolicy` are the closest neighbours.

tasodoufu:
## Summary

Adds unit test coverage for the programmatic policy construction API in `registry/remote/policy`, which previously had 0% coverage (reported in #1339):

- `NewPolicy` — pins that a fresh policy is invalid until a default is set (`Validate()` rejects an empty default), and valid after `SetDefault`.
- `NewInsecureAcceptAnythingPolicy` — default holds exactly one `*InsecureAcceptAnything`, `Validate()` passes, `GetRequirementsForImage` returns it.
- `NewRejectAllPolicy` — same shape with `*Reject`.
- `(*Policy) SetDefault` — replaces existing requirements rather than appending, returns the receiver for chaining, and a zero-arg call leaves the policy invalid (nil default).
- `(*Policy) SetTransportScope` — creates the `Transports` map when nil (the path reachable from `NewRejectAllPolicy`), creates the inner `TransportScopes` map when absent, overwrites an existing scope, and returns the receiver.
- `TestPolicy_BuildersRoundTrip` — builds a policy with the fluent API, `Save`s it, `LoadPolicy`s it back, and asserts the result matches.

## Validation

- New tests: all pass (`go test ./registry/remote/policy/ -run 'TestNewPolicy|TestNewInsecureAcceptAnythingPolicy|TestNewRejectAllPolicy|TestPolicy_SetDefault|TestPolicy_SetTransportScope|TestPolicy_BuildersRoundTrip'`)
- Full package: `ok github.com/oras-project/oras-go/v3/registry/remote/policy` with race detector
- Coverage of the five functions: 100% each (package total 94.1%)
- `go vet` clean, `gofmt` clean
- `go test ./registry/...` passes for all sub-packages

## Notes / not verified

- Full `make test` (all modules, vendor + encoding checks) was not run locally; only the `registry/...` tree and the policy package were exercised. The change is additive test-only code in one file and cannot affect runtime behavior.
- AI assistance was used to draft and verify this change, per the issue's request for added coverage.

Closes #1339

### DIFF
--- a/registry/remote/policy/policy_test.go
+++ b/registry/remote/policy/policy_test.go
@@ -936,3 +936,183 @@ func TestPolicyRequirements_UnmarshalJSON_Errors(t *testing.T) {
 		})
 	}
 }
+
+func TestNewPolicy(t *testing.T) {
+	tests := []struct {
+		name    string
+		policy  *Policy
+		wantErr bool
+	}{
+		{
+			name:    "new policy is invalid until a default is set",
+			policy:  NewPolicy(),
+			wantErr: true,
+		},
+		{
+			name:    "policy with default set is valid",
+			policy:  NewPolicy().SetDefault(&InsecureAcceptAnything{}),
+			wantErr: false,
+		},
+		{
+			name:    "policy with reject default is valid",
+			policy:  NewPolicy().SetDefault(&Reject{}),
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.policy.Default == nil {
+				t.Error("Default should not be nil")
+			}
+			if tt.policy.Transports == nil {
+				t.Error("Transports should not be nil")
+			}
+			if len(tt.policy.Default) != 0 && tt.wantErr {
+				t.Errorf("Default should be empty, got %d entries", len(tt.policy.Default))
+			}
+			err := tt.policy.Validate()
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Validate() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
+
+func TestNewInsecureAcceptAnythingPolicy(t *testing.T) {
+	p := NewInsecureAcceptAnythingPolicy()
+
+	if len(p.Default) != 1 {
+		t.Fatalf("Default should have exactly 1 requirement, got %d", len(p.Default))
+	}
+	if p.Default[0].Type() != TypeInsecureAcceptAnything {
+		t.Errorf("Default[0].Type() = %s, want %s", p.Default[0].Type(), TypeInsecureAcceptAnything)
+	}
+	if err := p.Validate(); err != nil {
+		t.Errorf("Validate() error = %v, want nil", err)
+	}
+
+	reqs := p.GetRequirementsForImage(TransportNameDocker, "docker.io/library/nginx")
+	if len(reqs) != 1 || reqs[0].Type() != TypeInsecureAcceptAnything {
+		t.Errorf("GetRequirementsForImage() = %v, want single %s", reqs, TypeInsecureAcceptAnything)
+	}
+}
+
+func TestNewRejectAllPolicy(t *testing.T) {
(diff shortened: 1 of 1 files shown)


TerryHowe:
DCO failure https://github.com/oras-project/oras-go/pull/1340/checks?check_run_id=97811410160

TerryHowe:
Dup

PR #1343, opened 2026-08-25
test: cover default certs.d discovery

### LINKED ISSUE #1342: Cover the default certs.d search-path wiring in LoadCertsDir
## Summary

`LoadCertsDir` and `defaultCertsDirPaths` in `registry/remote/config/certsd.go` have no test coverage. This is a small gap — the certificate *discovery* logic underneath them is already well covered — but the wiring between the exported entry point and the default search paths is not exercised anywhere.

Measured against `cf23175` with `-coverpkg` across the whole repo, so these are not per-package artifacts:

| Function | `certsd.go` | Coverage |
|---|---|---|
| `defaultCertsDirPaths` | 30 | **0.0%** |
| `defaultCertsDirPathsWithStrategy` | 36 | 100.0% |
| `LoadCertsDir` | 61 | **0.0%** |
| `LoadCertsDirFromPaths` | 68 | 92.3% |
| `(*CertsDir) ApplyToTransport` | 120 | 100.0% |

`configpaths.CertsDirPaths` is at 100% for both the `current` and `uapi` resolvers, and `certsd_test.go` covers `LoadCertsDirFromPaths` thoroughly against explicit base directories.

## What is actually missing

Both uncovered functions are one-line wrappers:

```go
func defaultCertsDirPaths() []string {
	return defaultCertsDirPathsWithStrategy(StrategyContainersImage)
}

func LoadCertsDir(host string) (*CertsDir, error) {
	return LoadCertsDirFromPaths(host, defaultCertsDirPaths())
}
```

So the risk they carry is not in the logic but in the wiring: nothing asserts that `LoadCertsDir` — the function callers actually reach for — consults the directories its doc comment promises:

```
//  1. /etc/containers/certs.d/<host>/
//  2. $HOME/.config/containers/certs.d/<host>/
```

A caller could have their certs silently not found if that wiring regressed (wrong strategy, wrong relative directory), and every existing test would still pass, because they all supply base directories explicitly.

## Suggested test

The user-level path is reachable in a test — `configpaths` resolves it via `os.UserHomeDir()`, and `resolver_test.go` already uses `t.Setenv("HOME", ...)` for exactly this:

```go
func TestLoadCertsDir_UserPath(t *testing.T) {
	home := t.TempDir()
	t.Setenv("HOME", home)
	t.Setenv("XDG_CONFIG_HOME", "")

	hostDir := filepath.Join(home, ".config", "containers", "certs.d", "example.com")
	if err := os.MkdirAll(hostDir, 0o700); err != nil {
		t.Fatal(err)
	}
	// write ca.crt / client.cert / client.key into hostDir

	cd, err := LoadCertsDir("example.com")
	// assert the CA cert, client cert and key were discovered
}
```

Worth covering:

- A host directory under `$HOME/.config/containers/certs.d/` is found, with its `.crt`, `.cert` and matching `.key` picked up.
- An unknown host returns `(nil, nil)` rather than an error.
- `defaultCertsDirPaths()` returns the `StrategyContainersImage` paths — i.e. it agrees with `defaultCertsDirPathsWithStrategy(StrategyContainersImage)`.

Note the system-wide path (`/etc/containers/certs.d`) is not controllable from a unit test, and on non-Linux platforms it is not part of the search set at all — so assert on the user-level path only, and keep the test platform-independent.

## Getting started

```sh
go test ./registry/remote/config/ -coverprofile=cover.out
go tool cover -func=cover.out | grep certsd.go
```

Existing conventions are in `registry/remote/config/certsd_test.go`; `registry/remote/internal/configpaths/resolver_test.go` shows the `t.Setenv` pattern for controlling the resolved home directory.

nightcityblade:
## What

Adds test coverage for the default `certs.d` search-path wiring:

- verifies `defaultCertsDirPaths` uses `StrategyContainersImage`
- verifies `LoadCertsDir` discovers CA, client certificate, and key files from the user-level default directory
- verifies an unknown host returns `(nil, nil)`

The test overrides both Unix and Windows home-directory environment variables so the user path remains platform-independent.

## Why

The existing discovery tests pass explicit base directories to `LoadCertsDirFromPaths`. They did not exercise the exported `LoadCertsDir` entry point or pin its default strategy, so a regression in that wiring could silently skip the documented user certificate directory.

## Testing

- `go test ./registry/remote/config/ -coverprofile=/tmp/oras-go-1342-cover.out`
- `go test -race ./registry/remote/config/...`
- `go vet ./registry/remote/config/...`
- `GOFLAGS=-mod=mod golangci-lint run`

Both previously uncovered wrappers now report 100% coverage. No production code was changed.

Fixes #1342

### DIFF
--- a/registry/remote/config/certsd_test.go
+++ b/registry/remote/config/certsd_test.go
@@ -24,6 +24,52 @@ import (
 	"github.com/oras-project/oras-go/v3/registry/remote/properties"
 )
 
+func TestDefaultCertsDirPaths(t *testing.T) {
+	got := defaultCertsDirPaths()
+	want := defaultCertsDirPathsWithStrategy(StrategyContainersImage)
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("defaultCertsDirPaths() = %v, want %v", got, want)
+	}
+}
+
+func TestLoadCertsDir(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	t.Setenv("USERPROFILE", home)
+	t.Setenv("XDG_CONFIG_HOME", "")
+
+	host := filepath.Base(home) + ".invalid"
+	hostDir := filepath.Join(home, ".config", "containers", "certs.d", host)
+	if err := os.MkdirAll(hostDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+	caCertPath := filepath.Join(hostDir, "ca.crt")
+	clientCertPath := filepath.Join(hostDir, "client.cert")
+	clientKeyPath := filepath.Join(hostDir, "client.key")
+	for _, path := range []string{caCertPath, clientCertPath, clientKeyPath} {
+		if err := os.WriteFile(path, []byte("test data"), 0600); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	got, err := LoadCertsDir(host)
+	if err != nil {
+		t.Fatalf("LoadCertsDir() error: %v", err)
+	}
+	want := &CertsDir{CACertPaths: []string{caCertPath}, ClientCert: clientCertPath, ClientKey: clientKeyPath}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("LoadCertsDir() = %v, want %v", got, want)
+	}
+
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

PR #1345, opened 2026-08-25
refactor(registry)!: include Untagger in ManifestStore interface

### LINKED ISSUE #1344: Remove unreachable content.Untagger assertion in Repository.Untag
### What

`Repository.Untag` (added in #1217, merged as 138b906) guards its delegation to the manifest store behind an optional-interface assertion that can never fail:

https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L563-L574

```go
manifests := r.Manifests()
untagger, ok := manifests.(content.Untagger)
if !ok {
	return fmt.Errorf("manifest store %T does not implement content.Untagger", manifests)
}
return untagger.Untag(withPolicyChecked(ctx), reference)
```

`Manifests()` returns `&manifestStore{repo: r}` unconditionally, and `*manifestStore` implements `content.Untagger`, so `ok` is always `true`. The `!ok` branch is unreachable — it is the only uncovered block in the new code (`Repository.Untag` sits at 85.7% coverage while `manifestStore.Untag` is at 95.0%), and it cannot be covered by a test.

### Why it is shaped that way

This was discussed during review of #1217. Every sibling method delegates directly:

```go
func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference string) error {
	if err := r.checkPolicy(ctx, reference); err != nil {
		return err
	}
	return r.Manifests().Tag(withPolicyChecked(ctx), desc, reference)
}
```

`Untag` cannot do the same because `registry.ManifestStore` does not include `content.Untagger`:

```go
type ManifestStore interface {
	BlobStore
	content.Tagger
	ReferencePusher
}
```

The assertion was the conservative way to land the feature without changing that interface.

### Proposal

Add `content.Untagger` to `registry.ManifestStore` and delegate directly, matching `Tag` and `PushReference`:

```go
type ManifestStore interface {
	BlobStore
	content.Tagger
	content.Untagger
	ReferencePusher
}
```

```go
func (r *Repository) Untag(ctx context.Context, reference string) error {
	if err := r.checkPolicy(ctx, reference); err != nil {
		return err
	}
	return r.Manifests().Untag(withPolicyChecked(ctx), reference)
}
```

This removes the unreachable branch and the untestable line, and makes `Untag` reachable through the `ManifestStore` interface rather than only via a type assertion on the concrete type.

**This is a breaking change for external implementors of `registry.ManifestStore`.** v3 is still pre-release (latest tag `v3.0.2-dev`), so this seems like the right window to take it — but that is the call to make here.

If the interface change is not wanted, the fallback would be to at least wrap the error in `errdef.ErrUnsupported` instead of a bare `fmt.Errorf`, so callers can detect the condition programmatically. That leaves the branch unreachable and uncovered, though.

### Related

- #1217 — added `Untag` for remote registries
- #647 — introduced `content.Untagger`
- #1028 — original feature request

aswinsreeraj:
### Description
Resolves #1344 

`Repository.Untag` (added in #1217) previously guarded its delegation to the manifest store behind a runtime type assertion:

```go
manifests := r.Manifests()
untagger, ok := manifests.(content.Untagger)
if !ok {
    return fmt.Errorf("manifest store %T does not implement content.Untagger", manifests)
}
return untagger.Untag(withPolicyChecked(ctx), reference)
```

Because `Manifests()` returns `*manifestStore` which implements `content.Untagger`, `ok` was always `true`. This left the `!ok` error branch completely unreachable and untestable.

`registry.ManifestStore` already embeds `content.Tagger`. This PR adds `content.Untagger` directly to `registry.ManifestStore`, bringing symmetry to reference tag management and allowing `Repository.Untag` to delegate directly to `r.Manifests().Untag(...)` matching its sibling methods (`Tag`, `PushReference`).

### What Changes Were Made?

- Embedded `content.Untagger` into `registry.ManifestStore` and updated its doc comment in `registry/interfaces.go`.
- Simplified `Repository.Untag` in `registry/remote/repository.go` by removing the unreachable type assertion and directly calling `r.Manifests().Untag(withPolicyChecked(ctx), reference)`.
- Added policy-enforced `Untag` to `policyEnforcingManifestStore` in `registry/remote/middleware.go`.
- Updated `mockManifestStore` and added unit test assertions for `Untag` under allowed and denied policies in `registry/remote/middleware_test.go`.
- Removed unnecessary `.(content.Untagger)` casts in `registry/remote/repository_policy_test.go` and `registry/remote/repository_test.go`, and cleaned up unused imports.

### How Was This Verified?

- `go test -v ./registry/remote -run "TestRepository_Untag|TestManifestStoreInterface|Test_ManifestStore_Untag|TestWithPolicyEnforcement"` (all pass).
- Coverage checked with `go test -coverprofile`: `Repository.Untag` statement coverage increased to **100.0%** (from 85.7%).
- `make check-encoding` passed without CRLF or formatting issues.
- `go build ./...` and `go vet ./...` pass cleanly.

### Checklist

- [x] `content.Untagger` added to `registry.ManifestStore` in `registry/interfaces.go`
- [x] `Repository.Untag` simplified to delegate directly without defensive type assertion
- [x] `policyEnforcingManifestStore` updated to implement `Untag` with policy checks
- [x] Unit tests and mock implementations updated in `middleware_test.go`, `repository_policy_test.go`, and `repository_test.go`
- [x] All tests pass and `Repository.Untag` reaches 100% statement coverage
- [x] Code passes `go build ./...`, `go vet ./...`, and `make check-encoding`

### DIFF
--- a/registry/interfaces.go
+++ b/registry/interfaces.go
@@ -62,10 +62,11 @@ type BlobStore interface {
 }
 
 // ManifestStore is a CAS with the ability to stat and delete its content.
-// Besides, ManifestStore provides reference tagging.
+// Besides, ManifestStore provides reference tagging and untagging.
 type ManifestStore interface {
 	BlobStore
 	content.Tagger
+	content.Untagger
 	ReferencePusher
 }
 
--- a/registry/remote/middleware.go
+++ b/registry/remote/middleware.go
@@ -301,6 +301,14 @@ func (s *policyEnforcingManifestStore) Tag(ctx context.Context, desc ocispec.Des
 	return s.manifests.Tag(ctx, desc, reference)
 }
 
+// Untag removes a tag reference with policy enforcement.
+func (s *policyEnforcingManifestStore) Untag(ctx context.Context, reference string) error {
+	if err := s.repo.checkPolicy(ctx, reference); err != nil {
+		return err
+	}
+	return s.manifests.Untag(ctx, reference)
+}
+
 // Ensure the wrappers satisfy the registry interfaces they enforce policy on.
 var (
 	_ registry.Repository    = (*policyEnforcingRepository)(nil)
--- a/registry/remote/middleware_test.go
+++ b/registry/remote/middleware_test.go
@@ -275,6 +275,12 @@ func TestWithPolicyEnforcement_Manifests(t *testing.T) {
 	if err != nil {
 		t.Errorf("Manifests().Tag() error = %v, want nil", err)
 	}
+
+	// Test Untag
+	err = manifests.Untag(ctx, "latest")
+	if err != nil {
+		t.Errorf("Manifests().Untag() error = %v, want nil", err)
+	}
 }
 
 func TestWithPolicyEnforcement_PolicyDenied_AllOperations(t *testing.T) {
@@ -355,6 +361,9 @@ func TestWithPolicyEnforcement_PolicyDenied_AllOperations(t *testing.T) {
 	}
 	if err := manifests.Tag(ctx, desc, "latest"); err == nil {
 		t.Error("Manifests().Tag() should return error when policy denies access")
+	}
+	if err := manifests.Untag(ctx, "latest"); err == nil {
+		t.Error("Manifests().Untag() should return error when policy denies access")
 	}
 }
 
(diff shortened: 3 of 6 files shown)


TerryHowe (approved):
/lgtm

PR #1349, opened 2026-08-28
feat(remote): separate registry warning logging from HTTP debug logging

### LINKED ISSUE #1347: Separate registry warning logging from HTTP debug logging in ClientBuilder
## Summary

`ClientBuilder` already attaches a warning logger — but only as a side effect of `ClientBuilder.Logger`, which primarily exists to enable HTTP request/response debug logging. The two concerns share one field, so you cannot have one without the other.

`registry/remote/builder.go:298`:

```go
if builder.Logger != nil {
    reg.HandleWarning = NewWarningLogger(props.Reference.Registry, builder.Logger)
}
```

and `builder.go:209`, in the same construction path:

```go
if b.Logger != nil {
    transport = NewLoggingTransport(transport, b.Logger)
}
```

Registry warnings and HTTP debug logging have different audiences. A deprecation warning at `slog.LevelWarn` is something you want on in production; full request/response logging at `slog.LevelDebug` is a troubleshooting tool. Coupling them means opting into the first drags in the second.

## Why the level filter is not enough

The obvious answer — set `Logger` to a handler at `LevelInfo`, so the debug records are dropped — does not avoid the cost. `LoggingTransport.RoundTrip` has no `Enabled()` guard (`registry/remote/logging_transport.go:74`), and builds its arguments eagerly:

```go
t.logger.Debug("Response",
    "id", id,
    "status", resp.Status,
    "header", formatHeaders(resp.Header),
    "body", formatResponseBody(resp),   // <-- always runs
)
```

`formatResponseBody` reads up to 16 KiB of the response body and reconstructs it so the caller still sees the full stream. That happens on **every** response, on every retry attempt, even when the handler discards the record. Anyone who sets `Logger` purely to get warnings pays that on the whole copy path.

### 1. Add a dedicated field

```go
type ClientBuilder struct {
    // ...

    // Logger enables HTTP request/response debug logging when non-nil.
    Logger *slog.Logger

    // WarningLogger logs registry warnings (RFC 7234 Warning headers) at
    // slog.LevelWarn, deduplicated per registry. If nil, Logger is used;
    // if both are nil, warnings are not logged.
    WarningLogger *slog.Logger
}
```

Falling back to `Logger` keeps today's behaviour for existing callers — setting only `Logger` still gets warnings — so this is additive and non-breaking.

### 2. Route both construction sites through one helper

```go
func (b *ClientBuilder) warningLogger() *slog.Logger {
    if b.WarningLogger != nil {
        return b.WarningLogger
    }
    return b.Logger
}
```

Two call sites need updating, and they must stay in step — mirrors are separate `Registry` values built by a different function:

- `NewRegistryWithProperties`, `builder.go:298` — the primary registry.
- `buildMirrorRepositories`, `builder.go:378` — each mirror, which correctly passes the *mirror's* location rather than the primary's.

### 3. Guard the logging transport

Independently worth fixing, and it is what makes the separation actually pay off:

```go
func (t *LoggingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
    if !t.logger.Enabled(req.Context(), slog.LevelDebug) {
        return t.inner.RoundTrip(req)
    }
    // ... existing logging path
}
```

Note this also skips the request-ID increment on the fast path, which is fine — the IDs only need to correlate records that are actually emitted.

## Tests

- `WarningLogger` set and `Logger` nil: warnings logged, no `LoggingTransport` in the stack.
- `Logger` set and `WarningLogger` nil: current behaviour preserved (both active).
- Both set to different loggers: each record lands in the right one.
- Mirror registries get a warning logger keyed to the mirror's own location, not the primary's.
- `LoggingTransport` with a handler that reports `Enabled() == false` for `LevelDebug`: response body is not consumed, and the caller still reads the complete body.

## Docs

- Field comments on `ClientBuilder.Logger` and the new field, saying explicitly what each covers.
- `docs/SCENARIOS.md` scenario 13 presents `ClientBuilder.Logger` as the debug-logging switch without mentioning that it also turns on warnings; scenario 11 shows the manual `HandleWarning` wiring, which stays correct for hand-built repositories but should point at the builder field as the easier path.

## Notes

`builder.go`, `warning.go`, `registry.go`, and `logging_transport.go` are currently byte-identical between `main` and the v3 working branch, so this can be implemented on `main` and merges forward without conflict.

This came out of checking the usage-scenarios doc in #1107 against the implementation.

jjj-n:
## What changed

- add `ClientBuilder.WarningLogger` while retaining `Logger` as the backward-compatible fallback
- route warning handlers for both primary and mirror registries through the dedicated warning logger
- bypass `LoggingTransport` formatting and body inspection when debug logging is disabled
- cover independent logger routing, fallback behavior, mirror attribution, and the disabled-debug fast path

## Why

Registry warnings are production-facing signals, while full HTTP request/response debug logs are a troubleshooting feature. Separating the loggers lets callers enable warnings without enabling the logging transport or paying its response-body inspection cost.

## Testing

- `go test -race ./...` (Go 1.25.14, Linux container, non-root)
- `golangci-lint v2.11.4`: `govet` and `ineffassign`, 0 issues
- `gofmt` and `git diff --check`

`docs/SCENARIOS.md` referenced by the issue is not present on `main`, so this PR does not add a scenario-document update.

Fixes #1347

### DIFF
--- a/registry/remote/builder.go
+++ b/registry/remote/builder.go
@@ -69,8 +69,14 @@ type ClientBuilder struct {
 
 	// Logger enables HTTP request/response debug logging when non-nil.
 	// Each retry attempt is logged individually. If nil, no logging transport
-	// is added. Use slog.Default() to log to the default handler.
+	// is added. Logger is also used for registry warnings when WarningLogger is
+	// nil. Use slog.Default() to log to the default handler.
 	Logger *slog.Logger
+
+	// WarningLogger logs registry warnings at slog.LevelWarn, deduplicated per
+	// registry. If nil, Logger is used for backward compatibility. If both are
+	// nil, warnings are not logged.
+	WarningLogger *slog.Logger
 }
 
 // NewClientBuilder creates a new ClientBuilder with default settings.
@@ -213,6 +219,13 @@ func (b *ClientBuilder) buildTransport(tlsConfig *tls.Config) http.RoundTripper 
 	return transport
 }
 
+func (b *ClientBuilder) warningLogger() *slog.Logger {
+	if b.WarningLogger != nil {
+		return b.WarningLogger
+	}
+	return b.Logger
+}
+
 // buildHTTPClient creates an HTTP client with the given transport.
 func (b *ClientBuilder) buildHTTPClient(transport http.RoundTripper) *http.Client {
 	return &http.Client{
@@ -295,8 +308,8 @@ func NewRegistryWithProperties(props *properties.Registry, builder *ClientBuilde
 		Policy:    builder.PolicyEvaluator,
 	}
 
-	if builder.Logger != nil {
-		reg.HandleWarning = NewWarningLogger(props.Reference.Registry, builder.Logger)
+	if logger := builder.warningLogger(); logger != nil {
+		reg.HandleWarning = NewWarningLogger(props.Reference.Registry, logger)
 	}
 
 	return reg, nil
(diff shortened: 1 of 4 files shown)


TerryHowe (approved):
/lgtm

PR #1350, opened 2026-08-28
feat: identify the failing content in CopyError

### LINKED ISSUE #1346: CopyError does not identify which descriptor failed
## Summary

`oras.CopyError` records the operation that failed (`Op`), the side it failed on (`Origin`), and the underlying error (`Err`), but not *which* content it failed on. For a graph copy of a multi-layer image, `failed to perform "Fetch" on source: unexpected EOF` does not tell the caller which blob is the problem.

## Current shape

```go
type CopyError struct {
	Op     string
	Origin CopyErrorOrigin
	Err    error
}
```

## Proposal

Add an optional descriptor field:

```go
type CopyError struct {
	Op     string
	Origin CopyErrorOrigin
	// Descriptor is the content the operation was acting on. It is the zero
	// value for errors raised before a specific node was selected.
	Descriptor ocispec.Descriptor
	Err        error
}
```

This is additive and non-breaking. Of the 24 `newCopyError` call sites in `copy.go` and `extendedcopy.go`, **13 already have a descriptor in scope**:

- `copy.go` — 11 of 17: `desc` at the `Exists`, `FindSuccessors`, `Mount`, `Fetch`, `Push`, and `PushReference` sites, and `root` at `MapRoot`, `Successors`, and the two `Tag` sites.
- `extendedcopy.go` — 2 of 7: `node` at `Tag` and `currentNode` at `FindPredecessors`.

The other 11 have nothing meaningful to record and would take the zero value:

- The four nil source/destination target guards, and the two `CopyGraph` / `ExtendedCopyGraph` guards — no copy has begun.
- The `Resolve` and `FetchReference` sites in both files — these fail *before* a descriptor exists, so only the reference string is known. If per-site context matters there, a separate reference field would be the way to cover them, but that is a bigger change and probably belongs in its own discussion.

Scope note: `copyerror.go`, `copy.go`, and `extendedcopy.go` are currently byte-identical between `main` and the v3 working branch, so this can be implemented on `main` and will merge forward without conflict. `copyerror.go` imports only `fmt` today and would need `ocispec` added, which is already a direct dependency of the package.

`Error()` would include the digest when set, and callers doing `errors.As(err, &copyErr)` could log or retry per-blob.

## Motivation

This came out of reviewing the usage-scenarios doc in #1107, whose error-handling example assumed a `Descriptor` field existed. The example was wrong and has been corrected, but the underlying ergonomic gap seemed worth raising: "which blob failed" is the first thing you want when a large copy fails partway.

prakrititz:
## What this PR does

Closes #1346.

`CopyError` records the operation that failed, the side it failed on and the
underlying error, but not *which* content the operation was acting on. For a
graph copy of a multi-layer image, `failed to perform "Fetch" on source:
unexpected EOF` does not tell the caller which blob is the problem.

This adds an optional `Descriptor` field to `CopyError` and threads it through
the `newCopyError` call sites that already have a descriptor, root or node in
scope, exactly as proposed in the issue:

```go
type CopyError struct {
	Op     string
	Origin CopyErrorOrigin
	// Descriptor is the content that the operation was acting on. It is the
	// zero value for errors raised before a specific node was selected.
	Descriptor ocispec.Descriptor
	Err        error
}
```

Call sites, matching the audit in the issue:

- `copy.go` — 11 of 17 sites get a real descriptor (`Exists`, `FindSuccessors`,
  `Mount`, `Fetch` x2, `Push`, `PushReference` take `desc`; `MapRoot`,
  `Successors` and both `Tag` sites take `root`). The remaining 6 — the nil
  target guards, the `CopyGraph` guards, `Resolve` and `FetchReference` — pass
  the zero value because they fail before a node is selected.
- `extendedcopy.go` — 2 of 7 sites (`Tag` takes `node`, `FindPredecessors`
  takes `currentNode`); the other 5 pass the zero value.

`Error()` includes the digest when the descriptor is set:

```
failed to perform "Fetch" on source for sha256:6ae8a755...: unexpected EOF
```

and renders **exactly** as before when it is not, so existing messages and the
`ExampleCopyError` output block are unchanged.

As the issue notes, the `Resolve` / `FetchReference` case — where a reference
string rather than a descriptor is the useful context — is deliberately left
out of scope; no reference field is added here.

## Compatibility

Additive and non-breaking. No existing field is renamed, retyped or removed.
Every `CopyError{...}` composite literal in the repo is keyed, so inserting
`Descriptor` between `Origin` and `Err` breaks nothing. `newCopyError` is
unexported, so the added parameter is internal.

## Tests

- `copyerror_test.go` — descriptor case added to `TestNewCopyError`; three new
  `TestCopyError_Error` cases (source / destination / undefined origin, all with
  a descriptor); new `TestCopyError_Descriptor` covering `errors.As` exposure
  and the zero-value message being unchanged.
- `copy_test.go` — `TestCopy_CopyError_Descriptor`: end-to-end through
  `oras.CopyGraph` (`Exists` site) and `oras.Copy` (`Tag` site, proving `root`
  threading), plus a nil-source-target case asserting the descriptor is the zero
  value and the message is byte-identical to before.
- `extendedcopy_test.go` — `TestExtendedCopy_CopyError_Descriptor`, proving the
  `node` threading at the `ExtendedCopy` `Tag` site.

Root-package coverage 89.1% -> 89.2%; `go test -race ./...`, `go vet ./...` and
`gofmt -l .` are clean.

## Note for reviewers

The `for <digest>` wording in `Error()` is my choice — the issue only asked that
the digest be included. Happy to change it to a plain `: <digest>` separator (as
in `copy.go`'s `failed to check cache existence: %s: %w`) if you prefer; it is
one `fmt.Sprintf` per branch.

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -138,10 +138,10 @@ type CopyGraphOptions struct {
 // Returns the descriptor of the root node on successful copy.
 func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, dstRef string, opts CopyOptions) (ocispec.Descriptor, error) {
 	if src == nil {
-		return ocispec.Descriptor{}, newCopyError("Copy", CopyErrorOriginSource, errors.New("nil source target"))
+		return ocispec.Descriptor{}, newCopyError("Copy", CopyErrorOriginSource, ocispec.Descriptor{}, errors.New("nil source target"))
 	}
 	if dst == nil {
-		return ocispec.Descriptor{}, newCopyError("Copy", CopyErrorOriginDestination, errors.New("nil destination target"))
+		return ocispec.Descriptor{}, newCopyError("Copy", CopyErrorOriginDestination, ocispec.Descriptor{}, errors.New("nil destination target"))
 	}
 
 	// Run policy check before copy if configured.
@@ -169,7 +169,7 @@ func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, ds
 		proxy.StopCaching = true
 		root, err = opts.MapRoot(ctx, proxy, root)
 		if err != nil {
-			return ocispec.Descriptor{}, newCopyError("MapRoot", CopyErrorOriginSource, err)
+			return ocispec.Descriptor{}, newCopyError("MapRoot", CopyErrorOriginSource, root, err)
 		}
 		proxy.StopCaching = false
 	}
@@ -190,10 +190,10 @@ func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, ds
 // The root node (e.g. a manifest of the artifact) is identified by a descriptor.
 func CopyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Storage, root ocispec.Descriptor, opts CopyGraphOptions) error {
 	if src == nil {
-		return newCopyError("CopyGraph", CopyErrorOriginSource, errors.New("nil source target"))
+		return newCopyError("CopyGraph", CopyErrorOriginSource, ocispec.Descriptor{}, errors.New("nil source target"))
 	}
 	if dst == nil {
-		return newCopyError("CopyGraph", CopyErrorOriginDestination, errors.New("nil destination target"))
+		return newCopyError("CopyGraph", CopyErrorOriginDestination, ocispec.Descriptor{}, errors.New("nil destination target"))
 	}
 	return copyGraph(ctx, src, dst, root, nil, nil, nil, opts)
 }
@@ -243,7 +243,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 		// skip if a rooted sub-DAG exists
 		exists, err := dst.Exists(ctx, desc)
 		if err != nil {
-			return newCopyError("Exists", CopyErrorOriginDestination, err)
+			return newCopyError("Exists", CopyErrorOriginDestination, desc, err)
 		}
 		if exists {
 			if opts.OnCopySkipped != nil {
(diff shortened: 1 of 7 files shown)


copy.go
```diff
@@ -169,7 +169,7 @@ func Copy(ctx context.Context, src ReadOnlyTarget, srcRef string, dst Target, ds
 		proxy.StopCaching = true
 		root, err = opts.MapRoot(ctx, proxy, root)
```
TerryHowe: This line overwrites `root`. Make a copy maybe before this and use the copy below? Maybe a test for it too.

TerryHowe (changes requested):

copy.go
prakrititz: Thank you for pointing that out. You were right the root, err = opts.MapRoot(...) was overwriting root before the error check ran, so when MapRoot failed and returned a zero-value descriptor (which is what the existing test scaffolding here already does for MapRoot failures), the CopyError on the next line ended up carrying that zero descriptor instead of the actual root that was being mapped. That defeats the point of this PR, since the whole idea is for CopyError to say which descriptor failed.

I have fixed it the way you suggested: MapRoot's result now goes into a separate mappedRoot variable, and root only gets reassigned to it after the error check passes. So on failure, root is still the original pre-map descriptor and that's what gets passed into newCopyError.

I have also added added a test for it: MapRoot returning errdef.ErrNotFound plus a zero-value descriptor (mirroring the existing pattern in this file), then asserting the resulting CopyError.Descriptor is the original root and explicitly not the zero value. That second assertion is the one that fails against the old code, so it should catch this if it ever regresses.

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1352, opened 2026-08-29
test: cover buildMirrorRepositories CACert build error

### LINKED ISSUE #1309: Cover the mirror client build-error path in buildMirrorRepositories
> **Updated 2026-08-27.** This issue has been rescoped. The original body was written against a stale snapshot and overstated the gap — it claimed 14.3% coverage and listed five cases, four of which are already covered (and one of which asserted the opposite of the actual behavior). `buildMirrorRepositories` has also since landed on `main` via #1328, so `main` is now the correct base. The accurate, remaining scope is below.

## What

`buildMirrorRepositories` in [`registry/remote/builder.go`](https://github.com/oras-project/oras-go/blob/main/registry/remote/builder.go#L346) sits at **87.5%** statement coverage (14 of 16). Two blocks are unexercised:

**1. The mirror client build-error path** (`builder.go:369-371`):

```go
mirrorClient, err := builder.Build(mirrorProps)
if err != nil {
    return nil, fmt.Errorf("failed to build mirror client for %s: %w", m.Location, err)
}
```

**2. The warning-logger branch** (`builder.go:378-380`):

```go
if builder.Logger != nil {
    mirrorReg.HandleWarning = NewWarningLogger(m.Location, builder.Logger)
}
```

Everything else in the function — the happy-path loop, `PullFromMirror` defaulting, credential isolation, the empty-mirrors guard — is already covered by `Test_buildMirrorRepositories_doesNotShareCredential` (`registry/remote/mirror_isolation_test.go:198`) and the `Test_Repository_*_usesMirror` / `Test_withMirrorFallback*` families in `registry/remote/mirror_test.go`.

## Why it matters

A mirror whose client fails to build must return a wrapped error naming the offending `m.Location`. Without that, a misconfigured mirror is very hard to diagnose: the user sees a failure with no indication of *which* mirror entry is at fault. The wrapping also has to stay intact so callers can inspect the underlying cause.

## What to do

Add tests to `registry/remote/mirror_test.go`.

**Case A — build failure.** Give a mirror a `Transport` that makes `ClientBuilder.Build` fail. The cheapest lever is a CA certificate path that does not exist, which fails in `buildTLSConfig` (`builder.go:160`) — no network access required. `TestClientBuilder_Build_WithInvalidCACert` (`registry/remote/builder_test.go:438`) already uses this trick for the primary registry; the new test puts the same bad path in `properties.Mirror.Transport` instead.

Assert that:
- the returned slice is `nil`,
- the error is non-nil and its message contains the mirror's `Location`,
- the underlying cause survives the wrap. The chain is `buildMirrorRepositories` -> `Build` -> `buildTLSConfig` -> `os.ReadFile`, three `%w` wraps deep, and it has been verified that `errors.Is(err, fs.ErrNotExist)` holds end to end. Prefer that over a string match on the cause.

Also worth a second row where the **second** of two mirrors is the one that fails, confirming the whole slice is discarded rather than partially returned.

**Case B — warning logger.** Set `builder.Logger` to a real `*slog.Logger` (e.g. `slog.New(slog.NewTextHandler(&buf, nil))`) and assert `mirrors[0].Registry.HandleWarning` is wired up; assert it stays `nil` when `Logger` is nil. Please drive an actual warning through the handler and assert on the logged output rather than only checking non-nil — a test whose sole check is that a value is non-nil does not tell us much.

**Optional nicety.** `m.Transport.PlainHTTP` flows into `mirrorReg.PlainHTTP` (`builder.go:376`) and nothing currently asserts it. Cheap to fold into a table row if you are already in there.

### Do NOT test credential inheritance

The original body claimed mirrors inherit `Credential` from the primary registry. **They do not, by design.** The primary's credential is deliberately not propagated, because a mirror is a different host that may be operated by a different party; mirror credentials resolve per host through the `ClientBuilder`'s credential store. See the comment at `builder.go:352-364` and the existing test at `mirror_isolation_test.go:198`. Please do not weaken that assertion.

## Verifying

```
go test ./registry/remote/ -coverprofile=cover.out
go tool cover -func=cover.out | grep buildMirrorRepositories
```

`buildMirrorRepositories` should go from `87.5%` to `100.0%`.

## Notes for first-time contributors

- Base your branch on `main`. `buildMirrorRepositories` landed there in #1328 on 2026-08-26.
- Table-driven tests are the norm in this package.
- Assert on observable behavior rather than on constructor calls succeeding.
- Commits need a DCO sign-off (`git commit -s`). See [CONTRIBUTING.md](https://github.com/oras-project/oras-go/blob/main/CONTRIBUTING.md).

This is still a genuine good-first-issue — it is now just accurately scoped at roughly two tests. No one has claimed it.

Lishkinfeld:
**What this PR does / why we need it**:

Adds a unit test for the `buildMirrorRepositories` mirror-client Build failure path (#1309).

A mirror `Transport` with a nonexistent `CACert` (`/nonexistent/ca.crt`) makes `ClientBuilder.Build` fail in `buildTLSConfig` via `os.ReadFile`. The test asserts:

- the returned slice is `nil`
- the error names the failing mirror `Location`
- `errors.Is(err, fs.ErrNotExist)` holds through the wrap chain (`buildMirrorRepositories` → `Build` → `buildTLSConfig` → `os.ReadFile`)

Includes a second table row where the second of two mirrors is the one that fails, so a partial slice is not returned.

Does not add warning-logger coverage (already covered by `t.Run("mirror warnings use the mirror location")`). Production code is unchanged.

**Which issue(s) this PR fixes**:
Fixes #1309

**Test plan**:
- [x] `go test ./registry/remote/ -run Test_buildMirrorRepositories_invalidCACert`
- [x] `go test ./registry/remote/`


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1354, opened 2026-08-31
feat(registry): expose predecessor finder

### LINKED ISSUE #1353: Add content.PredecessorFinder to the registry.Repository interface
## Summary

Add `content.PredecessorFinder` to the `registry.Repository` interface so that a
repository obtained from `Registry.Repository()` satisfies
`content.ReadOnlyGraphStorage` and can be passed to `oras.ExtendedCopyGraph`
without a type assertion.

Follow-up to #806, which described the problem but predates v3. Filing a fresh
issue because the work is now concrete and the breaking-change window is open.

## Background

`registry.Registry.Repository()` returns the `registry.Repository` interface:

```go
repo, err := reg.Repository(ctx, "myrepo")   // registry.Repository
```

That interface (`registry/interfaces.go:39`) has no `Predecessors` method, so it
does not satisfy `content.ReadOnlyGraphStorage` (`content/graph.go:43`). Callers
who reach a repository this way can use `oras.CopyGraph` but not
`oras.ExtendedCopyGraph`, and have to fall back to a type assertion:

```go
target, ok := repo.(content.GraphTarget)
if !ok {
    // handle exception
}
```

The concrete type already has the method — `(*remote.Repository).Predecessors`
is at `registry/remote/repository.go:715` — so this is purely an interface gap.

Adding a method to an exported interface is breaking, which is why it was
deferred out of v2. v3 is unreleased, so this is the window.

## What needs to be done

1. **Add the embed** in `registry/interfaces.go`:

   ```go
   type Repository interface {
       content.Storage
       content.Deleter
       content.TagResolver
       content.PredecessorFinder   // <- new
       ReferenceFetcher
       ReferencePusher
       ReferrerLister
       TagLister
       ...
   }
   ```

2. **Implement `Predecessors` on `policyEnforcingRepository`**
   (`registry/remote/middleware.go:73`). This type deliberately holds its inner
   repository in a named field rather than embedding it, so that
   "every method of `registry.Repository` must be implemented explicitly" — it
   will fail to compile until the new method is handled. That is working as
   intended, and it forces the real decision here:

   **Does `Predecessors` get a policy check?** It is a read of the referrers
   graph, so the natural answer is to mirror what `Referrers` already does in
   that file. Whichever way it goes, it should be deliberate and covered by a
   test, not just delegated by reflex.

3. **Sweep for other implementers.** `registry.Repository` is also the parameter
   and return type of `RepositoryMiddleware` (`registry/remote/middleware.go:29`),
   so any other wrapper in the tree needs the same treatment. A
   `var _ registry.Repository = (*T)(nil)` assertion exists for the policy
   wrapper at `middleware.go:314`; add equivalents for anything new.

4. **Tests.** Cover that a repository obtained via `Registry.Repository()` can be
   passed straight to `oras.ExtendedCopyGraph` with no type assertion — that is
   the user-visible behavior this is for.

5. **Migration guide.** Note the interface change for anyone implementing
   `registry.Repository` themselves: they must add a `Predecessors` method.
   Consumers who only *call* the interface are unaffected.

## Notes

- No change is needed to `*remote.Repository`; it already has the method.
- If the decision is instead to leave the interface alone, that is a reasonable
  outcome too — but it should be recorded, because after v3 GA it cannot be
  revisited.

BLCCoreStudio:
## Summary

- embed `content.PredecessorFinder` in `registry.Repository`
- make the policy-enforcing repository explicitly delegate `Predecessors`, mirroring the existing referrers carve-out
- add coverage that a repository returned by `Registry.Repository()` can be passed directly to `oras.ExtendedCopyGraph` without a type assertion
- document the `registry.Repository` interface change in the migration guide

## Testing

- `go test ./...`
- `make test`
- `git diff --check`

Fixes #1353

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -1,5 +1,9 @@
 # Migration Guide
 
+## Major Changes in `v3`
+
+- `registry.Repository` now embeds `content.PredecessorFinder`. Implementations of `registry.Repository` must provide a `Predecessors` method. Consumers who only call the interface are unaffected.
+
 In version `v2`, ORAS Go library has been completely refreshed with:
 
 - More unified interfaces
--- a/extendedcopy_registry_test.go
+++ b/extendedcopy_registry_test.go
@@ -1,0 +1,50 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package oras_test
+
+import (
+	"context"
+	"testing"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	oras "github.com/oras-project/oras-go/v3"
+	"github.com/oras-project/oras-go/v3/registry/remote"
+)
+
+func TestRegistryRepositorySupportsExtendedCopyGraph(t *testing.T) {
+	ctx := context.Background()
+
+	reg, err := remote.NewRegistry("example.com")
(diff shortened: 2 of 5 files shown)


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1356, opened 2026-08-31
fix: validate successor descriptors during copy

### LINKED ISSUE #958: Validating successor descriptor during copy
## Problem
When copying a manifest, it's possible for the manifest to reference a malformed descriptor. For example, the following manifest includes a subject descriptor that is missing the required `size` field:

```json
{
  "artifactType": "application/vnd.dev.sigstore.bundle.v0.3+json",
  "config": {
    "digest": "sha256:b198fd23d763339899149b9d9c96bf43d4ba9fa97e6c81d502c9fbbb58eb319f",
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 233
  },
  "layers": [
    {
      "digest": "sha256:eb14f1aeeecfb76481f4ff70b2d523fe065d947932768317bb7469581cd0a0a5",
      "mediaType": "application/vnd.dev.cosign.simplesigning.v1+json",
      "size": 432
    }
  ],
  "schemaVersion": 2,
"subject": {
    "digest": "sha256:60ac88fe91aee1a990c7f89f8003e391ac80c847321ccf697986320a6e9717c8",
    "mediaType": "application/vnd.oci.image.manifest.v1+json"
  }
}
````

The relationship between the manifest and its `subject` can be visualized as:

```mermaid
graph TD;

Manifest--subject-->Subject["Subject (sha256:60ac88...)"]
```

During the copy process, the [`Successors`](https://github.com/oras-project/oras-go/blob/v2.6.0/content/graph.go#L50) method is [invoked](https://github.com/oras-project/oras-go/blob/v2.6.0/copy.go#L243) on the manifest. This returns the subject descriptor, which is then passed to a goroutine responsible for copying the subject:

```json
{
    "digest": "sha256:60ac88fe91aee1a990c7f89f8003e391ac80c847321ccf697986320a6e9717c8",
    "mediaType": "application/vnd.oci.image.manifest.v1+json"
}
```

When processing the subject, since this descriptor lacks a `size` field, it may fail subsequent validation, especially when compared against the `Content-Length` header from a remote registry. This results in an error like:

> Error: GET "https://{registry}/v2/{repo}/manifests/sha256:60ac88fe91aee1a990c7f89f8003e391ac80c847321ccf697986320a6e9717c8": mismatch Content-Length

This error is not intuitive and makes it difficult for users to identify the root cause.

## Proposal

To address this, we could consider validating the successor descriptors earlier in the copy process and returning a clear error message when issues are detected.

One possible error message could be:

> invalid successor descriptor for {manifest_digest}: successor media type: application/vnd.oci.image.manifest.v1+json; successor size: 0; successor digest: sha256:60ac88fe91aee1a990c7f89f8003e391ac80c847321ccf697986320a6e9717c8

pentaoa:
## What this PR does

Validate each non-foreign successor descriptor immediately after `FindSuccessors` and before any successor copy is scheduled. The validation checks digest validity, negative sizes, and zero-sized manifest descriptors. Errors identify both the parent digest and all core fields of the invalid successor.

Zero-byte blobs remain valid and are covered by a regression test; only zero-sized descriptors for manifest media types are rejected because a manifest must contain a document. Media type itself is not rejected because custom graph implementations can return traversable descriptors without one.

This turns late, registry-specific failures such as `mismatch Content-Length` into an actionable error at the graph edge that introduced the malformed descriptor.

## Testing

- `go test .`
- focused invalid digest, negative size, and missing manifest size cases
- focused valid zero-byte blob traversal case

Closes #958

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -263,8 +263,9 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 		successors = removeForeignLayers(successors)
 		for _, successor := range successors {
 			if err := validateSuccessor(successor); err != nil {
-				return fmt.Errorf("invalid successor descriptor for %s: successor media type: %s; successor size: %d; successor digest: %s: %w",
-					desc.Digest, successor.MediaType, successor.Size, successor.Digest, err)
+				return newCopyError("FindSuccessors", CopyErrorOriginSource, desc,
+					fmt.Errorf("invalid successor descriptor: successor media type: %s; successor size: %d; successor digest: %s: %w",
+						successor.MediaType, successor.Size, successor.Digest, err))
 			}
 		}
 
@@ -334,7 +335,7 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 // the next node in a content graph.
 func validateSuccessor(desc ocispec.Descriptor) error {
 	if err := desc.Digest.Validate(); err != nil {
-		return fmt.Errorf("invalid digest: %w", err)
+		return fmt.Errorf("invalid digest: %v: %w", err, errdef.ErrInvalidDigest)
 	}
 	if desc.Size < 0 {
 		return fmt.Errorf("invalid size %d", desc.Size)
--- a/copy.go
+++ b/copy.go
@@ -261,6 +261,12 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 			return newCopyError("FindSuccessors", CopyErrorOriginSource, desc, err)
 		}
 		successors = removeForeignLayers(successors)
+		for _, successor := range successors {
+			if err := validateSuccessor(successor); err != nil {
+				return fmt.Errorf("invalid successor descriptor for %s: successor media type: %s; successor size: %d; successor digest: %s: %w",
+					desc.Digest, successor.MediaType, successor.Size, successor.Digest, err)
+			}
+		}
 
 		if len(successors) != 0 {
 			// for non-leaf nodes, process successors and wait for them to complete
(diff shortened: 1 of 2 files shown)


copy.go
```diff
@@ -324,6 +330,27 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 	return syncutil.Go(ctx, limiter, fn, root)
 }
 
+// validateSuccessor checks the descriptor fields needed to safely traverse to
+// the next node in a content graph.
+func validateSuccessor(desc ocispec.Descriptor) error {
+	if err := validateMediaType(desc.MediaType); err != nil {
+		return fmt.Errorf("invalid media type: %w", err)
+	}
```
TerryHowe: I think the media type validation may be too strict.

TerryHowe (dismissed):

copy.go
pentaoa: Agreed. `CopyGraph` does not require a successor media type to traverse custom graphs, so validating it here was stricter than the copy operation itself. I removed that check and its test, then rebased the PR onto the current `main`. The focused package tests pass.

copy.go
```diff
@@ -261,6 +261,12 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 			return newCopyError("FindSuccessors", CopyErrorOriginSource, desc, err)
 		}
 		successors = removeForeignLayers(successors)
+		for _, successor := range successors {
+			if err := validateSuccessor(successor); err != nil {
+				return fmt.Errorf("invalid successor descriptor for %s: successor media type: %s; successor size: %d; successor digest: %s: %w",
+					desc.Digest, successor.MediaType, successor.Size, successor.Digest, err)
```
TerryHowe:
```suggestion
				return newCopyError("FindSuccessors", CopyErrorOriginSource, desc,
					fmt.Errorf("invalid successor descriptor: successor media type: %s; successor size: %d; successor digest: %s: %w",
						successor.MediaType, successor.Size, successor.Digest, err))
```

copy.go
```diff
@@ -324,6 +330,24 @@ func copyGraph(ctx context.Context, src content.ReadOnlyStorage, dst content.Sto
 	return syncutil.Go(ctx, limiter, fn, root)
 }
 
+// validateSuccessor checks the descriptor fields needed to safely traverse to
+// the next node in a content graph.
+func validateSuccessor(desc ocispec.Descriptor) error {
+	if err := desc.Digest.Validate(); err != nil {
+		return fmt.Errorf("invalid digest: %w", err)
```
TerryHowe:
```suggestion
		return fmt.Errorf("invalid digest: %v: %w", err, errdef.ErrInvalidDigest)
```

copy_test.go
```diff
@@ -1892,6 +1893,105 @@ func TestCopyGraph_WithOptions(t *testing.T) {
...
+					return []ocispec.Descriptor{tt.successor}, nil
+				},
+			}
+
+			err := oras.CopyGraph(ctx, src, cas.NewMemory(), root, opts)
+			if err == nil {
+				t.Fatal("CopyGraph() error = nil, want invalid successor descriptor error")
+			}
+			want := fmt.Sprintf("invalid successor descriptor for %s: successor media type: %s; successor size: %d; successor digest: %s",
+				root.Digest, tt.successor.MediaType, tt.successor.Size, tt.successor.Digest)
```
TerryHowe:
```suggestion
			want := fmt.Sprintf("failed to perform \"FindSuccessors\" on source for %s: invalid successor descriptor: successor media type: %s; successor size: %d; successor digest: %s",
				root.Digest, tt.successor.MediaType, tt.successor.Size, tt.successor.Digest)
```

TerryHowe (approved):
/lgtm

PR #1368, opened 2026-09-02
fix(registry/remote): reject tag or digest in NewRepository reference

### LINKED ISSUE #1365: remote.NewRepository silently discards the tag or digest from its reference argument
## Summary

`remote.NewRepository` accepts a reference with a tag or a digest, parses it,
and then silently throws that part away. There is no way to read it back off
the returned `*Repository`. In `v2` the same call preserved it.

The break surfaces in two steps. Unmodified `v2` code fails loudly, because
`Reference` went from a field to a method:

```
vet: repo.Reference.Reference undefined
     (type func() registry.Reference has no field or method Reference)
```

The migration guide's mapping table then prescribes `repo.Reference` →
`repo.Reference()`. Applying that one-character fix compiles, and the value
silently goes empty:

```go
repo, _ := remote.NewRepository("localhost:5000/hello:v1")
ref := repo.Reference()
// String()="localhost:5000/hello" Reference="" Tag="" Digest=""
```

So the compile error is not a safety net — it routes the developer to the edit
that hides the semantic change.

## Detail

`v2` stored the full parsed reference on the repository:

```go
// v2 — registry/remote/repository.go
func NewRepository(reference string) (*Repository, error) {
	ref, err := registry.ParseReference(reference)
	if err != nil {
		return nil, err
	}
	return &Repository{
		Reference: ref,
	}, nil
}
```

so after `remote.NewRepository("localhost:5000/hello:v1")`, both
`repo.Reference.Reference` and `repo.Reference.ReferenceOrDefault()` returned
`"v1"`.

`v3` parses the same input but keeps only two of the four fields:

```go
// v3 — registry/remote/repository.go:180
func NewRepository(reference string) (*Repository, error) {
	ref, err := registry.ParseReference(reference)
	if err != nil {
		return nil, err
	}

	// Create a default Registry
	reg := &Registry{
		Reference: registry.Reference{
			Registry: ref.Registry,
		},
	}

	return &Repository{
		Registry:       reg,
		RepositoryName: ref.Repository,
	}, nil
}
```

`ref.Tag` and `ref.Digest` are dropped on the floor. The accessor cannot
recover them either:

```go
// v3 — registry/remote/repository.go:200
func (r *Repository) Reference() registry.Reference {
	ref := registry.Reference{Repository: r.RepositoryName}
	if r.Registry != nil {
		ref.Registry = r.Registry.Reference.Registry
	}
	return ref
}
```

So `remote.NewRepository("localhost:5000/hello:v1").Reference()` yields
`localhost:5000/hello` with an empty tag and digest, and the `:v1` the caller
passed in is unrecoverable.

## Why it matters for migration

The migration guide's mapping table lists

| v2 | v3 |
| --- | --- |
| `repo.Reference` | `repo.Reference()` |

as a straight rename. It is not: it is a lossy one. A caller who reads the tag
back off the repository — a reasonable thing to do when the reference came in
as a single string from a CLI flag or a config file — gets `""` after the
migration, with no compile error and no runtime error. That is the worst shape
for a breaking change to take.

I could not find any in-repo caller relying on this, so the exposure is to
downstream consumers rather than to this repository's own examples.

## Possible resolutions

Any of these would close the hole; they are not mutually exclusive:

1. **Reject the input.** Have `NewRepository` return an error when given a
   reference carrying a tag or digest, since a `Repository` models a
   repository rather than a specific artifact. This turns a silent data loss
   into a loud one at the call site.
2. **Document it.** State on `NewRepository` that any tag or digest in the
   argument is ignored, and say the same in `MIGRATION_GUIDE.md` under
   **Observable behavior changes** rather than presenting
   `repo.Reference` → `repo.Reference()` as an equivalence.
3. **Preserve it.** Keep the parsed tag/digest somewhere readable, so the
   accessor can round-trip what was passed in.

Option 1 or 2 seems right — dropping the reference looks like a deliberate
consequence of #1297 splitting registry-level from repository-level state, so
preserving it would cut against that design. Either way the migration guide
needs the warning.

### Naming

Option 1 also settles a naming question. `Reference()` returning Form D is
accurate *if* a `Repository` can only ever be Form D — which rejecting tagged
input would guarantee. If instead the narrower contract stays, the name is
worth revisiting: it is byte-identical to `v2`'s field while the semantics
narrowed, and `remote.Registry.Reference` is a field of the same name and type
on an adjacent struct.

`BaseReference()` would be the candidate, matching the vocabulary already in
`Repository.ParseReference`'s doc comment ("resolves ... from a base reference
`r.Reference()`"). A rename on its own is not sufficient, though — the
constructor would still quietly discard the tag.

Found while reviewing #1357.

7487:
Fixes #1365

`remote.NewRepository` parsed its reference argument and silently dropped any tag or digest it carried, so

```go
repo, _ := remote.NewRepository("localhost:5000/hello:v1")
repo.Reference() // localhost:5000/hello — the ":v1" is unrecoverable
```

succeeded with the `:v1` gone and no way to read it back. This implements option 1 from the issue: `NewRepository` now returns an error wrapping `errdef.ErrInvalidReference` when the reference contains a tag or digest, turning the silent data loss into a loud one at the call site. A `Repository` models a repository rather than a specific artifact, so rejecting the input also keeps `Reference()` accurate for every value the constructor accepts.

Changes:

- `registry/remote/repository.go`: reject tag/digest-bearing references in `NewRepository`, and state the contract in its doc comment.
- `registry/remote/repository_test.go`: table cases for a tagged and a digested reference.
- `MIGRATION_GUIDE.md`: note that `repo.Reference` → `repo.Reference()` is not a pure rename, and record the new constructor error under **Observable behavior changes**.

No in-repo caller passes a tagged or digested reference to `NewRepository`, so nothing else changes. `NewRepositoryWithProperties` is untouched: its input is a structured `properties.Registry` built field by field, not a parsed string.

`go build ./...`, `go vet ./registry/...`, and `go test ./...` pass.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -72,6 +72,15 @@ repo.Registry.PlainHTTP = true
 ref := repo.Reference()
 ```
 
+`repo.Reference` → `repo.Reference()` is not a pure rename. In v2 the field
+held the full parsed reference, including any tag or digest passed to
+`remote.NewRepository`; in v3 a `remote.Repository` identifies a repository
+only, so `Reference()` returns just the registry and repository name, and
+`remote.NewRepository` rejects a reference carrying a tag or digest with an
+error wrapping `errdef.ErrInvalidReference`. Parse such input with
+`properties.NewReference` and keep the tag or digest for the operation that
+needs it.
+
 `remote.NewRegistryWithProperties` and
 `remote.NewRepositoryWithProperties` are the preferred constructors when
 configuration comes from `registries.conf`, Docker configuration, certificates,
@@ -238,6 +247,10 @@ changes:
 
 ## Observable behavior changes
 
+- `remote.NewRepository` returns an error wrapping `errdef.ErrInvalidReference`
+  when the reference contains a tag or digest (e.g.
+  `localhost:5000/example:v1`). In v2 such a reference was accepted and the
+  tag or digest was kept on `repo.Reference`.
 - `oras.CopyError` includes a `Descriptor` when a copy operation had already
   selected content. Its error string includes the digest in those cases. Use
   `errors.As` and the structured fields rather than comparing error strings.
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -177,10 +177,18 @@ type Repository struct {
 // NewRepository creates a client to the remote repository identified by a
 // reference. A default Registry is created internally.
 // Example: localhost:5000/hello-world
+//
+// The reference must not contain a tag or a digest, since a Repository
+// identifies a repository rather than a specific artifact. A reference
+// carrying one is rejected with an error wrapping
+// [errdef.ErrInvalidReference].
 func NewRepository(reference string) (*Repository, error) {
 	ref, err := registry.ParseReference(reference)
 	if err != nil {
 		return nil, err
+	}
+	if ref.Tag != "" || ref.Digest != "" {
+		return nil, fmt.Errorf("%w: %q: expected a repository without a tag or digest", errdef.ErrInvalidReference, reference)
 	}
 
 	// Create a default Registry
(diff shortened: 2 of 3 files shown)


TerryHowe (approved):
/lgtm

PR #1369, opened 2026-09-02
fix(registry/remote): make signedBy policy satisfiable for tag references and tagless operations

### LINKED ISSUE #1337: signedBy policy can never be satisfied for tag references or tagless operations
## Summary

A `signedBy` policy requirement can never be satisfied unless the caller uses an explicit `@sha256:` reference. Tag references are rejected, and so is every operation that does not carry a reference at all.

It fails closed, so this is a functional gap rather than a security hole — but the feature is unusable for the most common way people name images.

## Impact

Two distinct paths, both broken:

1. **Tag references.** `Resolve`, `FetchReference`, `Tag` and `PushReference` pass the caller's reference straight through, so a tag stays a tag.
2. **Everything else.** All tagless operations call `checkPolicy(ctx, "")`, which falls back to `repoRef.String()` — bare `registry/repo`, no digest at all. That covers `Blobs().Fetch`, `Push`, `Mount`, `Delete`, `Exists`, `Tags`, `Referrers` and `Predecessors`.

So under a `signedBy` policy, a digest-pinned `Resolve` works and essentially nothing else does.

## Root cause

Policy is evaluated *before* the reference is resolved. `checkPolicy` builds the `ImageReference` from whatever the caller passed (`registry/remote/repository.go:426-436`):

```go
repoRef := r.Reference()
ref := repoRef.String()
if reference != "" {
	ref = reference
}

imageRef := policy.ImageReference{
	Transport: policy.TransportNameDocker,
	Scope:     repoRef.Registry + "/" + repoRef.Repository,
	Reference: ref,
}
```

That reaches `DefaultSignedByVerifier.Verify` unresolved, which needs a manifest digest because that is what a simple-signing payload binds to (`critical.image.docker-manifest-digest`). It gets one via `parseImageDigest` (`registry/remote/signature/verifier.go:154`), which scans for `@` and errors out if there is none:

```go
func parseImageDigest(ref string) (digest.Digest, error) {
	for i := len(ref) - 1; i >= 0; i-- {
		if ref[i] == '@' { /* ... */ }
	}
	return "", fmt.Errorf("reference %s does not contain a digest", ref)
}
```

Chain: `checkPolicy` → `IsImageAllowed` → `evaluateSignedBy` → `Verify` → `parseImageDigest` fails → `Verify` returns an error → the pull is denied.

## Reproduction

Against `725b6d8`, with a real keyring so `loadKeyRing` succeeds and the verifier reaches the digest parse:

```go
v := signature.NewSignedByVerifier(signature.NewLookasideStore("file://"+t.TempDir(), ""))
req := &policy.PRSignedBy{KeyType: "GPGKeys", KeyData: string(armoredPublicKey)}

for _, ref := range []string{
	"registry.example.com/app:v1", // Resolve / FetchReference with a tag
	"registry.example.com/app",    // every checkPolicy(ctx, "") operation
	"registry.example.com/app@sha256:e3b0c442...52b855",
} {
	ok, err := v.Verify(context.Background(), req, policy.ImageReference{
		Transport: policy.TransportNameDocker,
		Scope:     "registry.example.com/app",
		Reference: ref,
	})
	t.Logf("%q allowed=%v err=%v", ref, ok, err)
}
```

```
"registry.example.com/app:v1"      allowed=false err=failed to parse image digest from reference registry.example.com/app:v1: reference registry.example.com/app:v1 does not contain a digest
"registry.example.com/app"         allowed=false err=failed to parse image digest from reference registry.example.com/app: reference registry.example.com/app does not contain a digest
"registry.example.com/app@sha256:" allowed=false err=<nil>
```

The digest case returns `err=nil` — it is `false` only because the lookaside store is empty, which is correct. That isolates the failure in the other two to the digest parse specifically.

## Why it is not a one-line fix

The obvious move — resolve the tag first, then evaluate policy — inverts the current ordering, and that ordering exists for a reason: policy should be able to reject a request *before* the client talks to the registry about it. Scope- and transport-level rules (`reject`, `insecureAcceptAnything`) are meaningful pre-flight; signature requirements inherently are not.

There is also a re-entrancy hazard. Resolving inside the verifier means calling back into `Repository.Resolve`, which calls `checkPolicy` again. The `policyCheckedKey` context marker prevents infinite recursion, but only if it is threaded correctly through the verifier.

## Options

**1. Two-phase evaluation** (suggested target). Split requirements by what they need:

- *Pre-resolve*: `reject`, `insecureAcceptAnything`, and any future reference-shaped rule — evaluated in `checkPolicy` as today.
- *Post-resolve*: `signedBy`, `sigstoreSigned` — evaluated once the descriptor is known.

```go
func (r *Repository) Resolve(ctx context.Context, reference string) (ocispec.Descriptor, error) {
	if err := r.checkPolicyPreResolve(ctx, reference); err != nil {
	}
	desc, err := r.Manifests().Resolve(withPolicyChecked(ctx), reference)
	if err != nil {
	}
	if err := r.checkPolicyPostResolve(ctx, reference, desc); err != nil {
	}
	return desc, nil
}
```

This is roughly how `containers/image` splits it: `IsRunningImageAllowed` operates on an `UnparsedImage` already fetched far enough to know its manifest digest, while reference-level rules are applied earlier against the parsed reference. Signature requirements only ever see a resolved image.

Note this still leaves the tagless operations to decide on — a blob `Fetch` has no manifest to verify, so `signedBy` may simply not be applicable there rather than denying.

**2. Resolve inside the verifier.** Smaller diff, but puts a network round-trip behind a policy evaluation and has the re-entrancy hazard above.

**3. Document the limitation** and reject `signedBy` at policy-construction time with a clear error, instead of failing per-operation with a confusing digest-parse message. Weakest option, but better than the current behavior if the real fix misses v3.

## Notes

The `checkPolicy` doc currently claims signature verification is "performed regardless of operation type" (`repository.go:405-415`), which reads as intentional coverage. In practice it is performed and always fails for those operations — worth correcting alongside whichever option is taken.

Needs a decision before v3 GA.

7487:
Fixes #1337

### Problem

A `signedBy` requirement could only be satisfied by an explicit `@digest` reference. `checkPolicy` evaluated the full policy before resolution, so:

- tag references (`Resolve`, `FetchReference`, `Tag`, `PushReference`) reached `DefaultSignedByVerifier` without a digest and failed at `parseImageDigest`;
- every tagless operation called `checkPolicy(ctx, "")`, fell back to the bare repository reference, and failed the same way.

### Approach (option 1 from the issue: two-phase evaluation)

**Pre-resolve** — `checkPolicy` now evaluates only requirements decidable from the reference alone (`reject`, `insecureAcceptAnything`) via the new `Evaluator.IsReferenceAllowed`, so it can still deny before any registry round-trip.

**Post-resolve** — the new `checkPolicyResolved` evaluates the full policy, signature requirements included, once the manifest descriptor is known:

- `Resolve` / `FetchReference` check the resolved descriptor (at both `Repository` and `Manifests()` levels, guarded by `policyCheckedKey` so the wrapped path evaluates once);
- `Tag` / `PushReference` check the supplied descriptor before any network access;
- on a post-resolve denial, `FetchReference` closes the fetched body and returns the policy error.

The resolved digest travels in a new additive `policy.ImageReference.Digest` field, so the caller's original reference stays intact for signed identity matching (`matchExact` etc. keep working for tag pulls, mirroring how containers/image feeds identity and digest separately). `DefaultSignedByVerifier` prefers that field and falls back to parsing the reference, validating the supplied digest before using it for signature lookup.

Operations that never involve a manifest reference (blob operations, `Tags`, `Referrers`, `Predecessors`, `Delete`, `Mount`, `Exists`, `Untag`) are now subject only to reference-level requirements instead of failing unconditionally — per the issue, `signedBy` is not applicable there rather than denying. The stale `checkPolicy` doc claim ("performed regardless of operation type") is corrected accordingly.

Behavior summary under a `signedBy` policy:

| Operation | Before | After |
|---|---|---|
| `Resolve`/`FetchReference` with `@digest` | verified | verified (unchanged) |
| `Resolve`/`FetchReference` with tag | always denied | verified against resolved digest |
| `Tag`/`PushReference` | always denied | verified against descriptor digest |
| tagless/blob/metadata ops | always denied | reference-level requirements only |

Not covered here: the opt-in `WithPolicyEnforcement` middleware has the same limitation (it passes raw references/digest strings to `IsImageAllowed`). It can now be fixed the same way using `IsReferenceAllowed` + `ImageReference.Digest`; happy to do that as a follow-up if wanted.

Tested with `go test -race ./...`; new coverage for the evaluator phase split, the verifier digest-field path, and repository/manifest-store two-phase enforcement (allow + deny, verifier call recording).

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -410,25 +410,17 @@ func withPolicyChecked(ctx context.Context) context.Context {
 	return context.WithValue(ctx, policyCheckedKey{}, true)
 }
 
-// checkPolicy validates the repository access against the requirements of
-// the configured policy that can be decided from the reference alone, such
-// as reject and insecureAcceptAnything. If no policy is configured (Policy
-// is nil), this is a no-op.
+// checkPolicy validates repository access before network I/O. It evaluates
+// reference-level requirements for every reference and digest-dependent
+// requirements when the reference already carries a digest. If no policy is
+// configured (Policy is nil), this is a no-op.
 //
-// Policy is enforced on all registry operations including:
-//   - Content operations: Fetch, Push, Resolve, FetchReference, PushReference
-//   - Metadata operations: Exists, Tags, Referrers, Predecessors
-//   - Mutating operations: Delete, Tag, Mount
-//
-// Signature requirements (signedBy/sigstoreSigned) verify signatures
-// against a manifest digest, which a tag reference or a reference-less
-// operation does not carry, so they are evaluated separately by
-// checkPolicyResolved on the operations that resolve or supply a manifest
-// descriptor: Fetch, Push, Exists, Delete, Resolve, FetchReference, Tag and
-// PushReference. Operations that never involve a manifest are subject only
-// to reference-level requirements. Descriptor-based push operations verify
-// the supplied manifest before uploading it, so its signing material must be
-// discoverable by the configured verifier at that point.
+// Signature requirements (signedBy/sigstoreSigned) are otherwise evaluated
+// by checkPolicyResolved when an operation resolves or supplies a manifest
+// descriptor. Blob operations and operations without a manifest remain
+// reference-level only. Descriptor-based push operations verify the supplied
+// manifest before uploading it, so its signing material must be discoverable
+// by the configured verifier at that point.
 func (r *Repository) checkPolicy(ctx context.Context, reference string) error {
 	if ctx.Value(policyCheckedKey{}) != nil {
 		return nil
@@ -447,7 +439,8 @@ func (r *Repository) checkPolicy(ctx context.Context, reference string) error {
 	if !allowed {
 		return fmt.Errorf("access denied by policy for %s", imageRef.Reference)
 	}
-	if imageRef.Digest = r.policyDigest(reference); imageRef.Digest != "" {
+	imageRef.Digest = r.policyDigest(reference)
+	if imageRef.Digest != "" {
 		allowed, err = pol.IsResolvedImageAllowed(withPolicyChecked(ctx), imageRef)
 		if err != nil {
 			return fmt.Errorf("policy check failed: %w", err)
@@ -462,7 +455,8 @@ func (r *Repository) checkPolicy(ctx context.Context, reference string) error {
 // checkPolicyResolved validates the manifest descriptor a reference resolves
 // to against digest-dependent policy requirements. The caller's reference is
 // kept for signed identity matching while the resolved digest is used for
-// signature lookup and payload validation.
+// signature lookup and payload validation. Callers must first call checkPolicy
+// with the same reference because digest references are fully checked there.
 func (r *Repository) checkPolicyResolved(ctx context.Context, reference string, desc ocispec.Descriptor) error {
 	if ctx.Value(policyCheckedKey{}) != nil {
 		return nil
@@ -478,7 +472,11 @@ func (r *Repository) checkPolicyResolved(ctx context.Context, reference string, 
 		return nil
 	}
 
-	imageRef := r.policyImageReference(reference)
+	policyReference := reference
+	if policyReference == "" {
+		policyReference = desc.Digest.String()
+	}
+	imageRef := r.policyImageReference(policyReference)
 	imageRef.Digest = desc.Digest
 
 	// Mark the context as checked so verifiers that call back into this
@@ -516,7 +514,11 @@ func (r *Repository) policyImageReference(reference string) policy.ImageReferenc
 	repoRef := r.Reference()
 	ref := repoRef.String()
 	if reference != "" {
-		ref = reference
+		if parsed, err := r.ParseReference(reference); err == nil {
+			ref = parsed.String()
+		} else {
+			ref = reference
+		}
 	}
 	return policy.ImageReference{
 		Transport: policy.TransportNameDocker,
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -424,9 +424,11 @@ func withPolicyChecked(ctx context.Context) context.Context {
 // against a manifest digest, which a tag reference or a reference-less
 // operation does not carry, so they are evaluated separately by
 // checkPolicyResolved on the operations that resolve or supply a manifest
-// descriptor: Resolve, FetchReference, Tag and PushReference. Operations
-// that never involve a manifest reference are subject only to
-// reference-level requirements.
+// descriptor: Fetch, Push, Exists, Delete, Resolve, FetchReference, Tag and
+// PushReference. Operations that never involve a manifest are subject only
+// to reference-level requirements. Descriptor-based push operations verify
+// the supplied manifest before uploading it, so its signing material must be
+// discoverable by the configured verifier at that point.
 func (r *Repository) checkPolicy(ctx context.Context, reference string) error {
 	if ctx.Value(policyCheckedKey{}) != nil {
 		return nil
@@ -445,15 +447,29 @@ func (r *Repository) checkPolicy(ctx context.Context, reference string) error {
 	if !allowed {
 		return fmt.Errorf("access denied by policy for %s", imageRef.Reference)
 	}
+	if imageRef.Digest = r.policyDigest(reference); imageRef.Digest != "" {
+		allowed, err = pol.IsResolvedImageAllowed(withPolicyChecked(ctx), imageRef)
+		if err != nil {
+			return fmt.Errorf("policy check failed: %w", err)
+		}
+		if !allowed {
+			return fmt.Errorf("access denied by policy for %s", imageRef.Reference)
+		}
+	}
 	return nil
 }
 
 // checkPolicyResolved validates the manifest descriptor a reference resolves
-// to against the full policy, including signature requirements. The caller's
-// reference is kept for signed identity matching while the resolved digest
-// is used for signature lookup and payload validation.
+// to against digest-dependent policy requirements. The caller's reference is
+// kept for signed identity matching while the resolved digest is used for
+// signature lookup and payload validation.
 func (r *Repository) checkPolicyResolved(ctx context.Context, reference string, desc ocispec.Descriptor) error {
 	if ctx.Value(policyCheckedKey{}) != nil {
+		return nil
+	}
+	// Digest references are fully evaluated by checkPolicy before any
+	// registry request is made.
+	if r.policyDigest(reference) != "" {
 		return nil
 	}
 
@@ -468,7 +484,7 @@ func (r *Repository) checkPolicyResolved(ctx context.Context, reference string, 
 	// Mark the context as checked so verifiers that call back into this
 	// repository (e.g. to fetch signature artifacts) do not recursively
 	// evaluate policy.
-	allowed, err := pol.IsImageAllowed(withPolicyChecked(ctx), imageRef)
+	allowed, err := pol.IsResolvedImageAllowed(withPolicyChecked(ctx), imageRef)
 	if err != nil {
 		return fmt.Errorf("policy check failed: %w", err)
 	}
@@ -476,6 +492,22 @@ func (r *Repository) checkPolicyResolved(ctx context.Context, reference string, 
 		return fmt.Errorf("access denied by policy for %s", imageRef.Reference)
 	}
 	return nil
+}
+
+// checkManifestPolicy evaluates both policy phases for a known manifest.
+func (r *Repository) checkManifestPolicy(ctx context.Context, reference string, desc ocispec.Descriptor) error {
+	if err := r.checkPolicy(ctx, reference); err != nil {
+		return err
+	}
+	return r.checkPolicyResolved(ctx, reference, desc)
+}
+
+// checkDescriptorPolicy evaluates signature requirements only for manifests.
+func (r *Repository) checkDescriptorPolicy(ctx context.Context, desc ocispec.Descriptor) error {
+	if isManifest(r.manifestMediaTypes(), desc) {
+		return r.checkManifestPolicy(ctx, "", desc)
+	}
+	return r.checkPolicy(ctx, "")
 }
 
 // policyImageReference builds the policy image reference for the given
@@ -493,10 +525,26 @@ func (r *Repository) policyImageReference(reference string) policy.ImageReferenc
 	}
 }
 
+// policyDigest returns the digest carried by reference, if any.
+func (r *Repository) policyDigest(reference string) digest.Digest {
+	if reference == "" {
+		return ""
+	}
+	ref, err := r.ParseReference(reference)
+	if err != nil {
+		return ""
+	}
+	d, err := digest.Parse(ref.Reference)
+	if err != nil {
+		return ""
+	}
+	return d
+}
+
 // Fetch fetches the content identified by the descriptor.
 // If mirrors are configured, they are tried in order before the primary.
 func (r *Repository) Fetch(ctx context.Context, target ocispec.Descriptor) (io.ReadCloser, error) {
-	if err := r.checkPolicy(ctx, ""); err != nil {
+	if err := r.checkDescriptorPolicy(ctx, target); err != nil {
 		return nil, err
 	}
 	ctx = withPolicyChecked(ctx)
@@ -511,7 +559,7 @@ func (r *Repository) Fetch(ctx context.Context, target ocispec.Descriptor) (io.R
 
 // Push pushes the content, matching the expected descriptor.
 func (r *Repository) Push(ctx context.Context, expected ocispec.Descriptor, content io.Reader) error {
-	if err := r.checkPolicy(ctx, ""); err != nil {
+	if err := r.checkDescriptorPolicy(ctx, expected); err != nil {
 		return err
 	}
 	return r.blobStore(expected).Push(withPolicyChecked(ctx), expected, content)
@@ -540,7 +588,7 @@ func (r *Repository) Mount(ctx context.Context, desc ocispec.Descriptor, fromRep
 // Exists returns true if the described content exists.
 // If mirrors are configured, they are tried in order before the primary.
 func (r *Repository) Exists(ctx context.Context, target ocispec.Descriptor) (bool, error) {
-	if err := r.checkPolicy(ctx, ""); err != nil {
+	if err := r.checkDescriptorPolicy(ctx, target); err != nil {
 		return false, err
 	}
 	ctx = withPolicyChecked(ctx)
@@ -555,7 +603,7 @@ func (r *Repository) Exists(ctx context.Context, target ocispec.Descriptor) (boo
 
 // Delete removes the content identified by the descriptor.
 func (r *Repository) Delete(ctx context.Context, target ocispec.Descriptor) error {
-	if err := r.checkPolicy(ctx, ""); err != nil {
+	if err := r.checkDescriptorPolicy(ctx, target); err != nil {
 		return err
 	}
 	return r.blobStore(target).Delete(withPolicyChecked(ctx), target)
@@ -601,10 +649,7 @@ func (r *Repository) Resolve(ctx context.Context, reference string) (ocispec.Des
 
 // Tag tags a manifest descriptor with a reference string.
 func (r *Repository) Tag(ctx context.Context, desc ocispec.Descriptor, reference string) error {
-	if err := r.checkPolicy(ctx, reference); err != nil {
-		return err
-	}
-	if err := r.checkPolicyResolved(ctx, reference, desc); err != nil {
+	if err := r.checkManifestPolicy(ctx, reference, desc); err != nil {
 		return err
 	}
 	return r.Manifests().Tag(withPolicyChecked(ctx), desc, reference)
@@ -630,10 +675,7 @@ func (r *Repository) Untag(ctx context.Context, reference string) error {
 
 // PushReference pushes the manifest with a reference tag.
 func (r *Repository) PushReference(ctx context.Context, expected ocispec.Descriptor, content io.Reader, reference string) error {
-	if err := r.checkPolicy(ctx, reference); err != nil {
-		return err
-	}
-	if err := r.checkPolicyResolved(ctx, reference, expected); err != nil {
+	if err := r.checkManifestPolicy(ctx, reference, expected); err != nil {
 		return err
 	}
 	return r.Manifests().PushReference(withPolicyChecked(ctx), expected, content, reference)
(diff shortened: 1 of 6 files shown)


TerryHowe:
[registry/remote/repository.go:1509 and :1458] Manifest Fetch and Push now bypass signedBy entirely, even though the descriptor carries the digest. Call s.repo.checkPolicyResolved(ctx, "", target) alongside the existing checkPolicy.

registry/remote/repository.go
```diff
@@ -410,17 +410,23 @@ func withPolicyChecked(ctx context.Context) context.Context {
...
-// For signedBy/sigstoreSigned requirements, signature verification is
-// performed regardless of operation type. If this is too restrictive for
-// administrative operations, configure a separate policy scope.
+// Signature requirements (signedBy/sigstoreSigned) verify signatures
+// against a manifest digest, which a tag reference or a reference-less
+// operation does not carry, so they are evaluated separately by
+// checkPolicyResolved on the operations that resolve or supply a manifest
+// descriptor: Resolve, FetchReference, Tag and PushReference. Operations
+// that never involve a manifest reference are subject only to
+// reference-level requirements.
```
TerryHowe: **Manifest `Fetch` and `Push` now bypass `signedBy` entirely, even though the descriptor carries the digest.**

This paragraph says operations that "never involve a manifest reference" get reference-level requirements only — but `manifestStore.Fetch` (~L1517) and `manifestStore.Push` (~L1466) receive an `ocispec.Descriptor` whose `Digest` is exactly what `checkPolicyResolved` needs. They still call only `checkPolicy(ctx, "")`.

Under a deny-all `signedBy` policy, `Repository.Fetch` on a manifest descriptor returns the manifest body with the verifier never invoked. Probe test against this branch:

```
BYPASS: Fetch returned manifest content "{\"manifests\":[]}" under a deny-all
signedBy policy; verifier calls=0
```

Before this PR that call was denied — fail-closed, albeit for the wrong reason. `oras.CopyGraph` starting from a caller-supplied descriptor never calls `Resolve`, so nothing on that path verifies signatures at all now.

The PR's own rationale — check where a digest is known — is what justifies adding `checkPolicyResolved` to `Tag`/`PushReference` from a caller-supplied descriptor. The same argument applies verbatim here. Worth noting `policyEnforcingRepository.Fetch` in `middleware.go` passes `target.Digest.String()` and *does* get signature verification, so after this PR the two policy mechanisms in the same package enforce different things for the same operation.

Suggested fix: in `manifestStore.Fetch`/`Push`/`Delete`, call `s.repo.checkPolicyResolved(ctx, "", target)` alongside the existing `checkPolicy`. Blob stores correctly stay reference-level only.

registry/remote/repository.go
```diff
@@ -539,21 +579,34 @@ func (r *Repository) Resolve(ctx context.Context, reference string) (ocispec.Des
...
 			func(ctx context.Context, repo *Repository, ref string) (ocispec.Descriptor, error) {
 				return repo.Manifests().Resolve(ctx, ref)
 			})
+	} else {
+		desc, err = r.Manifests().Resolve(resolveCtx, reference)
+	}
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	if err := r.checkPolicyResolved(ctx, reference, desc); err != nil {
```
TerryHowe, line 646: **Digest references lose their pre-flight signature check.**

Previously `Resolve(ctx, "…@sha256:…")` evaluated the full policy before any network I/O. Now the HEAD request goes out first and the signature check runs on the result.

For a reference that already carries a digest, the deferral isn't needed — the digest is known up front. Short-circuiting that case would avoid both the wasted round trip and contacting the registry at all for content the policy is going to reject. Same applies to `FetchReference` below.

registry/remote/repository.go
```diff
@@ -431,28 +437,62 @@ func (r *Repository) checkPolicy(ctx context.Context, reference string) error {
...
+		return nil
+	}
+
+	imageRef := r.policyImageReference(reference)
+	imageRef.Digest = desc.Digest
+
+	// Mark the context as checked so verifiers that call back into this
+	// repository (e.g. to fetch signature artifacts) do not recursively
+	// evaluate policy.
+	allowed, err := pol.IsImageAllowed(withPolicyChecked(ctx), imageRef)
```
TerryHowe: **Non-signature requirements are evaluated twice per operation.**

With a policy of `[insecureAcceptAnything, signedBy]`, `checkPolicy` evaluates `insecureAcceptAnything` via `IsReferenceAllowed`, then this `IsImageAllowed` evaluates it again along with `signedBy`.

Harmless for the built-in requirement types since both are pure, but it makes the contract "a requirement may be evaluated more than once per operation", which is a surprising thing to impose on third-party requirement implementations.

Consider evaluating only the deferred requirements here, mirroring the `requiresResolvedImage` filter in `IsReferenceAllowed` — e.g. a shared `isAllowed(ctx, image, phase)` on `Evaluator`, which would also collapse the two near-identical loops and keep the "no requirements → reject" invariant in one place.

TerryHowe (changes requested):
Thanks for tackling #1337 — the two-phase split is the right shape for this, and a couple of the details are nicely handled: using `err =` (not `:=`) in `manifestStore.FetchReference` so the existing named-return `defer` closes the body on a policy denial, and wrapping the resolved evaluation in `withPolicyChecked` so a verifier fetching signature artifacts through the same repository can't recurse. Keeping the caller's reference for identity matching while carrying the digest separately also matches how containers/image feeds identity and digest.

I built and ran the branch locally at 8cfcbea: `go build ./...`, `go vet ./registry/...` and `go test ./registry/remote/...` all pass.

One blocking issue and two smaller ones, left inline. The blocker is that manifest `Fetch`/`Push` go from fail-closed to entirely unverified under a `signedBy` policy, even though the descriptor they receive already carries the digest — I confirmed it with a probe test on this branch.

Separately, non-blocking, worth a look before merge:

- **Push-side semantics.** `Tag`/`PushReference` now run signature verification against the caller-supplied descriptor, which means a `signedBy` policy blocks pushing new content (the signature artifact doesn't exist in the target repo yet). Not a regression — it was denied before too — but since this PR is defining the intended semantics, it'd help to state in the `checkPolicy` doc whether push-side signature enforcement is intended. containers/image deliberately keeps push out of the `policy.json` accept path.
- **Coverage.** Codecov flags 83.58% on `repository.go`. The uncovered paths are the interesting ones: the mirror-fallback branches of `Resolve`/`FetchReference` under policy, `manifestStore.FetchReference` with `ContentLength == -1` (the new `s.Resolve(withPolicyChecked(ctx), …)` de-duplication), the `rc.Close()` on post-resolve denial, and a regression test that a `@digest` reference still verifies through the new field-vs-fallback path.
- **Follow-up tracking.** Worth an issue for the `WithPolicyEnforcement` middleware so the two policy paths don't drift further apart — note `middleware.go` passes `target.Digest.String()` as the reference, so it *does* verify signatures on descriptor-based `Fetch` today, which is exactly the asymmetry the first inline comment is about.

7487:
Addressed the blocking issue and both inline follow-ups in commit 6aaf710:

- Manifest descriptor operations now run the resolved signature phase before network I/O in both Repository and Manifests() paths; blobs remain reference-only.
- Digest references are signature-checked preflight and the post-response phase is skipped, so verification still runs exactly once.
- Evaluator now shares one filtered evaluation loop: the resolved phase evaluates only digest-dependent requirements, while the empty-requirements fail-closed invariant remains centralized.
- The policy documentation now states the push-side behavior explicitly.
- Added coverage for descriptor operations, mirror paths, unknown Content-Length de-duplication, response-body closure on denial, and @digest preflight/one-shot verification.

Validation: make test (full race-enabled suite) passed; go vet ./... and go build ./... passed; golangci-lint v2.11.4 reports 0 issues. Local repository.go statement coverage is now 87.89% (791/900).

I kept the WithPolicyEnforcement alignment follow-up out of this focused fix. I found no open tracker beyond closed #1265 and can file a dedicated issue if you would like.

registry/remote/repository.go
```diff
@@ -431,32 +439,112 @@ func (r *Repository) checkPolicy(ctx context.Context, reference string) error {
...
+	return r.checkPolicy(ctx, "")
+}
+
+// policyImageReference builds the policy image reference for the given
+// caller reference, falling back to the repository reference when empty.
+func (r *Repository) policyImageReference(reference string) policy.ImageReference {
 	repoRef := r.Reference()
 	ref := repoRef.String()
 	if reference != "" {
 		ref = reference
```
TerryHowe: **Blocking: the reference passed to the verifier is the raw caller string, so identity matching can never succeed.**

`policyImageReference` puts the caller's reference into `ImageReference.Reference` verbatim. For `repo.Resolve(ctx, "signed")` that is the bare tag `"signed"`; for descriptor-based operations it falls back to `repoRef.String()`, i.e. `127.0.0.1:5000/test` with no tag or digest.

`DefaultSignedByVerifier.Verify` ends with an unconditional `MatchSignedIdentity(req.SignedIdentity, image.Reference, payload.DockerReference())`, where `payload.DockerReference()` is a fully-qualified reference like `127.0.0.1:5000/test:signed`. I recorded what the verifier actually receives on this branch and ran each form through `MatchSignedIdentity` with the default (nil) identity:

```
imageRef="signed"                      -> matched=false   // Resolve(ctx, "signed")
imageRef="127.0.0.1:5000/test"         -> matched=false   // Fetch(ctx, desc)
imageRef="127.0.0.1:5000/test:signed"  -> matched=true
imageRef="127.0.0.1:5000/test@sha256:…"-> matched=true
```

Neither form the code produces today matches, under any matcher: `matchExact` compares the strings directly, `matchRepoDigestOrExact` only falls back to repository comparison when the image reference contains `@`, and `matchRepository` would compare `repositoryOf("signed")` == `"signed"` against `127.0.0.1:5000/test`.

So the tag path now reaches the verifier — the part this PR set out to fix — and is then rejected at the last step. The signature digest lookup succeeds and identity matching fails. The existing tests don't catch it because `recordingSignedByVerifier` returns a canned bool and never inspects `Reference`.

Suggested fix: have `policyImageReference` emit a fully-qualified reference — `registry/repository:tag` when the caller passed a tag, and `registry/repository@sha256:…` for digest references and descriptor-based operations. As the table above shows, the digest form makes `matchRepoDigestOrExact` compare repositories and succeed, which is exactly the containers/image behavior being modeled. Since `checkPolicyResolved` already has the resolved digest, it could also upgrade the reference to the `@digest` form for the resolved phase.

Worth adding one end-to-end test using the real `DefaultSignedByVerifier` with a signed payload, rather than only the recording stub — that's what would have caught this.

Minor, same function: because `Reference` is the raw string, the denial for a digest reference reads `access denied by policy for sha256:27b7…` with no repository. Fully qualifying it fixes the error message too.

TerryHowe (changes requested):
Re-reviewed at 6aaf710. All three inline findings from the last round are genuinely fixed — I verified each with probe tests on the branch rather than by reading:

- **Manifest `Fetch`/`Push` bypass — closed.** `checkDescriptorPolicy` + `isManifest` routing works: under a deny-all `signedBy` policy, `Fetch`, `Push` and `Exists` on a manifest descriptor are each denied with exactly one verifier call, and a blob descriptor still makes zero verifier calls.
- **Digest preflight — fixed, and tighter than I suggested.** `repo.Resolve(ctx, "sha256:…")` under a deny-all policy now denies with **0 HTTP requests** and exactly **1** verifier call; the allow path is also exactly 1 call, so the `checkPolicyResolved` short-circuit is doing its job with no double evaluation.
- **Duplicate evaluation — fixed.** The shared `isImageAllowed(ctx, image, filter)` collapses the two loops and keeps the empty-requirements fail-closed invariant in one place. `IsResolvedImageAllowed` as the symmetric counterpart to `IsReferenceAllowed` is a nicer API than what I proposed.

`go build ./...`, `go vet ./...` and `go test -race ./registry/... ./content/... .` all pass locally, and every coverage gap I listed now has a test (`MirrorPaths`, `DeniedClosesBody`, `UnknownContentLengthChecksPolicyOnce`, `DigestPreflight`, `DigestChecksOnce`).

Unfortunately, chasing the descriptor path surfaced something I should have caught in the first pass, and it's blocking: the reference handed to the verifier can never satisfy identity matching, so `#1337` still isn't fixed end-to-end for the in-tree `DefaultSignedByVerifier`. Details inline.

Smaller items, none blocking:

- **`checkPolicy` doc no longer matches the code.** The lead paragraph still says it validates "requirements that can be decided from the reference alone", but the function now also runs `IsResolvedImageAllowed` for digest references. The operation lists below it are also stale in one spot — `Mount` is grouped with `Delete` under "Mutating operations", but `Delete` is now descriptor-checked while `Mount` stays reference-level.
- **Implicit ordering contract.** `checkPolicyResolved` returning `nil` for digest references assumes `checkPolicy` already ran. `checkManifestPolicy` encapsulates the pair correctly, but `Repository.Resolve`/`FetchReference` and `manifestStore.Resolve`/`FetchReference` still call the two separately — a future call site that reaches only `checkPolicyResolved` with a digest reference silently skips every signature check. Worth a comment on `checkPolicyResolved` saying it must not be called without `checkPolicy`, or making the pairing structural.
- **`Exists` on the copy hot path.** `Exists` now runs full signature verification, and `oras.Copy` calls `dst.Exists(desc)` per node. Under a `signedBy` policy that's a signature lookup per node against a destination where the content and its signature don't exist yet. Consistent with the push-side semantics you documented, so I'll take it as intended — just flagging that it's a per-node cost rather than a one-shot one.
- **Style, `checkPolicy`.** `if imageRef.Digest = r.policyDigest(reference); imageRef.Digest != ""` assigns to an outer variable inside the `if` init; a plain assignment on its own line followed by `if imageRef.Digest != ""` reads more directly.
- **Follow-up issue.** Please do file the `WithPolicyEnforcement` alignment issue — agreed it doesn't belong in this PR.

registry/remote/repository.go
TerryHowe: Confirmed fixed at 6aaf710. Probed on the branch: under a deny-all signedBy policy, Fetch, Push and Exists on a manifest descriptor are each denied with exactly one verifier call, and a blob descriptor still makes zero verifier calls. The isManifest routing in checkDescriptorPolicy is the right seam.

registry/remote/repository.go
TerryHowe: Confirmed fixed at 6aaf710, and tighter than I asked for. repo.Resolve(ctx, "sha256:...") under a deny-all policy now denies with 0 HTTP requests and exactly 1 verifier call; the allow path is also exactly 1 call, so the checkPolicyResolved short-circuit avoids the double evaluation entirely.

registry/remote/repository.go
TerryHowe: Confirmed fixed at 6aaf710. The shared isImageAllowed(ctx, image, filter) collapses the two loops and keeps the empty-requirements fail-closed invariant in one place. IsResolvedImageAllowed as the symmetric counterpart to IsReferenceAllowed is a better shape than what I suggested.

7487:
Addressed the second-round review in d173ecf:

- Policy references are now fully qualified before identity matching: tags use registry/repository:tag, while descriptor-only and digest references use registry/repository@digest.
- Added an end-to-end regression with the in-tree DefaultSignedByVerifier, a real OpenPGP signature, and lookaside storage. It failed before the fix with the reported signed identity mismatch; both Resolve by tag and Fetch by descriptor now pass.
- Updated the checkPolicy documentation, made the checkPolicyResolved ordering contract explicit, and simplified the digest assignment style.
- Kept the documented Exists/push-side semantics unchanged.
- Filed follow-up #1374 for WithPolicyEnforcement alignment.

Validation: make test passed (full race-enabled suite); go vet ./... and go build ./... passed; golangci-lint v2.11.4 reports 0 issues.

registry/remote/repository.go
TerryHowe: Confirmed fixed at d173ecf. Qualifying via `r.ParseReference(reference).String()` plus synthesizing `desc.Digest.String()` for descriptor-only operations is exactly right.

I verified it end to end with the real `DefaultSignedByVerifier` and a genuine OpenPGP-signed simple-signing payload, at both commits:

| Operation (correctly signed content) | 6aaf710 | d173ecf |
|---|---|---|
| Resolve by tag | DENIED | ALLOWED |
| FetchReference by tag | DENIED | ALLOWED |
| Fetch by descriptor | DENIED | ALLOWED |
| Resolve by digest | DENIED | ALLOWED |

Every pre-fix failure was `signed identity "127.0.0.1:.../test:signed" does not match "signed"`. Worth noting Resolve by digest was broken too, which I had assumed still worked — the bare `sha256:...` string was equally unqualified, so the fix covers more than the case I reported.

Also checked the fix did not make identity matching vacuous: a payload naming `evil.example.com/other:signed` is still rejected with a mismatch error.

TerryHowe (approved):
LGTM at d173ecf. Every finding from both rounds is closed, and I verified each one on the branch rather than by reading.

The last blocking issue — the unqualified reference reaching the verifier — is fixed by qualifying through `r.ParseReference(reference).String()` and synthesizing `desc.Digest.String()` for descriptor-only operations. Driving the real `DefaultSignedByVerifier` with a genuine OpenPGP-signed simple-signing payload:

| Operation (correctly signed content) | 6aaf710 | d173ecf |
|---|---|---|
| `Resolve` by tag | DENIED | ALLOWED |
| `FetchReference` by tag | DENIED | ALLOWED |
| `Fetch` by descriptor | DENIED | ALLOWED |
| `Resolve` by digest | DENIED | ALLOWED |

The digest case was broken too, which I had assumed still worked — so the fix covers more than what I reported. Identity matching is still meaningful afterwards: a payload naming `evil.example.com/other:signed` is rejected with a mismatch error rather than waved through.

Earlier rounds, all re-verified at this commit: manifest `Fetch`/`Push`/`Exists` are denied under a deny-all `signedBy` policy with exactly one verifier call while blob descriptors make zero; a digest reference denies with 0 HTTP requests and exactly 1 verifier call on both the allow and deny paths; and the shared `isImageAllowed(ctx, image, filter)` keeps the empty-requirements fail-closed invariant in one place. `go build ./...`, `go vet ./...` and `go test -race ./registry/...` are all clean.

The two-phase split reads well now, and `TestRepository_DefaultSignedByVerifier_EndToEnd` plus the reference-shape assertions are the right regression guards — a stub verifier that ignores identity is precisely what let this slip through the first time.

Thanks for the thorough turnarounds on this. Two things for after the merge, neither blocking:

- The `WithPolicyEnforcement` middleware still passes raw reference and digest strings to `IsImageAllowed`; please file that follow-up so the two policy paths do not drift.
- `Exists` now runs full signature verification, and `oras.Copy` calls `dst.Exists(desc)` per node. Consistent with the push-side semantics you documented, so I am treating it as intended — but it is worth watching as a per-node cost if anyone reports copy slowdowns under a `signedBy` policy.

TerryHowe:
Thanks!

PR #1370, opened 2026-09-02
fix(auth): prevent scope hints from crossing registry hosts

### LINKED ISSUE #1362: auth: global scope hints are sent to every host contacted in the context
## Problem

Scope hints added with [`auth.WithScopes`](https://github.com/oras-project/oras-go/blob/main/registry/remote/auth/scope.go#L102) / [`auth.AppendScopes`](https://github.com/oras-project/oras-go/blob/main/registry/remote/auth/scope.go#L110) are host-less. `GetAllScopesForHost` merges them into the scope set of *whatever host is asked for*:

```go
func GetAllScopesForHost(ctx context.Context, host string) []string {
	scopes := GetScopesForHost(ctx, host)
	globalScopes := GetScopes(ctx)
	...
}
```

and `Client.Do` calls it with the host of the request in flight (`registry/remote/auth/client.go:274` and `:334`). So a global hint set for one registry is sent to the token endpoint of every other host contacted under the same context — the second registry in a cross-registry copy, a mirror, or a bearer realm on a third origin.

## Reproduction

Registry B is a completely unrelated server; the only hint in the context names a repository on registry A:

```go
c := &Client{CredentialFunc: credentials.StaticCredentialFunc(
	"whatever", credentials.Credential{Username: "u", Password: "p"})}
ctx := WithScopes(context.Background(), "repository:secret-project/internal-app:pull,push")
req, _ := http.NewRequestWithContext(ctx, http.MethodGet, registryB.URL+"/v2/", nil)
c.Do(req)
```

Registry B's token endpoint receives:

```
scopes sent to registry B's token endpoint: ["repository:secret-project/internal-app:pull,push"]
```

## Impact

Scopes are not secrets, so this is not credential exposure. It does mean private repository and namespace names are disclosed to servers that have no business seeing them, which sits badly next to the redirect-hardening in GHSA-vh4v-2xq2-g5cg and GHSA-xf85-363p-868w — same class of "the auth layer volunteers something to an origin the caller never meant to talk to."

There is a functional cost too: the bearer cache key is `strings.Join(scopes, " ")` (`client.go:275`), so a global hint perturbs the cache key for every host, splitting cache entries that should be shared.

## Recommended implementation

Drop the host-less variants entirely and keep only the per-host ones. Nothing in this repository depends on the global form — every internal call site goes through `AppendRepositoryScope`, which is already host-scoped ([`scope.go:78`](https://github.com/oras-project/oras-go/blob/main/registry/remote/auth/scope.go#L78)):

1. Remove `WithScopes`, `AppendScopes`, `GetScopes`, and `scopesContextKey`.
2. Remove `GetAllScopesForHost`; rename `GetScopesForHost` to take its place, or keep both names with `GetAllScopesForHost` as a deprecated alias that just forwards. `Client.Do` then does one lookup with no merge.
3. Callers who genuinely want a hint on several hosts call `WithScopesForHost` once per host — explicit, and it is the correct behaviour they almost certainly already assumed they were getting.

This also deletes the `len(scopes) == 0` / `len(globalScopes) == 0` special-casing and the re-`CleanScopes` pass in `GetAllScopesForHost`.

Removing exported functions is breaking, so this belongs in v3 alongside the other auth changes. If a global form has to survive for migration, the safer shape is one that stores an explicit host set rather than applying to all hosts implicitly.

## Related

- #1348 — namespace-scoped authentication; the scope-hint channel is load-bearing there.
- #873 — granular credentials interface.

pentaoa:
## Summary

- remove the v3 host-less scope context APIs that applied hints to every registry host
- make the auth client read only scopes associated with the request host
- migrate tests and examples to the host-scoped API, including an unrelated-host isolation assertion

Closes #1362

## Testing

- `go test -race ./registry/remote/auth`
- `go test -race ./...` (all relevant packages passed; two environment-dependent tests could not access the macOS user policy path or a system `registries.conf`)

AI-assisted tooling was used for implementation and review.

### DIFF
--- a/registry/remote/auth/client.go
+++ b/registry/remote/auth/client.go
@@ -288,7 +288,7 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
 				req.Header.Set(headerAuthorization, "Basic "+token)
 			}
 		case SchemeBearer:
-			scopes := GetAllScopesForHost(ctx, host)
+			scopes := GetScopesForHost(ctx, host)
 			attemptedKey = strings.Join(scopes, " ")
 			token, err := cache.GetToken(ctx, host, SchemeBearer, attemptedKey)
 			if err == nil {
@@ -331,7 +331,7 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
 	case SchemeBearer:
 		resp.Body.Close()
 
-		scopes := GetAllScopesForHost(ctx, host)
+		scopes := GetScopesForHost(ctx, host)
 		if paramScope := params["scope"]; paramScope != "" {
 			// merge hinted scopes with challenged scopes
 			scopes = append(scopes, strings.Split(paramScope, " ")...)
--- a/registry/remote/auth/client_test.go
+++ b/registry/remote/auth/client_test.go
@@ -392,7 +392,7 @@ func TestClient_Do_Bearer_AccessToken_Cached(t *testing.T) {
 	}
 
 	// first request
-	ctx := WithScopes(context.Background(), scope)
+	ctx := WithScopesForHost(context.Background(), uri.Host, scope)
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, ts.URL, nil)
 	if err != nil {
 		t.Fatalf("failed to create test request: %v", err)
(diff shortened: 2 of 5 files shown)


TerryHowe (changes requested):
This comment was wrong, reevaluating.

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

Issue #1361, opened 2026-09-01
ci: release workflow cannot push tags; needs RELEASE_PAT to bypass tag ruleset


TerryHowe:
## Problem

The `release` workflow cannot push the version tag, so every release requires a manual workaround by a repo admin.

`.github/workflows/release.yml` (present on both `v2` and `main`) creates and pushes the tag itself:

```yaml
      - name: Checkout
        uses: actions/checkout@...
        with:
          fetch-depth: 0
          ref: ${{ github.event.pull_request.merge_commit_sha }}
      # ...
      - name: Create and push tag
        run: |
          git config user.name "github-actions[bot]"
          git config user.email "github-actions[bot]@users.noreply.github.com"
          git tag ${{ steps.version.outputs.version }} ${{ github.event.pull_request.merge_commit_sha }}
          git push origin ${{ steps.version.outputs.version }}
```

`actions/checkout` has no `token:` override, so the push authenticates as `github-actions[bot]` via the default `GITHUB_TOKEN`.

Repository ruleset **#1628459 "Auto-imported tag create protections"** targets tags, includes `~ALL`, and enforces `creation` + `update`. Its bypass actors are only `OrganizationAdmin` and two built-in repository roles. `github-actions[bot]` matches none of them, so the push fails:

```
! [remote rejected] vX.Y.Z -> vX.Y.Z (GH013: Repository rule violations found ... Cannot create ref due to creations being restricted)
```

Note this is a ruleset bypass problem, not a token *scope* problem — the job already has `permissions: contents: write`.

Dead ends already tried:

- Adding the GitHub Actions app (id 15368) as an `Integration` bypass actor → 422, it is not an org-installed integration.
- Adding the `github-actions[bot]` user (id 41898282) as a `User` bypass actor → accepted by the API, but ignored at push time.

## Current workaround

Every release, an admin temporarily disables the ruleset, re-runs the failed job, and re-enables it:

```sh
gh api -X PUT repos/oras-project/oras-go/rulesets/1628459 -f enforcement=disabled
gh run rerun <RUN_ID> --repo oras-project/oras-go --failed
# wait for completion, verify tag + release
gh api -X PUT repos/oras-project/oras-go/rulesets/1628459 -f enforcement=active
```

This leaves tag-creation protection off repo-wide for the duration, depends on one person being available, and is easy to forget to undo. `v2.6.2` shipped this way.

## Proposed fix

[...]

PR #1371, opened 2026-09-02
ci: authenticate v2 release tag push with RELEASE_PAT

LarryHu0217:
Refs #1361.

The v2 release workflow currently checks out with the default `GITHUB_TOKEN`, whose `github-actions[bot]` identity cannot bypass the repository's protected-tag ruleset. This makes the release job fail when it pushes the version tag.

Configure `actions/checkout` to persist `secrets.RELEASE_PAT` for the tag push. Inline guidance records that the PAT must belong to an organization owner covered by the ruleset bypass and must be rotated before expiration. GoReleaser continues to use `GITHUB_TOKEN`.

Validation: parsed `.github/workflows/release.yml` successfully, asserted the checkout token is `secrets.RELEASE_PAT`, and passed `git diff --check`. `actionlint` and `go` are not installed on this host, so no actionlint run was performed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,9 @@ jobs:
         with:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.merge_commit_sha }}
+          # Must belong to an org owner so the protected-tag ruleset bypass applies.
+          # Rotate the repository secret before its PAT expires.
+          token: ${{ secrets.RELEASE_PAT }}
 
       - name: Set up Go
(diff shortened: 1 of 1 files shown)


TerryHowe (changes requested):
https://github.com/oras-project/oras-go/pull/1371/checks?check_run_id=100461740039 DCO failed

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1378, opened 2026-09-04
docs: clarify repository constructor migration

### LINKED ISSUE #1372: docs: migration guide points at NewRepositoryWithProperties as a struct-literal replacement
`MIGRATION_GUIDE.md:51-55` (added in #1367) says:

> Note that `Reference()` is a read-only getter; the reference can no longer be
> set through it. The v2 struct-literal pattern
> `&remote.Repository{Reference: ref, Client: c}` does not compile in v3;
> construct repositories with `remote.NewRepository` or
> `remote.NewRepositoryWithProperties` instead.

Neither named constructor accepts a `Client`, so a reader migrating that exact
literal is pointed at two functions, neither of which covers the `Client: c`
half of it:

- `NewRepository(reference string) (*Repository, error)` —
  `registry/remote/repository.go:185` — takes only a reference string.
- `NewRepositoryWithProperties(props *properties.Registry, builder *ClientBuilder) (*Repository, error)` —
  `registry/remote/builder.go:320` — takes a structured `properties.Registry`
  plus a `ClientBuilder`. It is not a drop-in for a hand-built struct literal.

The actual replacement is `NewRepository(...)` followed by
`repo.Registry.Client = c`, which the example block immediately below
(`MIGRATION_GUIDE.md:58-81`) already demonstrates. The note sends the reader
away from the answer that is four lines further down the page.

There is also an internal inconsistency. Line 55 offers
`NewRepositoryWithProperties` as a general-purpose alternative, while
`MIGRATION_GUIDE.md:92-95` scopes it correctly:

> `remote.NewRegistryWithProperties` and
> `remote.NewRepositoryWithProperties` are the preferred constructors when
> configuration comes from `registries.conf`, Docker configuration,
> certificates, or policy files.

## Suggested fix

Point the note at the example that follows it, and either drop
`NewRepositoryWithProperties` from the sentence or scope it the way line 93
does.

## Acceptance criteria

- [ ] The note no longer implies `NewRepositoryWithProperties` replaces a
      `&remote.Repository{Reference: ..., Client: ...}` literal
- [ ] The note directs the reader to the `NewRepository` +
      `repo.Registry.Client = ...` example that follows

Verified against `main` @ f932f1c. Docs-only.

TheJhyeFactor:
## What changed

The migration note now points the v2 repository struct-literal example at the
actual v3 replacement shown immediately below it:

1. construct the repository with `remote.NewRepository(...)`
2. configure the shared client through `repo.Registry.Client`

`remote.NewRepositoryWithProperties` is no longer presented as a drop-in
replacement for a literal containing `Client: c`. Its separate, correct use
for configuration-file-driven setup remains documented later in the section.

## Verification

- `go test ./...` — Go 1.27.1, `darwin/arm64`
- `git diff --check`
- rendered `MIGRATION_GUIDE.md` through GitHub's GFM renderer and checked the
  corrected note and the later `NewRepositoryWithProperties` guidance

Closes #1372

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -51,8 +51,8 @@ sizes remain on `remote.Repository`.
 Note that `Reference()` is a read-only getter; the reference can no longer be
 set through it. The v2 struct-literal pattern
 `&remote.Repository{Reference: ref, Client: c}` does not compile in v3;
-construct repositories with `remote.NewRepository` or
-`remote.NewRepositoryWithProperties` instead.
+construct the repository with `remote.NewRepository`, then configure its
+shared client through `repo.Registry.Client`, as shown below.
 
 For example:
 


TerryHowe (approved):
/lgtm

PR #1391, opened 2026-09-04
auth: add a Resource type for registry scopes

### LINKED ISSUE #1380: auth: add a Resource type for a registry host optionally narrowed to a namespace

## Problem

There is no type in the library that represents *"a registry, optionally narrowed to a namespace or repository, with no tag or digest"* — the thing a user names when they run `oras login example.com/myspace`.

`properties.NewReference` is not it, in both directions:

- It **requires a path**. `splitRegistry` (`registry/remote/properties/reference.go:224`) returns an empty path for `"example.com"`, and parsing fails with `missing registry or repository` (`reference.go:100`).
- It **accepts tags and digests**, which are never valid for a login.

So neither the accept set nor the reject set matches.

## Proposal

A value type plus a parser:

```go
// Resource is a registry host, optionally narrowed to a namespace or
// repository. It never carries a tag or digest.
type Resource struct {
    Host string // host[:port], as dialed
    Path string // "" | "myspace" | "myspace/app"
}

func ParseResource(s string) (Resource, error)
```

Parsing rules:

- Strip `oci://`, `http://`, `https://`, as `NewReference` does.
- Split the host off **first**, then inspect the path. Splitting the other way makes `localhost:5000` parse `5000` as a tag.
- Empty path is valid and means the whole registry.
- Reject `:` or `@` in the path explicitly, so the error reads *"must not include a tag or digest"* rather than *"invalid repository"*.
- Validate using the same rules as `properties.Reference.ValidateRegistry` / `properties.Reference.ValidateRepository` (`registry/remote/properties/reference.go:253`, `:261`). Note these are **methods on `properties.Reference`**, not free functions, so this issue either extracts the shared check or constructs a `properties.Reference` internally to validate. Use `properties.Reference`, **not** `registry.Reference` — the latter is deprecated (`registry/reference.go:52`).

Deliberately **no attempt to distinguish a namespace from a repository** — `repositoryRegexp` (`registry/remote/properties/reference.go:38`) allows multi-segment paths, so `myspace` and `myspace/app` are syntactically identical. Which one it is, is the server's business.

## Accept / reject set

| Input | Result |
|---|---|
| `example.com` | `{Host: "example.com", Path: ""}` |
| `example.com/myspace` | `{Host: "example.com", Path: "myspace"}` |
| `example.com/myspace/app` | `{Host: "example.com", Path: "myspace/app"}` |
| `localhost:5000/ns` | `{Host: "localhost:5000", Path: "ns"}` |
| `oci://example.com/ns` | `{Host: "example.com", Path: "ns"}` |
| `example.com/myspace:v1` | error — must not include a tag |
| `example.com/myspace@sha256:...` | error — must not include a digest |

## Open question — where it lives and what it is called

`registry/remote/properties` sits next to `Reference` and is where new parsing belongs; a dedicated package would let it be plain `Scope` without stutter.

Naming is genuinely open, because this value is both a parse result *and* a resolution parameter. `RegistryScope` (the name in the original #1348 write-up) reads oddly on `CredentialFunc`; `Scope` collides with the bearer-token scopes in `registry/remote/auth`. `Resource` is used throughout these issues as a working name, not a decision.

Prior art for the vocabulary: `GetLookasideURLs(scope string)` already matches `registry.example.com/namespace` by longest prefix (`registry/remote/config/registriesd.go:60-67`), following containers-image.

## Scope of this issue

Type, parser, tests. **No behaviour change** — nothing consumes it yet. It is the prerequisite for the rest of the split.

alexsmolya:
Resolves #1380

Adds a Resource value and parser for a registry host optionally narrowed to a namespace or repository, without tags or digests. Includes focused acceptance and edge-case tests.

### DIFF
--- a/registry/remote/properties/resource.go
+++ b/registry/remote/properties/resource.go
@@ -1,0 +1,66 @@
+/*
+Copyright The ORAS Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package properties
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/oras-project/oras-go/v3/errdef"
+)
+
+// Resource represents a registry host, optionally narrowed to a namespace or
+// repository. It never carries a tag or digest.
+type Resource struct {
+	// Host is the name of the registry, usually a domain name optionally with a
+	// port.
+	Host string
+
+	// Path is the namespace or repository path, or empty for the whole registry.
+	Path string
+}
+
+// ParseResource parses a registry resource. The resource may contain an
+// optional oci://, http://, or https:// scheme, and may be narrowed to a
+// namespace or repository path. Tags and digests are not accepted.
+func ParseResource(resource string) (Resource, error) {
+	resource = strings.TrimPrefix(resource, "oci://")
+	resource = strings.TrimPrefix(resource, "http://")
+	resource = strings.TrimPrefix(resource, "https://")
+	if strings.HasSuffix(resource, "/") {
+		return Resource{}, fmt.Errorf("%w: invalid resource path %q", errdef.ErrInvalidReference, resource)
+	}
+
+	host, path := splitRegistry(resource)
+	if host == "" {
+		return Resource{}, fmt.Errorf("%w: invalid registry resource %q", errdef.ErrInvalidReference, resource)
+	}
+
+	// Validate the host independently so a port is not mistaken for a tag.
+	if err := (Reference{Registry: host}).ValidateRegistry(); err != nil {
+		return Resource{}, err
+	}
+	if path == "" {
+		return Resource{Host: host}, nil
+	}
+	if strings.ContainsAny(path, ":@") {
+		return Resource{}, fmt.Errorf("%w: resource must not include a tag or digest", errdef.ErrInvalidReference)
+	}
+	if err := (Reference{Repository: path}).ValidateRepository(); err != nil {
+		return Resource{}, err
+	}
+	return Resource{Host: host, Path: path}, nil
+}
--- a/registry/remote/properties/resource.go
+++ b/registry/remote/properties/resource.go
@@ -25,9 +25,9 @@ import (
 // Resource represents a registry host, optionally narrowed to a namespace or
 // repository. It never carries a tag or digest.
 type Resource struct {
-	// Host is the name of the registry, usually a domain name optionally with a
-	// port.
-	Host string
+	// Registry is the name of the registry, usually a domain name optionally
+	// with a port.
+	Registry string
 
 	// Path is the namespace or repository path, or empty for the whole registry.
 	Path string
@@ -54,7 +54,7 @@ func ParseResource(resource string) (Resource, error) {
 		return Resource{}, err
 	}
 	if path == "" {
-		return Resource{Host: host}, nil
+		return Resource{Registry: host}, nil
 	}
 	if strings.ContainsAny(path, ":@") {
 		return Resource{}, fmt.Errorf("%w: resource must not include a tag or digest", errdef.ErrInvalidReference)
@@ -62,5 +62,23 @@ func ParseResource(resource string) (Resource, error) {
 	if err := (Reference{Repository: path}).ValidateRepository(); err != nil {
 		return Resource{}, err
 	}
-	return Resource{Host: host, Path: path}, nil
+	return Resource{Registry: host, Path: path}, nil
+}
+
+// Host returns the host name of the registry.
+//
+// For docker.io, it returns registry-1.docker.io.
+func (r Resource) Host() string {
+	if r.Registry == "docker.io" {
+		return "registry-1.docker.io"
+	}
+	return r.Registry
+}
+
+// String returns the resource as a string.
+func (r Resource) String() string {
+	if r.Path == "" {
+		return r.Registry
+	}
+	return r.Registry + "/" + r.Path
 }
--- a/registry/remote/properties/resource.go
+++ b/registry/remote/properties/resource.go
@@ -38,15 +38,16 @@ type Resource struct {
 // optional oci://, http://, or https:// scheme, and may be narrowed to a
 // namespace or repository path. Tags and digests are not accepted.
 func ParseResource(resource string) (Resource, error) {
-	resource = strings.TrimPrefix(resource, "oci://")
-	resource = strings.TrimPrefix(resource, "http://")
-	resource = strings.TrimPrefix(resource, "https://")
-	resource = strings.TrimSuffix(resource, "/")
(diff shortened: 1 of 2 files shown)


registry/remote/properties/resource.go
```diff
@@ -0,0 +1,66 @@
...
+// repository. It never carries a tag or digest.
+type Resource struct {
+	// Host is the name of the registry, usually a domain name optionally with a
+	// port.
+	Host string
+
+	// Path is the namespace or repository path, or empty for the whole registry.
+	Path string
+}
+
```
TerryHowe, line 35:
```suggestion

// Host returns the host name of the registry.
// For docker.io, it returns registry-1.docker.io.
func (r Resource) Host() string {
    if r.Registry == "docker.io" {
        return "registry-1.docker.io"
    }
    return r.Registry
}

// String returns the resource as a string.
// The format is: registry/repository[:tag][@digest]
func (r Resource) String() string {
    if r.Path == "" {
        return r.Registry
    }
    ref := r.Registry + "/" + r.Path
    return ref
}

```

registry/remote/properties/resource.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/oras-project/oras-go/v3/errdef"
+)
+
+// Resource represents a registry host, optionally narrowed to a namespace or
+// repository. It never carries a tag or digest.
+type Resource struct {
+	// Host is the name of the registry, usually a domain name optionally with a
+	// port.
+	Host string
```
TerryHowe: For consistency, please rename this Registry

alexsmolya:
Implemented the requested `Resource.Registry`, `Resource.Host()`, and `Resource.String()` changes, with focused tests added.

registry/remote/properties/resource.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// ParseResource parses a registry resource. The resource may contain an
+// optional oci://, http://, or https:// scheme, and may be narrowed to a
+// namespace or repository path. Tags and digests are not accepted.
+func ParseResource(resource string) (Resource, error) {
+	resource = strings.TrimPrefix(resource, "oci://")
+	resource = strings.TrimPrefix(resource, "http://")
+	resource = strings.TrimPrefix(resource, "https://")
+	if strings.HasSuffix(resource, "/") {
```
TerryHowe, line 43: `https://example.com/` fails here. A pasted registry URL with a trailing slash is a realistic `oras login` input. Prefer trimming one trailing `/` over erroring; `"/"` alone then falls through to the empty-input error, which is right.

registry/remote/properties/resource.go
```diff
@@ -0,0 +1,84 @@
...
+		return Resource{}, fmt.Errorf("%w: invalid resource path %q", errdef.ErrInvalidReference, resource)
+	}
+
+	host, path := splitRegistry(resource)
+	if host == "" {
+		return Resource{}, fmt.Errorf("%w: invalid registry resource %q", errdef.ErrInvalidReference, resource)
+	}
+
+	// Validate the host independently so a port is not mistaken for a tag.
+	if err := (Reference{Registry: host}).ValidateRegistry(); err != nil {
```
TerryHowe, line 53: `EXAMPLE.com/ns` parses and keeps `Registry: "EXAMPLE.com"`. Hostnames are case-insensitive, and this value is headed for longest-prefix credential lookup, so `ghcr.io` and `GHCR.IO` silently become two keys. Suggest lowercasing the host here (`Path` correctly stays case-sensitive). Adding it later is a behavior change — worth deciding now, and documenting on the `Registry` field either way.

registry/remote/properties/resource_test.go
```diff
@@ -0,0 +1,117 @@
...
+		"example.com/",
+		"example.com/UPPER",
+		"example.com/a//b",
+		"example.com/a b",
+		"invalid host/ns",
+		"",
+	}
+	for _, input := range tests {
+		t.Run(strings.ReplaceAll(input, "/", "_"), func(t *testing.T) {
+			if _, err := ParseResource(input); err == nil {
```
TerryHowe, line 63: Nothing in the suite asserts `errors.Is(err, errdef.ErrInvalidReference)`, which is the package's error contract. A refactor returning a bare `fmt.Errorf` would still pass. Please check the sentinel here.

alexsmolya:
Implemented the three requested corrections:

- `ParseResource` trims one trailing slash while preserving invalid `/` and repeated-slash handling.
- Registry authorities are normalized to lowercase while preserving `Path` and ports.
- Negative parser tests assert `errors.Is(err, errdef.ErrInvalidReference)`.

Focused parser tests, race tests, `go vet ./...`, encoding checks, and `git diff --check` pass. The repository-wide test run remains environment-limited by IPv6 socket permissions, and `golangci-lint` was unavailable locally.

registry/remote/properties/resource.go
```diff
@@ -0,0 +1,87 @@
...
+func ParseResource(resource string) (Resource, error) {
+	resource = strings.TrimPrefix(resource, "oci://")
+	resource = strings.TrimPrefix(resource, "http://")
+	resource = strings.TrimPrefix(resource, "https://")
+	resource = strings.TrimSuffix(resource, "/")
+	if strings.HasSuffix(resource, "/") {
+		return Resource{}, fmt.Errorf("%w: invalid resource path %q", errdef.ErrInvalidReference, resource)
+	}
+
+	host, path := splitRegistry(resource)
```
TerryHowe: Thanks — the trailing-slash trim is right, and I've resolved my earlier threads; they're all addressed.

One follow-on from the trim: `resource` is reassigned in place, so both error messages now quote the post-trim string rather than what the caller passed. `ParseResource("https://example.com//")` reports `invalid resource path "example.com/"` — a value the user never typed, and one that would actually parse fine on its own. `ParseResource("/")` similarly reports `invalid registry resource ""`.

Parsing on a separate variable keeps the original for the error text:

```suggestion
	trimmed := strings.TrimPrefix(resource, "oci://")
	trimmed = strings.TrimPrefix(trimmed, "http://")
	trimmed = strings.TrimPrefix(trimmed, "https://")
	// A single trailing slash is tolerated; more than one is not a valid path.
	trimmed = strings.TrimSuffix(trimmed, "/")
	if strings.HasSuffix(trimmed, "/") {
		return Resource{}, fmt.Errorf("%w: invalid resource path %q", errdef.ErrInvalidReference, resource)
	}

	host, path := splitRegistry(trimmed)
```

| Input | Before | After |
|---|---|---|
| `example.com//` | `invalid resource path "example.com/"` | `invalid resource path "example.com//"` |
| `https://example.com//` | `invalid resource path "example.com/"` | `invalid resource path "https://example.com//"` |
| `example.com/myspace//` | `invalid resource path "example.com/myspace/"` | `invalid resource path "example.com/myspace//"` |
| `/` | `invalid registry resource ""` | `invalid registry resource "/"` |

The accept/reject set is unchanged — only the message text. I applied this locally on `24551c9`: package tests, `go vet`, `gofmt -l`, and `golangci-lint run` are all clean.

The comment also documents the single-slash tolerance, which is currently implicit. Worth noting the `HasSuffix` check is genuinely load-bearing and not redundant with `ValidateRepository` below — `splitRegistry` splits on the first `/`, so `example.com//` yields an empty path and would otherwise parse as a bare registry.

A case to lock the message in, since nothing asserts it today:

```go
{name: "double trailing slash echoes input", input: "https://example.com//"},
```

paired with a `strings.Contains(err.Error(), input)` assertion rather than only `errors.Is`.

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1392, opened 2026-09-04
refactor(remote): use properties references internally

### LINKED ISSUE #1386: registry/remote: migrate internal use of deprecated registry.Reference to properties.Reference
`registry.Reference` is deprecated in favour of `properties.Reference` (`registry/reference.go:52`), but is still used internally. This issue covers the portion that is **entirely below the exported API surface**, so it can land on its own with no user-visible change and no release note.

## In scope

All unexported, all in `registry/remote`:

| Location | Use |
|---|---|
| `url.go:38, 45, 51, 58, 65, 76, 87, 94, 105` | nine unexported URL builders take `ref registry.Reference` |
| `repository.go:2051` | `generateDescriptor(..., ref registry.Reference, ...)` |
| `repository.go:209` | local variable inside `Repository.Reference()` |
| `registry.go:108, 251` | local variables |
| `builder.go:249, 306, 388` | `registry.Reference{...}` construction |

`properties.Reference` has the same fields plus `Host()`, so these are mechanical.

## Out of scope

The exported surface — `Registry.Reference`, `Repository.Reference()`, `Repository.ParseReference`, `interfaces.ReferenceParser` — is tracked separately, because it is a breaking change with a subtle hazard. `registry/remote/auth`'s `AppendRepositoryScope` is #1385.

## Caveat on the boundary

Three of the in-scope sites — `builder.go:306, 388`, `repository.go:196` — construct a value that is assigned to the **exported** `Registry.Reference` field (`registry/remote/registry.go:41`). They can only change type once that field does.

So this issue splits one of two ways, and whoever picks it up should choose deliberately:

- **Narrow:** do only the sites that never cross an exported boundary — `url.go`, `generateDescriptor`, and the local variables — converting at the boundary where needed. Fully independent.
- **Wide:** wait for the exported-surface issue and do both together, avoiding a temporary conversion shim.

The narrow reading is what makes this issue independently pickup-able, and is the recommended one.

## Verification

`git grep -nE 'registry\.Reference([^A-Za-z]|$)'` should come back empty for `registry/remote/url.go` when this is done. Note that a plain grep for `registry.Reference` also matches `registry.ReferenceFetcher` and `registry.ReferencePusher`, which are **not** deprecated and must not be touched.

MarthalaJagruthiReddy:
## Summary

- migrate internal remote URL, repository, registry, and builder code to `registry/remote/properties.Reference`
- preserve the public legacy reference APIs and the existing auth scope compatibility boundary
- update regression coverage for the internal reference flow

Fixes #1386

## Tests

- `make test`

### DIFF
--- a/registry/remote/builder.go
+++ b/registry/remote/builder.go
@@ -246,7 +246,7 @@ func (b *ClientBuilder) buildCredentialFunc(props *properties.Registry) credenti
 	// Match what the transport actually dials: Host() maps the "docker.io"
 	// alias onto "registry-1.docker.io", and the credential func is called with
 	// the request host.
-	host := registry.Reference{Registry: props.Reference.Registry}.Host()
+	host := props.Reference.Host()
 
 	return func(ctx context.Context, reg string) (credentials.Credential, error) {
 		// Credential specified in properties, for its own registry only.
--- a/registry/remote/registry.go
+++ b/registry/remote/registry.go
@@ -29,6 +29,7 @@ import (
 	"github.com/oras-project/oras-go/v3/registry/remote/auth"
 	"github.com/oras-project/oras-go/v3/registry/remote/internal/errutil"
 	"github.com/oras-project/oras-go/v3/registry/remote/policy"
+	"github.com/oras-project/oras-go/v3/registry/remote/properties"
 )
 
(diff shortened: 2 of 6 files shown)


TerryHowe (approved):
/lgtm

PR #1393, opened 2026-09-05
refactor(remote)!: use properties references in exported APIs

### LINKED ISSUE #1387: registry/remote: exported API still uses the deprecated registry.Reference
`registry.Reference` is deprecated in favour of `properties.Reference` (`registry/reference.go:52`), but it is still part of the **exported** API surface of `registry/remote`. This is the breaking half of the migration; the internal-only half is #1386 and the auth package is #1385.

## In scope

| Location | Surface |
|---|---|
| `registry/remote/registry.go:41` | `Registry.Reference registry.Reference` — exported struct field |
| `registry/remote/repository.go:208` | `func (r *Repository) Reference() registry.Reference` |
| `registry/remote/repository.go:723` | `func (r *Repository) ParseReference(string) (registry.Reference, error)` |
| `registry/remote/repository.go:2045` | `manifestStore.ParseReference` — see below |
| `internal/interfaces/registry.go:23` | `ReferenceParser.ParseReference` — internal package, but see the hazard |

These are one coupled cluster: the field type, the two `ParseReference` implementations, and the interface they satisfy all have to move together.

## The hazard: a silent behaviour regression, not a compile error

`interfaces.ReferenceParser` is matched by **structural type assertion**, not by declaration:

```go
// content.go:87 and content.go:145
if repo, ok := target.(interfaces.ReferenceParser); ok {
    // add scope hints to minimize the number of auth requests
    ref, err := repo.ParseReference(srcReference)
    ...
    ctx = auth.AppendRepositoryScope(ctx, ref, auth.ActionPull, auth.ActionPush)
}
```

Any third-party `Target` that today declares `ParseReference(string) (registry.Reference, error)` satisfies this interface and gets scope hints. **The moment the interface's signature changes to `properties.Reference`, those implementations stop matching — silently.** There is no compile error, because the assertion simply fails and the `if` body is skipped. The only symptom is degraded scope hinting: more token round trips, and in the worst case a token that lacks a needed scope.

This is worth an explicit release note, and possibly a transition period where both signatures are accepted at the assertion sites.

## Worth confirming: `manifestStore.ParseReference` is *not* on an exported interface

`Repository.Manifests()` returns `registry.ManifestStore` (`repository.go:621`), but that interface is `BlobStore` + `Tagger` + `Untagger` + `ReferencePusher` (`registry/interfaces.go:67`) and does **not** include `ParseReference`. So `manifestStore.ParseReference` (`repository.go:2045`) is reachable only through the structural assertion above, not through a declared exported contract. It still has to change in lockstep, but it is not itself a documented API break.

## Suggested order

1. #1386 — internal sites, no API change.
2. This issue — the exported cluster, with a release note for the assertion hazard.
3. #1385 — `AppendRepositoryScope`, which consumes the output of `ParseReference` and is easiest once this has landed.

## Note

Do not touch `registry.ReferenceFetcher` or `registry.ReferencePusher` — despite the name prefix, they are not deprecated.

pentaoa:
## What

- migrate `remote.Registry.Reference`, `Repository.Reference`, and both `ParseReference` implementations to `properties.Reference`
- remove the legacy conversion layer from remote repository operations
- retain a temporary structural fallback for third-party targets whose `ParseReference` method still returns `registry.Reference`
- document the exported API migration and the compatibility window

## Why

This completes the exported half of the reference migration after #1392 while preventing existing third-party targets from silently losing authentication scope hints.

Closes #1387.

## Testing

- `go test -mod=mod ./registry/remote .`
- `go test -mod=mod -race -cover ./registry/remote .` (90.0% / 89.5% statement coverage)
- the remaining packages compile and their tests pass except two host-configuration tests that require access to system config paths unavailable in the local sandbox

### DIFF
--- a/registry/remote/repository.go
+++ b/registry/remote/repository.go
@@ -183,7 +183,7 @@ type Repository struct {
 // carrying one is rejected with an error wrapping
 // [errdef.ErrInvalidReference].
 func NewRepository(reference string) (*Repository, error) {
-	ref, err := registry.ParseReference(reference)
+	ref, err := properties.NewReference(reference)
 	if err != nil {
 		return nil, err
 	}
@@ -193,7 +193,7 @@ func NewRepository(reference string) (*Repository, error) {
 
 	// Create a default Registry
 	reg := &Registry{
-		Reference: registry.Reference{
+		Reference: properties.Reference{
 			Registry: ref.Registry,
 		},
 	}
@@ -213,25 +213,6 @@ func (r *Repository) reference() properties.Reference {
 	return ref
 }
 
-// toPropertiesReference converts a legacy registry reference at an exported
-// API boundary to the reference representation used internally by remote.
-func toPropertiesReference(ref registry.Reference) properties.Reference {
-	propertiesRef := properties.Reference{
-		Registry:   ref.Registry,
-		Repository: ref.Repository,
-		Tag:        ref.Tag,
-		Digest:     ref.Digest,
-	}
-	if propertiesRef.GetReference() == "" && ref.Reference != "" {
-		if d, err := digest.Parse(ref.Reference); err == nil {
-			propertiesRef.Digest = d.String()
-		} else {
-			propertiesRef.Tag = ref.Reference
-		}
-	}
-	return propertiesRef
-}
-
 // appendRepositoryScope adapts the internal reference representation to the
 // legacy authentication API.
 func appendRepositoryScope(ctx context.Context, ref properties.Reference, actions ...string) context.Context {
@@ -241,13 +222,9 @@ func appendRepositoryScope(ctx context.Context, ref properties.Reference, action
 	}, actions...)
 }
 
-// Reference returns the full registry.Reference for this repository.
-func (r *Repository) Reference() registry.Reference {
-	ref := r.reference()
-	return registry.Reference{
-		Registry:   ref.Registry,
-		Repository: ref.Repository,
-	}
+// Reference returns the full properties.Reference for this repository.
+func (r *Repository) Reference() properties.Reference {
+	return r.reference()
 }
 
 // clone makes a copy of the Repository being careful not to copy non-copyable fields (sync.Mutex and syncutil.Pool types)
@@ -573,7 +550,7 @@ func (r *Repository) policyDigest(reference string) digest.Digest {
 	if err != nil {
 		return ""
 	}
-	d, err := digest.Parse(ref.Reference)
+	d, err := digest.Parse(ref.GetReference())
 	if err != nil {
 		return ""
 	}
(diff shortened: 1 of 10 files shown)


registry/remote/repository.go
```diff
@@ -757,37 +734,40 @@ func (r *Repository) FetchReference(ctx context.Context, reference string) (ocis
 // and returns the parsed reference. If the parsed reference does not share
 // the same base reference with the Repository r, ParseReference returns a
 // wrapped error ErrInvalidReference.
-func (r *Repository) ParseReference(reference string) (registry.Reference, error) {
+func (r *Repository) ParseReference(reference string) (properties.Reference, error) {
```
TerryHowe, line 737: Form B (`tag@digest`) now returns a reference with both `Tag` and `Digest` set, and `properties.Reference.String()` renders both — whereas `registry.Reference.String()` rendered only the digest.

That changes what `policyImageReference` (line ~531) hands to policy verifiers:

- before: `registry.example.com/hello-world@sha256:b94d…`
- after: `registry.example.com/hello-world:foobar@sha256:b94d…`

That string reaches third-party `SignedByVerifier`/`SigstoreVerifier` implementations via `policy.ImageReference.Reference`. Tag-only and digest-only paths are unchanged; this is specific to Form B. Either clear `Tag` when `Digest` is set before stringifying, or accept it deliberately and add a `repository_policy_test.go` case pinning the new string — nothing in the suite covers `tag@digest` today, so this is currently invisible.

registry/remote/repository.go
```diff
@@ -1808,12 +1783,11 @@ func (s *manifestStore) Untag(ctx context.Context, reference string) error {
 	if err != nil {
 		return err
 	}
-	if err := ref.ValidateReferenceAsTag(); err != nil {
-		return err
+	if ref.Tag == "" || ref.Digest != "" {
+		return fmt.Errorf("%w: invalid tag %q", errdef.ErrInvalidReference, ref.GetReference())
```
TerryHowe, line 1787: This structural check replaces `ValidateReferenceAsTag()`. It is safe today — every `ParseReference` path that leaves `Digest` empty runs `ValidateTag()`, and Form B sets `Digest` so it is rejected here — but the guarantee now lives in a different function than the check, so a later tweak to `ParseReference` could let an unvalidated tag reach `buildRepositoryManifestURL`.

Suggest keeping it explicit:

```go
if ref.Digest != "" {
    return fmt.Errorf("%w: invalid tag %q", errdef.ErrInvalidReference, ref.GetReference())
}
if err := ref.ValidateTag(); err != nil {
    return err
}
```

pentaoa:
Addressed both review comments in 7c430e9:

- Preserve the previous digest-only identity passed to policy verifiers for `tag@digest` references. A regression test records the actual `SignedByVerifier` input for both short and fully qualified references.
- Explicitly call `ValidateTag()` in `Untag` after rejecting digest references.

The targeted policy/Untag tests and `make test` (full race-enabled suite and coverage) pass locally.

TerryHowe (approved):
/lgtm

PR #1397, opened 2026-09-05
Match policy scopes that name an exact tag or digest

### LINKED ISSUE #1390: policy.json scopes naming a tag or digest never match and are silently ignored
### What

`containers-policy.json` allows a docker transport scope to name an exact tagged or digested reference, e.g. `"docker.io/library/busybox:latest"`. Such an entry can never match, because the scope this library computes for an image never includes a tag or digest. Evaluation falls through to the next-most-specific scope, or to `default`.

### Where

`registry/remote/repository.go:513`:

```go
return policy.ImageReference{
	Transport: policy.TransportNameDocker,
	Scope:     repoRef.Registry + "/" + repoRef.Repository,
	Reference: ref,
}
```

`Scope` is always `registry/repository`. In `GetRequirementsForImage` (`registry/remote/policy/policy.go:216`) the exact-match lookup then misses, and `isPathPrefix("docker.io/library/busybox:latest", "docker.io/library/busybox")` is false because the key is longer than the scope, so the longest-prefix pass misses too.

### Reproduction

```go
package policy

import "testing"

func TestTaggedScopeIgnored(t *testing.T) {
	p := &Policy{
		Default: PolicyRequirements{&InsecureAcceptAnything{}},
		Transports: map[TransportName]TransportScopes{
			TransportNameDocker: {
				"docker.io/library/busybox:v1": PolicyRequirements{&Reject{}},
			},
		},
	}
	reqs := p.GetRequirementsForImage(TransportNameDocker, "docker.io/library/busybox")
	t.Logf("reqs=%v", reqs[0].Type()) // insecureAcceptAnything
}
```

### Impact

A policy author writes a tag-scoped rule, the file loads and validates without complaint, and the rule is silently inert. If the rule was tightening something relative to `default` — a `reject` or a `signedBy` on one tag — the result is a silent downgrade rather than a visible error.

Scope: v3 only; `v3.0.0-rc.1` is the only tag carrying `registry/remote/policy`.

### Suggested fix

Two parts, either or both:

1. Include the tag/digest in the computed scope so the containers-policy.json precedence chain (exact tagged ref → repo → namespace → host → transport default → global default) can be walked in full. `GetRequirementsForImage` would try the fully-qualified reference first, then progressively shorter namespaces.
2. Failing that, have `Policy.Validate` reject (or at least surface) docker-transport scope keys containing `:` or `@` that can never match, so a policy author finds out at load time instead of never.

Related: #1388 (scope matching is also case-sensitive).

arpitjain099:
`containers-policy.json` lets a docker scope name an exact tagged or digested reference, but the scope handed to the evaluator is always `registry/repository`, so such an entry can never match. It loads, validates, and is silently inert, and evaluation falls through to the next scope or to `default`. A `reject` or a `signedBy` written for one tag does nothing.

My first attempt was to put the tag into `ImageReference.Scope`, which turned out to be wrong: `signature.SignedByVerifier` uses `Scope` as the signature store key, and `TestRepository_DefaultSignedByVerifier_EndToEnd` failed. `Scope` is left alone here. Instead the evaluator resolves requirements against the fully qualified identity, which is what `Reference` already holds, and falls back to `Scope` when the reference carries no tag or digest.

Two parts:

- `Evaluator.isImageAllowed` looks up requirements with `image.policyScope()`, which returns `Reference` when it is `Scope` plus a `:tag` or `@digest` suffix.
- `GetRequirementsForImage` drops a trailing tag or digest before the namespace passes, so a repository or namespace entry still covers a tagged image. A `:` that belongs to a registry port is kept, since the part after it still contains a `/`.

Precedence follows the containers-policy.json order: an entry for `example.com/repo:v1` wins over one for `example.com/repo`, and an entry for `example.com/repo:v1` does not apply to `:v2`.

Verified with `go test ./registry/...`. Reverting `evaluator.go` alone with the tests in place gives

    checkPolicy() on the tag named by the policy should be denied, got nil
    checkPolicy() on the digest named by the policy should be denied, got nil

and reverting `policy.go` alone breaks the repository entry still applying to a tagged image. Both new repository tests drive `checkPolicy`, the same entry point `Fetch`, `Resolve` and `Push` use. The full `./registry/...` run is green, including the signature end-to-end test that caught the first approach.

Fixes #1390


registry/remote/policy/evaluator.go
```diff
@@ -39,6 +40,21 @@ type ImageReference struct {
 	Digest digest.Digest
 }
 
+// policyScope returns the scope used to look up policy requirements. When the
+// reference carries a tag or digest, that fully qualified form is used, so a
+// containers-policy.json entry naming an exact tag or digest can match. Scope
+// itself stays the repository, which signature verifiers use as the store key.
+func (r ImageReference) policyScope() string {
```
TerryHowe, line 47: `policyScope()` is not gated on the transport, but both fallback blocks in `GetRequirementsForImage` are (`transport == TransportNameDocker`). So for a non-docker transport the tagged `Reference` is used for the exact lookup, misses, and falls straight through to the transport/global default — the scope entry is skipped. `oci` scopes are legitimately `/path/to/dir:tag`, so this is reachable.

Verified against this branch and its parent:

```go
Transports: {"oci": {"/tmp/img": Reject{}}}, Default: InsecureAcceptAnything{}
IsImageAllowed({Transport: "oci", Scope: "/tmp/img", Reference: "/tmp/img:latest"})
//   parent: allowed=false  <- reject applied
//   here:   allowed=true   <- reject bypassed
```

A `reject` that used to fire no longer does. Either gate `policyScope()` on `TransportNameDocker`, or make the trim fallback transport-independent.

registry/remote/policy/policy.go
```diff
@@ -276,6 +289,19 @@ func (p *Policy) GetRequirementsForImage(transport TransportName, scope string)
...
 // isPathPrefix reports whether prefix is a prefix of s at a "/" boundary.
 // That is, prefix matches s if s == prefix or s starts with prefix + "/".
+// trimScopeReference drops a trailing tag or digest from a docker scope,
+// leaving the registry and repository. A ":" that belongs to a registry port
+// is kept, since the part after it still contains a "/".
+func trimScopeReference(scope string) string {
+	if i := strings.LastIndex(scope, "@"); i != -1 {
+		return scope[:i]
+	}
+	if i := strings.LastIndex(scope, ":"); i != -1 && !strings.Contains(scope[i+1:], "/") {
```
TerryHowe: This strips the port from a bare `host:port` scope, which containers-policy.json allows as a docker scope. The comment above says a port colon is kept "since the part after it still contains a `/`" — only true when a repository path follows. `"example.com:5000"` → `"example.com"`.

```go
Transports: {docker: {"example.com": InsecureAcceptAnything{}}}, Default: Reject{}
GetRequirementsForImage(docker, "example.com:5000") // -> insecureAcceptAnything, want reject
```

`example.com:5000` and `example.com:443` are distinct registries, so this can widen a policy, not just narrow it. Also require a `/` before the colon:

```go
if i := strings.LastIndex(scope, ":"); i != -1 &&
    !strings.Contains(scope[i+1:], "/") && strings.Contains(scope[:i], "/") {
    return scope[:i]
}
```

oras' own `policyImageReference` always builds `registry/repository`, so internal callers are safe — but `GetRequirementsForImage` is exported and documents `scope` generically.

registry/remote/policy/evaluator.go
arpitjain099: You are right, and your repro reproduces exactly as written. I turned it into a test before changing anything:

```
--- FAIL: TestGetRequirementsForImageScopeEdgeCases/non-docker_transport_still_falls_back_from_a_tag
    oci tagged scope bypassed its reject entry, got insecureAcceptAnything, want reject
```

I took the second option and made the trim fallback transport-independent rather than gating `policyScope()`, since gating it would leave `oci` scopes like `/path/to/dir:tag` unable to match an exact entry at all, which is the thing this PR is trying to fix. The fallback now runs before the docker-only prefix and wildcard matching.

registry/remote/policy/policy.go
arpitjain099: Confirmed, and it is the worse of the two because it widens rather than narrows. Your example fails on the parent of this commit:

```
--- FAIL: TestGetRequirementsForImageScopeEdgeCases/registry_port_is_not_mistaken_for_a_tag
    host:port scope matched the bare host, got insecureAcceptAnything, want reject
```

I took your suggested condition, requiring a `/` before the colon as well as none after it. I also folded in the `repo:tag@digest` case Copilot raised in the same function: the digest is dropped first and the tag after it, so `example.com/repo:tag@sha256:...` now reaches the `example.com/repo` entry instead of stopping at `example.com/repo:tag`. Both are covered by the new table test, and the misattached `isPathPrefix` doc comment is back where it belongs.

registry/remote/policy/policy.go
```diff
@@ -276,6 +289,19 @@ func (p *Policy) GetRequirementsForImage(transport TransportName, scope string)
 
 // isPathPrefix reports whether prefix is a prefix of s at a "/" boundary.
 // That is, prefix matches s if s == prefix or s starts with prefix + "/".
+// trimScopeReference drops a trailing tag or digest from a docker scope,
+// leaving the registry and repository. A ":" that belongs to a registry port
+// is kept, since the part after it still contains a "/".
```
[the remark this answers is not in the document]
arpitjain099: Both fixed. The `isPathPrefix` doc comment is back above `isPathPrefix`, and `trimScopeReference` has its own describing the port rule. The digest is now dropped first and the tag after it, so `repo:tag@sha256:...` falls back to `repo` rather than stopping at `repo:tag`; there is a test for that case.

registry/remote/repository_policy_test.go
```diff
@@ -877,3 +877,65 @@ func TestRepository_PushReference_SignedByPolicy_Denied(t *testing.T) {
...
+		Registry: &Registry{
+			Reference: registry.Reference{Registry: testReference.Registry},
+			Policy:    evaluator,
+		},
+		RepositoryName: testReference.Repository,
+	}
+
+	if err := repo.checkPolicy(context.Background(), "blocked"); err == nil {
+		t.Error("checkPolicy() on the tag named by the policy should be denied, got nil")
+	}
```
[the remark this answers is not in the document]
arpitjain099: Done, both this one and line 935 now use `assertPolicyDenied`, so they check the error is an access-denied policy failure rather than just non-nil.

registry/remote/policy/policy.go
```diff
@@ -274,6 +286,23 @@ func (p *Policy) GetRequirementsForImage(transport TransportName, scope string)
...
+func trimScopeReference(scope string) string {
+	if i := strings.LastIndex(scope, "@"); i != -1 {
+		scope = scope[:i]
+	}
+	if i := strings.LastIndex(scope, ":"); i != -1 &&
+		!strings.Contains(scope[i+1:], "/") && strings.Contains(scope[:i], "/") {
+		return scope[:i]
+	}
+	return scope
+}
```
TerryHowe, line 305: The `@` trim is unconditional. It fires on any `@`, not just a digest — and since the fallback is now transport-independent, an `oci` scope is a filesystem path where `@` is a legal directory character. This widens policy, the same direction as the `host:port` case:

```go
Transports: {"oci": {"/mnt/data": InsecureAcceptAnything{}}}, Default: Reject{}
GetRequirementsForImage("oci", "/mnt/data@2024/img")
//   parent: reject                  <- nothing matches, global default
//   here:   insecureAcceptAnything  <- trimmed to "/mnt/data", matched an unrelated entry
```

Note it trims past a `/`, so the result is not even a parent scope of the input.

The tag branch has the same gap, milder because the `/` guard catches most of it: `example.com/repo:not a tag!` trims to `example.com/repo`. Validating both halves fixes them together — `tagRegexp` at `registry/reference.go:45` subsumes the existing "no `/` after" check:

```suggestion
// scopeTagRegexp mirrors tagRegexp in the registry package.
var scopeTagRegexp = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)

// trimScopeReference drops a trailing tag or digest from a scope, leaving the
// registry and repository. A reference may carry both, as in
// "example.com/repo:tag@sha256:...", so the digest is removed first and the
// tag after it. Each half is trimmed only when it really is a digest or a tag,
// which leaves an "oci" path such as "/mnt/data@2024/img" and a registry port
// such as "example.com:5000" alone.
func trimScopeReference(scope string) string {
	if i := strings.LastIndex(scope, "@"); i != -1 {
		if _, err := digest.Parse(scope[i+1:]); err == nil {
			scope = scope[:i]
		}
	}
	if i := strings.LastIndex(scope, ":"); i != -1 &&
		strings.Contains(scope[:i], "/") && scopeTagRegexp.MatchString(scope[i+1:]) {
		return scope[:i]
	}
	return scope
}
```

Needs `"regexp"` and `"github.com/opencontainers/go-digest"` in the imports. With that, `/mnt/data@2024/img` and `example.com/repo:not a tag!` stop matching, while `/tmp/img:latest`, `repo:tag@sha256:...` and `example.com:5000` keep their current behavior; `go test ./registry/...` stays green.

TerryHowe (commented):
Two follow-ups on the current head. The `oci` and `host:port` fixes from the last round both look right.

registry/remote/policy/policy.go
```diff
@@ -224,6 +224,18 @@ func (p *Policy) GetRequirementsForImage(transport TransportName, scope string)
 		return reqs
 	}
 
+	// A scope may name an exact tag or digest, which is the most specific form
+	// containers-policy.json allows. An entry for the repository still applies
+	// to a tagged or digested image, so fall back to it. This is deliberately
+	// not limited to the docker transport: an oci scope is legitimately
+	// "/path/to/dir:tag", and gating it would let such an entry be skipped.
+	if repoScope := trimScopeReference(scope); repoScope != scope {
```
TerryHowe: The doc comment on this exported function (lines 212-215) is now stale — it states `exact > longest-prefix > wildcard > transport default > global default` for docker and exact-match-only for everything else. This block inserts a step into both chains, and non-docker transports are no longer exact-match-only. Worth documenting, since external callers rely on that precedence.

registry/remote/policy/policy.go
TerryHowe: Scratch the regex — no need to pay for one here, and it was more machinery than the problem needs. The last `/` is enough: a tag or digest can only live in the final path component, so start the search there and take whichever of `:` or `@` comes first.

```go
// trimScopeReference drops a trailing tag or digest from a scope, leaving the
// registry and repository. A reference can only carry one in its last path
// component, so the search starts after the last "/": that keeps a registry
// port ("example.com:5000"), which has no "/" after it, and an "@" or ":"
// inside an oci filesystem path ("/mnt/data@2024/img"). The first separator
// wins, so "example.com/repo:tag@sha256:..." trims to "example.com/repo" in
// one step.
func trimScopeReference(scope string) string {
	slash := strings.LastIndex(scope, "/")
	if slash == -1 {
		return scope
	}
	if i := strings.IndexAny(scope[slash+1:], ":@"); i != -1 {
		return scope[:slash+1+i]
	}
	return scope
}
```

This subsumes both of the guards you have now. The "no `/` after the colon" check and the "something before it contains a `/`" check are both just consequences of searching after the last slash, and the two-step digest-then-tag trim collapses into one pass, since `repo:tag@sha256:...` hits the `:` first.

Applied to this branch, the whole table holds:

| scope | result | |
|---|---|---|
| `example.com/repo:tag` | `example.com/repo` | tag |
| `example.com/repo@sha256:aaa` | `example.com/repo` | digest |
| `example.com/repo:tag@sha256:aaa` | `example.com/repo` | both, one pass |
| `example.com:5000` | `example.com:5000` | bare host:port kept |
| `example.com:5000/repo:tag` | `example.com:5000/repo` | port kept, tag trimmed |
| `/mnt/data@2024/img` | `/mnt/data@2024/img` | the bug above — `@` before the last `/` is left alone |
| `/mnt/data@2024/img:v1` | `/mnt/data@2024/img` | only the last component is trimmed |
| `/tmp/img:latest` | `/tmp/img` | oci tagged |
| `example.com/a/b/c:tag` | `example.com/a/b/c` | nested repo |

`go test ./registry/remote/...` passes with it.

One trade-off worth naming: this does not validate the tag, so `example.com/repo:not a tag!` still trims to `example.com/repo`. That is the narrowing direction — a malformed scope falls back to its own repository rather than reaching an unrelated entry — so it is not the widening problem the `@` case was, and it is not worth a regex on every lookup to close.

registry/remote/policy/policy.go
TerryHowe: Concretely, this is the replacement for the doc comment on lines 212-215. I could not attach it as an applyable `suggestion` block — those lines are outside the diff hunk, which starts at 224, so GitHub will not anchor a comment there. Copy it over the four existing lines:

```go
// GetRequirementsForImage returns the policy requirements for a given transport and scope.
// A scope naming an exact tag or digest falls back to an entry for the enclosing
// repository, and the trimmed scope is what any later matching sees. This applies
// to every transport.
// For the docker transport the precedence is exact match > repository match >
// longest-prefix match > wildcard subdomain match > transport default > global default.
// For other transports it is exact match > repository match > transport default >
// global default.
```

The part most worth spelling out is the second sentence: the fallback does not just add a step, it reassigns `scope`, so the longest-prefix and wildcard passes below run against the trimmed scope rather than the one the caller passed.

registry/remote/policy/policy.go
TerryHowe, line 305: Applyable version of the fix from the thread above, replacing the whole function.

```suggestion
// trimScopeReference drops a trailing tag or digest from a scope, leaving the
// registry and repository. A reference can only carry one in its last path
// component, so the search starts after the last "/": that keeps a registry
// port ("example.com:5000"), which has no "/" after it, and an "@" or ":"
// inside an oci filesystem path ("/mnt/data@2024/img"). The first separator
// wins, so "example.com/repo:tag@sha256:..." trims to "example.com/repo" in
// one step.
func trimScopeReference(scope string) string {
	slash := strings.LastIndex(scope, "/")
	if slash == -1 {
		return scope
	}
	if i := strings.IndexAny(scope[slash+1:], ":@"); i != -1 {
		return scope[:slash+1+i]
	}
	return scope
}
```

Searching after the last `/` subsumes both existing guards — "no `/` after the colon" and "something before it contains a `/`" are just consequences of starting there — and it collapses the digest-then-tag two-step into one pass, since `example.com/repo:tag@sha256:...` hits the `:` first.

It also fixes the `oci` case that prompted the thread: `/mnt/data@2024/img` is left alone, because the `@` is not in the last path component. I ran this on the branch against a 14-case table plus `go test ./registry/remote/...`; all green.

Trade-off, stated plainly: no tag validation, so `example.com/repo:not a tag!` still trims to `example.com/repo`. That direction is narrowing — a malformed scope falls back to its own repository rather than reaching an unrelated entry — so it is not the widening bug the `@` case was, and not worth a regex on every lookup.

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks

Issue #1383, opened 2026-09-04
auth: token cache key ignores the namespace, so credentials can be shared across namespaces on one host


TerryHowe:
Split out of #1348, which is closed in favour of smaller pieces.

**Related to #1381** (which derives the value the key needs).

## Problem

`auth.Cache` is keyed `(host, scheme, scopesKey)`, and **for basic auth the scopes key is hardcoded `""`**:

```go
// registry/remote/auth/client.go:286
token, err := cache.GetToken(ctx, host, SchemeBasic, "")

// registry/remote/auth/client.go:322
token, err := cache.Set(ctx, host, SchemeBasic, "", func(ctx context.Context) (string, error) {
```

The moment credentials vary by namespace (#1382), two namespaces on one host share a cached basic token — **first one in wins for both**. A caller holding separate credentials for `example.com/team-a` and `example.com/team-b` would silently send team-a's credential to team-b.

Bearer is only *incidentally* safe, because the scopes key usually contains the repository. On a bare `/v2/` ping with no hints it collides too.

## Proposal

Whatever `Resource` the credential was resolved with has to become part of the cache key, for both schemes.

## Why this is its own issue

It is a latent correctness bug rather than a signature change, and it is easy to lose inside the larger credential-resolution work. It does nothing observable until namespaced credentials exist, but if it is missed there, the failure is silent credential cross-use rather than an error.

Ordering: it must land **with or before** #1382, not after.

pieromachacuaycabrera-cyber:
This is a critical cache invalidation gap that would definitely lead to silent cross-contamination of credentials. 

To prevent this collision ahead of the namespaced credentials release (#1382), the namespace/repository resolved from the `Resource` context must be injected directly into the `scopesKey` (or a dedicated `namespace` field in the cache key) rather than hardcoding `""`.

Here is the structural Go patch to apply to `registry/remote/auth/client.go` (and wherever the `cache` interface is implemented) to fold the namespace into the key:

```go
// Helper to derive a stable cache key namespace from the Resource
func resolveNamespaceKey(hint *auth.ChallengeHint) string {
    if hint != nil && hint.Repository != "" {
        return hint.Repository
    }
    return "" // Fallback for bare /v2/ pings
}

// ... inside client.go (around line 286 / 322) ...

// 1. Derive the namespace key
namespaceKey := resolveNamespaceKey(hint) // Or from the resolved Resource object

// 2. Pass it down instead of "" for Basic Auth
[...]

Issue #1382, opened 2026-09-04
auth: CredentialFunc and TokenParams should take a Resource instead of a host string


TerryHowe:
Split out of #1348, which is closed in favour of smaller pieces.

**Depends on #1380** (the `Resource` type) and **#1381** (deriving the path in `Client.Do`).

## Problem

Credential *lookup* is host-only, so a namespaced credential could never be read back even if something wrote one.

```go
// registry/remote/credentials/credential.go:60
type CredentialFunc func(ctx context.Context, hostport string) (Credential, error)
```

`auth.Client` calls it with just the host. The store already supports hierarchical keys — `StoreOptions.Hierarchical` (`registry/remote/credentials/store.go:95`) and `GetAuthConfigHierarchical`'s longest-prefix walk (`registry/remote/internal/configfile/configfile.go:184`) — but nothing ever passes it a path to match against.

The same gap exists on the token-fetch side: `TokenParams.Registry` is a bare `string` (`registry/remote/auth/token.go:35`), so a custom `TokenFetcher` — the ECR/GAR/ACR token-exchange case, which is the reason that extension point exists — has to reverse-engineer the repository out of `params.Scopes` strings.

## Proposal

Both signatures take the `Resource` value instead of a host string:

```go
type CredentialFunc func(ctx context.Context, res Resource) (Credential, error)

type TokenParams struct {
    Resource Resource // was: Registry string
    ...
}
```

A struct parameter also means neither signature has to break again to add a field.

## Lookup order

The hierarchical walk should follow the containers `auth.json` semantics — most-specific to least-specific. A pull of `my-registry.local/namespace/user/image:latest` looks up:

- `my-registry.local/namespace/user/image`
- `my-registry.local/namespace/user`
- `my-registry.local/namespace`
- `my-registry.local`

Spec: <https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md#format> — this is what `GetAuthConfigHierarchical` already implements, and matching it exactly keeps us compatible with podman/skopeo credential files. Prior art in go-containerregistry: <https://github.com/google/go-containerregistry/pull/1966>.

## Open question — hierarchical matching is plaintext-only

Per the `StoreOptions.Hierarchical` doc comment, native helpers and the macOS keychain are always exact-key. A namespaced credential on a default macOS setup would be stored under a key that is never matched.

Two options:

1. Document the limitation.
2. Have the namespace-aware `CredentialFunc` do the prefix walk itself, so it works over any store.

[...]

PR #1398, opened 2026-09-06
auth: derive the repository path from the request URL in Client.Do

### LINKED ISSUE #1381: auth: derive the repository path from the request URL in Client.Do

**Depends on #1380** (the `Resource` type).

## Problem

`auth.Client` satisfies `remote.Client`, whose whole surface is `Do(*http.Request)` (`registry/remote/repository.go:85`). The request is therefore the only channel from caller to auth layer. For the auth layer to resolve a credential for `example.com/myspace` rather than just `example.com`, it has to know the repository path — and today it never computes one.

## Proposal

In `Client.Do`, derive a `Resource` from the request URL: strip a leading `/v2/`, then split at the **rightmost** `/manifests/`, `/blobs/`, `/tags/list`, or `/referrers/`. What remains is the repository path.

```
/v2/myspace/app/manifests/v1        -> {Host: h, Path: "myspace/app"}
/v2/myspace/app/blobs/sha256:...    -> {Host: h, Path: "myspace/app"}
/v2/myspace/app/tags/list           -> {Host: h, Path: "myspace/app"}
/v2/                                -> {Host: h, Path: ""}
/some/other/path                    -> {Host: h, Path: ""}
```

Rightmost, because `<name>` may itself contain a segment called `blobs` or `manifests`.

## Why derive it rather than read the context hints

The original #1348 write-up proposed recovering the path from the scope hints via the context helpers. Deriving it from the URL is better:

- **Deterministic, and needs no caller cooperation.** It works for a third party who hand-builds a request and never touched `AppendScopesForHost`.
- **Hint-based recovery fails when no hint was set**, and is ambiguous when hints cover several repositories — cross-repo blob mount sets two (`registry/remote/repository.go:1179-1184`).
- **Fails in the safe direction.** Anything not under `/v2/` yields an empty path and degrades to exactly today's host-only behaviour. A registry served behind a path prefix parses as no-path, i.e. today's behaviour.

`auth.Client` is already registry-aware — it speaks the distribution token flow and ships `ScopeRegistryCatalog` — so this is not a layering violation.

## Scope of this issue

The derivation helper and its tests, plus computing the value in `Do`. **Internal only** — no exported signature changes, and nothing consumes the derived path yet. Consumers land in the follow-ups:

- credential resolution keyed on it
- the token cache key including it

srikary12:
Resolves #1381

Depends on #1391, which adds `properties.Resource`. CI stays red until that merges — every failure is `undefined: properties.Resource`.

## What

`auth.Client` satisfies `remote.Client`, whose whole surface is `Do(*http.Request)`. The request is therefore the only channel from caller to auth layer, so the repository has to be recovered from the URL rather than from caller-supplied scope hints — that way it is deterministic and works for a third party who hand-builds a request.

`requestResource` strips a leading `/v2/`, then splits at the **rightmost** `/manifests/`, `/blobs/`, `/tags/list`, or `/referrers/`. What remains is the repository path. Rightmost, because `<name>` may itself contain a segment called `blobs` or `manifests`.

| Request path | Derived path |
|---|---|
| `/v2/myspace/app/manifests/v1` | `myspace/app` |
| `/v2/myspace/app/blobs/uploads/<uuid>?_state=…` | `myspace/app` |
| `/v2/myspace/app/tags/list?n=10` | `myspace/app` |
| `/v2/myspace/blobs/manifests/v1` | `myspace/blobs` |
| `/v2/` | `""` |
| `/v2/_catalog` | `""` |
| `/some/other/path` | `""` |

Anything that is not a recognized repository endpoint yields an empty path and degrades to exactly today's host-only behaviour — a registry served behind a path prefix included.

`Client.Do` now derives the resource and takes `host` from `resource.Registry`, replacing the inline `req.Host` / `req.URL.Host` fallback. **No behaviour change:** `resource.Path` is computed but nothing consumes it yet. Credential resolution (#1382) and the token cache key (#1383) pick it up in the follow-ups.

## One addition beyond the issue text

The derived path is validated with `properties.Reference.ValidateRepository`, and a failure degrades to an empty path. This keeps a derived `Resource` under the same invariant `ParseResource` enforces, so a malformed path cannot later become a credential lookup or a cache key. Happy to drop it if you would rather keep this strictly to the issue text.

## Tests

23 table cases: every repository endpoint (blob upload sessions and tag lists with query params included), the rightmost-split cases, `Host` header precedence over the URL host, and the degrade-to-registry cases — `/v2/`, `/v2`, `/v2/_catalog`, a path-prefixed registry, non-`/v2/` paths, and empty or invalid repository names. 100% coverage of the new file.


TerryHowe (changes requested):
Build issues

srikary12:
> Build issues

Should be resolved once #1380 is merged

Ricky-7-Yan (changes requested):
Blocking: requestResource is never wired into production code at the current head. Client.Do in registry/remote/auth/client.go still derives host directly from originalReq.Host / originalReq.URL.Host, and the PR changes only resource.go plus its unit test, so every runtime request bypasses the new helper. As written, merging this leaves #1381's Client.Do scope incomplete and gives #1382/#1383 no derived Resource to consume. Please replace the inline host derivation with resource := requestResource(originalReq) and host := resource.Registry (retaining the resource in Do for the follow-ups), then add/adjust a Client.Do-level test or otherwise demonstrate that the production path invokes the derivation.

srikary12:
> Blocking: requestResource is never wired into production code at the current head. Client.Do in registry/remote/auth/client.go still derives host directly from originalReq.Host / originalReq.URL.Host, and the PR changes only resource.go plus its unit test, so every runtime request bypasses the new helper. As written, merging this leaves #1381's Client.Do scope incomplete and gives #1382/#1383 no derived Resource to consume. Please replace the inline host derivation with resource := requestResource(originalReq) and host := resource.Registry (retaining the resource in Do for the follow-ups), then add/adjust a Client.Do-level test or otherwise demonstrate that the production path invokes the derivation.

Fixed.

Ricky-7-Yan (approved):
Re-reviewed current head 752f532. The production path now calls requestResource(originalReq).Registry in Client.Do, preserving the existing Request.Host-over-URL.Host behavior while routing runtime requests through the new derivation. The helper's table tests cover registry/repository shapes and all current build, lint, CodeQL, DCO, and coverage checks pass. My earlier blocking concern is resolved.

registry/remote/auth/resource.go
```diff
@@ -0,0 +1,77 @@
...
+	"/referrers/",
+}
+
+// requestResource derives the registry resource addressed by req.
+// A URL that does not address a repository yields the whole registry.
+func requestResource(req *http.Request) properties.Resource {
+	host := req.Host
+	if host == "" {
+		host = req.URL.Host
+	}
```
TerryHowe, line 44: `Do` used to dereference `originalReq.URL` only when `Host` was empty; `requestResource` now reads `req.URL.Path` on every call. A request with a non-empty `Host` and a nil `URL` previously flowed through `Clone` into `http.Client.Do` and came back as `http: nil Request.URL` — it now nil-derefs and panics. Since the stated motivation for deriving from the request is precisely the third party who hand-builds one, worth a guard:

```suggestion
	if req.URL == nil {
		return properties.Resource{Registry: req.Host}
	}
	host := req.Host
	if host == "" {
		host = req.URL.Host
	}
```

registry/remote/auth/resource.go
```diff
@@ -0,0 +1,77 @@
...
+	return properties.Resource{
+		Registry: host,
+		Path:     repositoryFromPath(req.URL.Path),
+	}
```
TerryHowe, line 48: `properties.Resource` documents `Registry` as normalized — *"The host portion is lowercase because registry hostnames are case-insensitive"* — and `ParseResource` enforces it with `strings.ToLower` plus `ValidateRegistry`. This stores `req.Host` verbatim, so `https://Example.COM/v2/myspace/app/manifests/v1` derives `{Registry: "Example.COM", Path: "myspace/app"}`.

The PR body argues for validating the path so a derived `Resource` stays "under the same invariant `ParseResource` enforces" — I agree with that, and the same argument applies to `Registry`. It also has an effect today: `AppendRepositoryScope` keys hints on `ref.Host()` (`registry/remote/auth/scope.go:78`), which *is* normalized, while `Do` looks them up with the raw value — so a mixed-case host silently misses its scope hints.

```suggestion
	return properties.Resource{
		Registry: strings.ToLower(host),
		Path:     repositoryFromPath(req.URL.Path),
	}
```

Lowercasing is only half of it, though — `docker.io` can't be fixed this way. Request URLs are built with `ref.Host()` (`registry/remote/url.go:49`), so a derived resource always carries `registry-1.docker.io`, while `ParseResource("docker.io/foo")` yields `docker.io`: two unequal `Resource` values for the same logical target, which becomes a credential miss in #1382 and a duplicate cache entry in #1383.

Could you settle here whether a derived `Resource` is canonical, or else document it as deliberately un-normalized and make the follow-ups canonicalize before use? Either way this changes cache keys for mixed-case hosts, so it wants a test case and a line in the PR description.

TerryHowe (changes requested):
Two things in `requestResource` before the follow-ups build on the derived `Resource`. The rightmost-split logic itself looks right to me, and I verified the new `host` derivation in `Do` is byte-identical to the inline fallback it replaces, so the "no behaviour change" claim holds.

Both inline, with suggestions. The `Registry` normalization one needs a decision from you rather than just the patch — see the note on `docker.io`.

registry/remote/auth/resource.go
```diff
@@ -0,0 +1,80 @@
...
+		return properties.Resource{Registry: req.Host}
+	}
+	host := req.Host
+	if host == "" {
+		host = req.URL.Host
+	}
+	return properties.Resource{
+		Registry: strings.ToLower(host),
+		Path:     repositoryFromPath(req.URL.Path),
+	}
```
TerryHowe: Applying my two suggestions independently left them inconsistent with each other: the nil-`URL` branch returns `req.Host` verbatim, so it skips the `ToLower` the branch below just gained. `&http.Request{Host: "Example.COM"}` derives `{Registry: "Example.COM"}` — I confirmed this against the current head.

That is exactly the invariant violation line 49 exists to prevent, and it lands on the path most likely to be taken by the hand-built third-party request the guard was added for. Hoisting the normalization out of the branch fixes it and drops a return:

```suggestion
	host := req.Host
	if host == "" && req.URL != nil {
		host = req.URL.Host
	}
	resource := properties.Resource{Registry: strings.ToLower(host)}
	if req.URL != nil {
		resource.Path = repositoryFromPath(req.URL.Path)
	}
	return resource
```

While you're here: the nil-`URL` branch is also the one line Codecov flags as uncovered (95.65%, 1 missing). A case built as `&http.Request{Host: ...}` rather than through `http.NewRequest` covers it and pins the normalization above. The test table's `url`/`host` shape doesn't accommodate that, so it may be easiest as a separate `t.Run` after the loop.

registry/remote/auth/resource.go
```diff
@@ -0,0 +1,80 @@
...
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#endpoints
+var endpointSeparators = []string{
+	"/manifests/",
+	"/blobs/",
+	"/tags/list",
+	"/referrers/",
+}
+
+// requestResource derives the registry resource addressed by req.
+// A URL that does not address a repository yields the whole registry.
```
TerryHowe, line 39: Following up on my `docker.io` question rather than leaving you to guess which way I wanted it — I think the answer for *this* PR is clearly "document it, canonicalize in the follow-ups", so here is that patch.

The reason not to canonicalize here: `host` from this function is passed to `c.credential(ctx, host)` and `fetchBearerToken(ctx, host, ...)`. Mapping `registry-1.docker.io` -> `docker.io` in `requestResource` would silently move every Docker Hub credential-store lookup to a different key, which is a live behaviour change and does not belong in a PR that is otherwise a pure refactor. Better to leave the wire host alone and make #1382/#1383 canonicalize at the point where they build a credential key or a cache key, where the change is visible and testable.

So all this PR needs is for the invariant to be written down, so the follow-ups can't assume a derived `Resource` compares equal to a parsed one:

```suggestion
// requestResource derives the registry resource addressed by req.
// A URL that does not address a repository yields the whole registry.
//
// Registry is the host as it appears on the wire, lowercased, and is not
// canonicalized: a Docker Hub request carries registry-1.docker.io, while
// ParseResource("docker.io/...") yields docker.io. Callers comparing a derived
// Resource against a parsed one must canonicalize first.
```

TerryHowe (commented):
Thanks for turning both suggestions around. My earlier CHANGES_REQUESTED still stands, so this is just the concrete patches for the two things left open — no new blocking concerns. I re-derived the rightmost-split logic against the spec and I'm satisfied it's correct: a manifest/blob/referrers reference is a tag or digest and neither can contain `/`, so the last separator is always the real one. `/v2/a/manifests/manifests/v1` -> `a/manifests` and `/v2/x/tags/list/y/manifests/v1` -> `x/tags/list/y` both come out right.

Still outstanding beyond the two inline suggestions: the description's **"No behaviour change"** is no longer accurate now that `ToLower` landed. For a mixed-case host it changes the value `Do` hands to the token cache key, to `GetScopesForHost` (that one is the intended fix — it now matches `AppendRepositoryScope`'s normalized `ref.Host()`), and to `c.credential(ctx, host)` via `fetchBasicAuth`/`fetchBearerToken`. That last one is a user-visible credential-lookup change. Please add a mixed-case host case to the table and a line to the description saying so.

registry/remote/auth/resource.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	host := req.Host
+	if host == "" {
+		host = req.URL.Host
+	}
+	return properties.Resource{
+		Registry: strings.ToLower(host),
+		Path:     repositoryFromPath(req.URL.Path),
+	}
+}
```
TerryHowe, line 60: Supersedes both of Monday's suggestions — this range covers the nil-`URL` one, so applying that separately would leave this outdated. Two changes in one patch:

**Canonicalize `docker.io`.** Rationale in the review body. Pairs with `Do` calling `.Host()` (separate comment on `client.go`) — neither works alone.

**Hoist the host derivation.** Applying my two earlier suggestions independently left them inconsistent: the nil-`URL` branch returned `req.Host` verbatim, skipping the `ToLower` the other branch had just gained, so `&http.Request{Host: "Example.COM"}` derived `{Registry: "Example.COM"}`. One derivation path, normalized once, fixes that and drops the early return.

```suggestion
// requestResource derives the registry resource addressed by req.
// A URL that does not address a repository yields the whole registry.
//
// Registry is the canonical registry name, not the host on the wire: a request
// dialed at registry-1.docker.io derives docker.io, so that a derived Resource
// compares equal to ParseResource("docker.io/..."). Callers that dial, or that
// resolve credentials, must map it back with Host().
func requestResource(req *http.Request) properties.Resource {
	host := req.Host
	if host == "" && req.URL != nil {
		host = req.URL.Host
	}
	host = strings.ToLower(host)
	// Store the canonical name; Resource.Host() maps it back for dialing.
	if host == "registry-1.docker.io" {
		host = "docker.io"
	}
	resource := properties.Resource{Registry: host}
	if req.URL != nil {
		resource.Path = repositoryFromPath(req.URL.Path)
	}
	return resource
}
```

On behaviour: I verified the round-trip is exact — `registry-1.docker.io` -> `docker.io` -> `Host()` -> `registry-1.docker.io`, identity for every other host — so for requests this library builds, `Do` is unchanged. One honest caveat I owe you, having claimed parity: a hand-built request addressed literally to `docker.io` previously gave `Do` the host `docker.io` and would now give it `registry-1.docker.io`. I'd argue that's a fix rather than a regression — `ServerAddressFromRegistry` already treats the two as one registry (`registry/remote/auth.go:96-98`) — but it is a real change and belongs in the description alongside the `ToLower` one.

Coverage: the nil-`URL` path is still the line Codecov flags. It needs a request built as `&http.Request{Host: ...}` rather than through `http.NewRequest`, which the table's `url`/`host` shape can't express, so probably a small `t.Run` after the loop. Worth a `registry-1.docker.io` case too.

TerryHowe (commented):
Correcting myself on `docker.io` — my suggestion from Monday took the wrong side of it, and I've resolved that thread. Short version: I told you to document a wire-form `Registry` and push canonicalization into #1382/#1383. I now think `requestResource` should do the mapping, and `Do` should call `Host()` to get back to the wire form. Superseding suggestion below; it folds in the nil-`URL` fix too, so there's one patch to apply rather than two overlapping ones.

What I got wrong: I treated "`Do` must pass the dialed host to `CredentialFunc`" as if it constrained `Resource.Registry`. It doesn't — it constrains what `Do` reads out of it. `Resource` already separates those the way `Reference` does everywhere else in this repo: `Registry` holds the canonical name, `Host()` yields the wire form. Putting `registry-1.docker.io` in a `Registry` field is storing a `Host()` output where the canonical name belongs — the same invariant violation I objected to for the un-lowercased host, so I should have been consistent about it.

The constraint itself is real and stronger than I wrote, which is worth recording for #1382: `CredentialFunc` is documented as taking the host actually dialed (`registry/remote/credentials/credential.go:60`, the parameter is named `hostport`), and two in-tree implementations match on it strictly — `StaticCredentialFunc` maps `docker.io` onto `registry-1.docker.io` at construction and then compares for equality (`credential.go:64-73`), and `ClientBuilder.buildCredentialFunc` does `strings.EqualFold(reg, props.Reference.Host())` with a comment saying exactly why (`registry/remote/builder.go:245-252`). I checked: `StaticCredentialFunc("docker.io", cred)` called with `"docker.io"` returns `EmptyCredential`. So canonicalizing *without* the `Host()` call in `Do` would break Docker Hub basic auth outright, not merely shift a key. The two changes below go together.

registry/remote/auth/client.go
```diff
@@ -275,10 +275,7 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
 	// attempt cached auth token
 	var attemptedKey string
 	cache := c.cache()
-	host := originalReq.Host
-	if host == "" {
-		host = originalReq.URL.Host
-	}
+	host := requestResource(originalReq).Registry
```
TerryHowe: Goes with the `resource.go` suggestion — apply both or neither.

Once `Registry` is canonical, `Do` needs the wire form back, and every consumer of this variable wants the dialed host: `cache.GetScheme`/`GetToken`/`Set`, `GetScopesForHost` (hints are keyed on `ref.Host()` at `registry/remote/auth/scope.go:78`, which is the wire form), and `c.credential(ctx, host)` via `fetchBasicAuth`/`fetchBearerToken`, where `StaticCredentialFunc` and `buildCredentialFunc` both compare against `Host()` output.

```suggestion
	host := requestResource(originalReq).Host()
```

That keeps `Do` byte-identical for library-built requests while letting #1382 and #1383 key on a canonical `Resource` — which is the whole point of deriving one. The alternative I suggested on Monday, leaving `Registry` in wire form, gets the same result today but asks both follow-ups to remember to canonicalize, and silently produces a duplicate token-cache entry or a credential miss whenever one forgets.

registry/remote/auth/resource.go
TerryHowe: Withdrawing this one — I had it backwards. `Registry` should hold the canonical name and `Do` should call `Host()` to dial, rather than `Registry` holding the wire form and the follow-ups canonicalizing. Replaced by https://github.com/oras-project/oras-go/pull/1398#discussion_r4028675526 (plus a paired one on `client.go`); rationale in the review body. Please ignore the suggestion above.

registry/remote/auth/resource.go
TerryHowe: Still stands, but folded into https://github.com/oras-project/oras-go/pull/1398#discussion_r4028675526 so the two suggestions do not overlap — apply that one instead of this.

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

Issue #1401, opened 2026-09-06
Tracking: sequencing the remaining breaking changes before v3.0.0 GA


TerryHowe:
## Why this issue exists

`v3.0.0-rc.1` is tagged. **GA closes the breaking-change window**, and `v2` is GA so nothing breaking lands there either. There are currently a dozen-plus open issues that change exported signatures or observable contracts, and several of them mutate *the same symbols*. Taken in filing order, they would break the same exported functions two or three times inside one release window.

This issue is a sequencing proposal, not new scope. Every item below is an existing issue. All references verified against `main` @ `9f50ec9`.

---

## Tier 1 — the auth cluster: one break, or three

These eight issues all mutate the same small set of exported symbols in `registry/remote/auth` and `registry/remote/credentials`.

| Issue | Breaking? | Exported surface |
|---|---|---|
| #1380 `Resource` type | additive | new type + `ParseResource` |
| #1381 derive repo path in `Client.Do` | no | internal only |
| #1363 typed `Scope` | **yes** | `ScopeRepository`, `CleanScopes`, `With`/`Append`/`GetScopesForHost` — `string`/`[]string` today (`auth/scope.go:47,103,112,121,135`) |
| #1376 hints keyed on `Resource` | **yes** | the same three context helpers, *again* |
| #1385 `AppendRepositoryScope` | **yes** | `auth/scope.go:73` — still takes the deprecated `registry.Reference` |
| #1382 `CredentialFunc` / `TokenParams` | **yes** | `credentials/credential.go:60`, `auth/token.go:33` (`Registry string`) |
| #1383 token cache key ignores namespace | **yes** (silent) | `auth/client.go:286,322` hardcode `""` as the basic-auth scopes key |
| #1384 `Login` / `Logout` host-keyed | **yes** | `remote/auth.go:56` |
| #873 granular `credentials` interface | **yes** | must be defined *after* #1382, or its key type is wrong on arrival |

**The collisions, concretely:**

- #1363 and #1376 both rewrite `WithScopesForHost` / `AppendScopesForHost` / `GetScopesForHost`.
- #1385 and #1376 both rewrite `AppendRepositoryScope`. #1385 already flags this and asks for an explicit call.
- #1382 and #873 both define the credential lookup key.

Landing any of these alone guarantees a second break on the same signature before GA.

[...]

PR #1402, opened 2026-09-07
fix(remote): normalize registry host case in references and registries.conf matching

### LINKED ISSUE #1388: Registry host case is not normalized, so policy.json transport scopes can be bypassed
### What

The registry host component of a reference is never normalized to lower case, and `Policy.GetRequirementsForImage` matches scopes with case-sensitive string comparison. Changing the case of the registry host in a reference makes a `containers-policy.json` transport scope stop matching, so evaluation silently falls through to `Policy.Default`.

Because DNS is case-insensitive, the request still reaches the same registry — only the policy lookup misses.

### Where

- `registry/remote/policy/policy.go:216` — `GetRequirementsForImage` uses `==`, `strings.HasPrefix` (via `isPathPrefix`) and `strings.HasSuffix` on the raw scope string.
- `registry/remote/repository.go:513` — `policyImageReference` builds the scope as `repoRef.Registry + "/" + repoRef.Repository`, verbatim from the parsed reference.
- `registry/remote/properties/reference.go` — `NewReference` / `ValidateRegistry` accept a mixed-case host and preserve it. `url.ParseRequestURI` does not lower-case `Host`, so `ValidateRegistry`'s `uri.Host != r.Registry` check passes.

`grep -rn --include='*.go' ToLower .` shows the only host normalization in the tree is in the redirect origin checks (`registry/remote/utils.go`, `registry/remote/auth/client.go`). Reference parsing, policy scope matching and registries.conf prefix matching all skip it.

Repository names are already constrained to lower case by `repositoryRegexp`, so the exposed component is the host.

### Reproduction

```go
package policy

import "testing"

func TestScopeCaseBypass(t *testing.T) {
	p := &Policy{
		Default: PolicyRequirements{&InsecureAcceptAnything{}},
		Transports: map[TransportName]TransportScopes{
			TransportNameDocker: {
				"quay.io/secure": PolicyRequirements{&Reject{}},
			},
		},
	}
	for _, s := range []string{"quay.io/secure", "QUAY.IO/secure"} {
		reqs := p.GetRequirementsForImage(TransportNameDocker, s)
		t.Logf("scope=%-20s reqs=%v", s, reqs[0].Type())
	}
}
```

```
scope=quay.io/secure       reqs=reject
scope=QUAY.IO/secure       reqs=insecureAcceptAnything
```

### Impact

Any transport scope is bypassable by case-varying the host in the reference string: `reject` scopes stop rejecting, and `signedBy` / `sigstoreSigned` scopes stop requiring a signature, falling back to whatever `default` is. This matters most where references arrive as data (a chart, a manifest, a lockfile) rather than being typed by a person.

Scope: v3 only. `registry/remote/policy` does not exist on the `v2` branch, so no GA release is affected — `v3.0.0-rc.1` is the only tag carrying this code.

### Suggested fix

Lower-case the registry host once during parsing, in `properties.NewReference`, so every downstream consumer (policy scope, registries.conf prefix match, URL construction, auth cache key) gets a normalized value. Belt-and-braces normalization in `policyImageReference` and `matchesPrefix` would also work but leaves the next consumer to rediscover this.

Related: #1389 (same root cause in `registries.conf` prefix matching), #1390 (scope matching also ignores tagged scopes).

### LINKED ISSUE #1389: registries.conf prefix matching is case-sensitive, so a blocked registry can be reached
### What

`matchesPrefix` and `extractHost` compare registry prefixes case-sensitively, so `FindRegistry` fails to match an entry when the reference varies the case of the host. A registry marked `blocked = true` in `registries.conf` is reachable by writing the host in a different case.

### Where

- `registry/remote/config/registries.go:413` — `matchesPrefix`, all four comparisons (`==`, `HasPrefix(ref, prefix+"/")`, `HasPrefix(ref, prefix+":")`, and the wildcard `HasSuffix`) are case-sensitive.
- `registry/remote/config/registries.go:441` — `extractHost`, used for the `*.` wildcard path.
- `registry/remote/config/registries.go:477` — `IsBlocked` returns `false` whenever `FindRegistry` misses.

### Reproduction

```go
package config

import "testing"

func TestBlockedCaseBypass(t *testing.T) {
	rc := &RegistriesConfig{
		Registries: []Registry{{Prefix: "docker.io", Blocked: true}},
	}
	for _, ref := range []string{
		"docker.io/library/alpine:latest",
		"Docker.IO/library/alpine:latest",
	} {
		t.Logf("%-35s blocked=%v", ref, rc.IsBlocked(ref))
	}
}
```

```
docker.io/library/alpine:latest     blocked=true
Docker.IO/library/alpine:latest     blocked=false
```

The reference parses cleanly (`properties.NewReference` preserves the mixed-case host) and `Reference.Host()` returns it verbatim, so the request goes to the same registry over DNS.

### Impact

- `blocked = true` is bypassable — this is the one that fails open.
- The same miss also drops the entry's `insecure`, `mirrors`, and `token-flow` settings. Those fail safe (you get stricter TLS and no mirrors), but they make the behaviour confusing to debug.

Scope: v3 only. `registry/remote/config` does not exist on the `v2` branch; `v3.0.0-rc.1` is the only tag carrying it.

### Suggested fix

Same root cause as #1388 — best fixed once by lower-casing the registry host in `properties.NewReference`. If `matchesPrefix` is also hardened directly, note that both `ref` and `prefix` need folding, since the prefix comes from user-authored TOML.

7487:
Fixes #1388
Fixes #1389

The registry host is case-insensitive, but `properties.NewReference` preserved it verbatim and `config.matchesPrefix` / `RewriteReference` compared it with plain string operations. Case-varying the host in a reference bypassed `policy.json` transport scopes (fell through to `default`) and `registries.conf` `blocked = true` entries, and silently dropped the entry's `insecure` / `mirrors` / `token-flow` / `location` settings.

Changes:

- `properties.NewReference` lower-cases the registry host once, so every downstream consumer (policy scope, URL construction, auth cache key, `Repository.ParseReference` mismatch check) sees a normalized value. Repository, tag and digest are left as written. This is the Tier 3 fix proposed in #1401.
- `config.matchesPrefix` and `RewriteReference` fold the host on both the reference and the prefix, since `NewRegistryProperties` consults `registries.conf` before parsing and the prefix is user-authored TOML. Only the host component is folded: a prefix naming a tag (`docker.io/nginx:latest`) still does not match `docker.io/nginx:Latest`.

Tests: mixed-case host cases in `TestNewReference`, `TestMatchesPrefix` (including a "tag case is significant" guard), `TestRegistriesConfig_IsBlocked`, `TestRegistriesConfig_RewriteReference`, plus `TestRepository_policyImageReference_LowerCasesHost` pinning the policy scope end to end.

Note for release notes: `Reference.Registry` (and `String()`) now returns a lower-cased host for mixed-case input. v3 only; `v3.0.0-rc.1` is the only tag carrying this code.

### DIFF
--- a/registry/remote/config/registries.go
+++ b/registry/remote/config/registries.go
@@ -471,7 +471,11 @@ func extractHost(ref string) string {
 // case-sensitive repository, tag and digest untouched.
 func lowerHost(ref string) string {
 	host := extractHost(ref)
-	return strings.ToLower(host) + ref[len(host):]
+	// extractHost cannot tell a port from a tag in a slash-less reference, so
+	// it hands back "myimage:V1" whole. Fold only up to the ":" so a short-name
+	// tag survives; a port is digits, and folding it would be a no-op anyway.
+	name, _, _ := strings.Cut(host, ":")
+	return strings.ToLower(name) + ref[len(name):]
 }
 
 // ResolveAlias resolves a short name to a fully qualified reference.
--- a/registry/remote/config/registries.go
+++ b/registry/remote/config/registries.go
@@ -410,7 +410,11 @@ func (rc *RegistriesConfig) FindRegistry(ref string) *Registry {
 
 // matchesPrefix checks if the reference matches the given prefix.
 // Supports wildcard prefixes like "*.example.com".
+// The host part is compared case-insensitively on both sides, since the
+// prefix comes from user-authored TOML and the reference from the caller.
 func matchesPrefix(ref, prefix string) bool {
+	ref, prefix = lowerHost(ref), lowerHost(prefix)
+
 	// Handle wildcard prefix
 	if strings.HasPrefix(prefix, "*.") {
 		suffix := prefix[1:] // Remove the "*", keep the "."
@@ -461,6 +465,13 @@ func extractHost(ref string) string {
 	}
 
 	return ref
+}
+
+// lowerHost lower-cases the host part of a reference or prefix, leaving the
+// case-sensitive repository, tag and digest untouched.
+func lowerHost(ref string) string {
+	host := extractHost(ref)
+	return strings.ToLower(host) + ref[len(host):]
 }
 
 // ResolveAlias resolves a short name to a fully qualified reference.
@@ -514,8 +525,8 @@ func (rc *RegistriesConfig) RewriteReference(ref string) string {
 	}
 
 	// Replace prefix with location
-	if strings.HasPrefix(ref, prefix) {
-		return location + ref[len(prefix):]
+	if lref, lprefix := lowerHost(ref), lowerHost(prefix); strings.HasPrefix(lref, lprefix) {
+		return location + lref[len(lprefix):]
 	}
 
 	return ref
--- a/registry/remote/config/registries_test.go
+++ b/registry/remote/config/registries_test.go
@@ -427,6 +427,16 @@ func TestRegistriesConfig_IsBlocked(t *testing.T) {
 			ref:  "unknown.registry.com/image:tag",
 			want: false,
 		},
+		{
+			name: "blocked registry mixed-case host",
+			ref:  "Blocked.Registry.COM/image:tag",
+			want: true,
+		},
+		{
+			name: "wildcard blocked mixed-case host",
+			ref:  "Sub.Blocked.Example.COM/image:tag",
+			want: true,
+		},
 	}
 
 	for _, tt := range tests {
@@ -528,6 +538,11 @@ func TestRegistriesConfig_RewriteReference(t *testing.T) {
 			ref:  "gcr.io/myproject/image:v1",
 			want: "gcr.io/myproject/image:v1",
 		},
+		{
+			name: "rewrite mixed-case host",
+			ref:  "Docker.IO/library/alpine:Latest",
+			want: "library-mirror.example.com/alpine:Latest",
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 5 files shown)


registry/remote/config/registries.go
```diff
@@ -463,6 +467,13 @@ func extractHost(ref string) string {
 	return ref
 }
 
+// lowerHost lower-cases the host part of a reference or prefix, leaving the
+// case-sensitive repository, tag and digest untouched.
+func lowerHost(ref string) string {
+	host := extractHost(ref)
+	return strings.ToLower(host) + ref[len(host):]
```
TerryHowe: `extractHost` deliberately returns the whole string for a slash-less `host:something`, since it can't tell a port from a tag there. So `lowerHost` folds the tag too, and `RewriteReference` bakes that into its result:

```
lowerHost("myimage:V1")        = "myimage:v1"
RewriteReference("myimage:V1") = "mirror.example.com/myimage:v1"   // was ...:V1 before this PR
```

With a `[[registry]] prefix = "myimage", location = "mirror.example.com/myimage"` entry, that reaches the public API — `NewRegistryProperties("myimage:V1", rc)` yields `tag="v1"`, so a short-name pull of `V1` is issued against a different tag. It also contradicts the PR's own "tag case is significant" guarantee; the guard test only covers `docker.io/nginx:Latest`, where the `/` makes `extractHost` stop at the host.

Folding only up to the `:` fixes it — a port is digits, so nothing is lost:

```suggestion
	host := extractHost(ref)
	// extractHost cannot tell a port from a tag in a slash-less reference, so
	// it hands back "myimage:V1" whole. Fold only up to the ":" so a short-name
	// tag survives; a port is digits, and folding it would be a no-op anyway.
	name, _, _ := strings.Cut(host, ":")
	return strings.ToLower(name) + ref[len(name):]
```

I ran this against the branch: `go build ./...` plus `registry/remote/...`, `config` and `properties` all pass, including the new `LocalHost:5000` prefix and wildcard cases. Worth adding table cases for `myimage:V1` (unchanged) and `Reg.COM:5000/a/B:Tag` → `reg.com:5000/a/B:Tag` to pin both halves.

TerryHowe (approved):
/lgtm

PR #1403, opened 2026-09-07
fix(remote): validate referrers tag fallback descriptors

### LINKED ISSUE #1396: Referrers tag schema fallback trusts any manifest returned at the referrers tag
## What happened

`referrersByTagSchema` treats *whatever manifest the registry returns at the referrers tag* as a referrers index, without validating it. If a registry resolves the reference `<alg>-<hex>` to the manifest `<alg>:<hex>` — i.e. normalizes the dash form to a digest instead of treating it as a tag — the fallback fetches **the subject itself** and returns the subject's own `manifests` array as its referrers.

For a multi-arch subject this means the index's own child manifests are reported as referrers of the index.

GitLab's container registry (`registry.gitlab.com`) behaves exactly this way, and it does not implement the Referrers API, so every referrers lookup there takes the fallback path.

## Steps to reproduce

Public image, no local registry needed.

```console
$ TOK=$(curl -s "https://gitlab.com/jwt/auth?service=container_registry&scope=repository:gitlab-org/build/cng/gitlab-workhorse-ee:pull" | jq -r .token)

# Referrers API is not implemented
$ curl -s -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer $TOK" \
    "https://registry.gitlab.com/v2/gitlab-org/build/cng/gitlab-workhorse-ee/referrers/sha256:11e4068791cb99f2871ddbbeb5b3c213814a7cba4299d33acc5d31b1171064cb"
404

# ... so we fall back to the referrers tag, which resolves to the subject itself
$ curl -s -D- -o /dev/null -H "Authorization: Bearer $TOK" \
    "https://registry.gitlab.com/v2/gitlab-org/build/cng/gitlab-workhorse-ee/manifests/sha256-11e4068791cb99f2871ddbbeb5b3c213814a7cba4299d33acc5d31b1171064cb"
HTTP/2 200
content-type: application/vnd.docker.distribution.manifest.list.v2+json
docker-content-digest: sha256:11e4068791cb99f2871ddbbeb5b3c213814a7cba4299d33acc5d31b1171064cb
```

Note `docker-content-digest` is the subject's own digest. This is not a stray tag pushed into the repository — `sha256-…` does not appear in `/tags/list`, and the same request for a digest that does not exist in the repository returns 404. The registry is normalizing `sha256-<hex>` to `sha256:<hex>`.

Against `main` (`9f50ec9`):

```go
package main

import (
	"context"
	"fmt"

	"github.com/oras-project/oras-go/v3/registry"
	"github.com/oras-project/oras-go/v3/registry/remote"
)

func main() {
	ctx := context.Background()
	repo, err := remote.NewRepository("registry.gitlab.com/gitlab-org/build/cng/gitlab-workhorse-ee")
	if err != nil {
		panic(err)
	}

	desc, err := repo.Resolve(ctx, "v19.3.0")
	if err != nil {
		panic(err)
	}
	fmt.Printf("subject: %s (%s)\n", desc.Digest, desc.MediaType)

	refs, err := registry.Referrers(ctx, repo, desc, "")
	if err != nil {
		panic(err)
	}
	fmt.Printf("referrers reported: %d\n", len(refs))
	for _, r := range refs {
		fmt.Printf("  %s %s artifactType=%q\n", r.Digest, r.MediaType, r.ArtifactType)
	}
}
```

## What did you expect to happen?

`registry.Referrers` should report **0 referrers** — this image has none in the OCI sense.

## Actual output

```
subject: sha256:11e4068791cb99f2871ddbbeb5b3c213814a7cba4299d33acc5d31b1171064cb (application/vnd.docker.distribution.manifest.list.v2+json)
referrers reported: 2
  sha256:246a3ea40233f442b915d2829e3fc8e043299d4a8a1bbe0bb21b990290422125 application/vnd.docker.distribution.manifest.v2+json artifactType=""
  sha256:be904ba6967a4c114a08ad6d44e8de39fc4af8966403d368025d2badab262fcc application/vnd.docker.distribution.manifest.v2+json artifactType=""
```

Those two digests are the `linux/arm64` and `linux/amd64` children of the subject index — its *successors*, returned as its *predecessors*.

## Root cause

`referrersFromIndex` (`registry/remote/repository.go:1049`) decodes the response into an `ocispec.Index` and returns `index.Manifests` verbatim:

```go
var index ocispec.Index
if err := decodeJSON(rc, desc, &index); err != nil {
	return ocispec.Descriptor{}, nil, fmt.Errorf("failed to decode referrers index from referrers tag %s: %w", referrersTag, err)
}

return desc, index.Manifests, nil
```

`referrersByTagSchema` (`registry/remote/repository.go:1025`) then only filters by artifact type. Nothing checks that:

- the fetched content is an OCI image index (`application/vnd.oci.image.index.v1+json`), as [required by the spec](https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#backwards-compatibility) — here it is a Docker manifest list; or
- the fetched descriptor is not the subject itself.

Worth noting that `registry.Referrers` (`registry/repository.go`) *does* verify `subject` for stores that are not a `ReferrerLister`, but remote repositories take the `ReferrerLister` short-circuit, so that verification never runs on this path.

## Impact

Anything built on `registry.Referrers` against such a registry gets phantom referrers. Two concrete symptoms in the ORAS CLI:

- `oras discover` lists an index's own children as its referrers.
- `oras cp -r` **fails outright** with `MANIFEST_UNKNOWN` — see oras-project/oras#2148. The mechanism: the CLI's `prepareCopyOption` sees non-empty root referrers, so `ExtendedCopyGraph` runs with the index as the entry node; `findRoots` (`extendedcopy.go:150`) walks predecessors upward and lands on the two child manifests as the graph roots. `copyGraph` from a child never reaches the index above it, so the root index is never pushed — and the subsequent `dst.Tag(root, ref)` fetches that manifest from the destination, which 404s.

The second one is arguably the more serious shape of this: unvalidated fallback content steers which nodes an extended copy walks.

## Suggested fix

In `referrersByTagSchema`, treat the response as "no referrers" when either guard trips. Both are free — no extra requests:

1. the fetched descriptor's media type is not `ocispec.MediaTypeImageIndex`; or
2. the fetched descriptor's digest equals the subject's digest.

Optionally, verify each candidate's `subject` field the way `registry.Referrers` already does for other store types, though that costs one fetch per candidate.

A note on placement: `referrersFromIndex` has a second caller in the referrers-tag GC/merge path (`registry/remote/repository.go:2034`) that uses the returned index descriptor, so the guards are better placed in `referrersByTagSchema` than inside `referrersFromIndex`, to leave that path as a raw fetch.

## Versions affected

- `main` (v3), confirmed at `9f50ec9`
- v2, confirmed at `v2.6.2` — the code in both branches is identical here, so this needs a backport

pentaoa:
When a registry resolves a referrers fallback tag to the subject itself, ORAS reports the subject index's children as its referrers. This can make recursive copies walk the wrong graph and fail to copy the root manifest.

Validate the descriptor returned by the tag-schema lookup before listing referrers: it must be an OCI image index and its digest must differ from the subject. Invalid fallback content yields no referrers, without additional requests. The shared index-fetch helper used by referrers-tag updates is unchanged.

Fixes #1396.

Validation: `make test` passed, including the full race detector and coverage suite. Added HTTP regression cases for a valid OCI index, a Docker manifest list, and both Docker/OCI subject aliases; all three invalid cases failed before the fix. Existing referrers tests also pass.

AI assistance: implementation and tests prepared with OpenAI Codex. No live GitLab image download was required; the registry behavior is reproduced with local HTTP fixtures.


registry/remote/repository.go
```diff
@@ -1027,7 +1027,7 @@ func (r *Repository) referrersByTagSchema(ctx context.Context, desc ocispec.Desc
 	if err != nil {
 		return err
 	}
-	_, referrers, err := r.referrersFromIndex(ctx, referrersTag)
+	indexDesc, referrers, err := r.referrersFromIndex(ctx, referrersTag)
```
TerryHowe: Pass the subject through so the helper validates (see review body) — that covers `updateReferrersIndex` as well.

```suggestion
	_, referrers, err := r.referrersFromIndex(ctx, referrersTag, desc)
```

TerryHowe (commented):
Two things beyond the inline suggestions.

**1. The write path has the same bug, and there it is destructive.** `manifestStore.updateReferrersIndex` calls the same unvalidated helper at `registry/remote/repository.go:2034`. On a normalizing registry `oldIndexDesc` becomes the *subject*, so step 4 runs `s.repo.delete(ctx, *oldIndexDesc, true)` and deletes the subject manifest (unless `skipReferrersGC` is set) — and republishes the subject's children as the referrers index. `oras attach` against the registry in #1396 can GC the user's image.

**2. Validating inside the helper fixes both call sites.** Wrapping as `ErrNotFound` means both callers already do the right thing: read path returns no referrers, write path leaves `oldIndexDesc == nil` so it pushes a fresh index instead of deleting the subject.

```go
// subject is the manifest the referrers tag is derived from.
func (r *Repository) referrersFromIndex(ctx context.Context, referrersTag string, subject ocispec.Descriptor) (ocispec.Descriptor, []ocispec.Descriptor, error) {
	desc, rc, err := r.FetchReference(ctx, referrersTag)
	if err != nil {
		return ocispec.Descriptor{}, nil, err
	}
	defer rc.Close()

	// Some registries resolve the fallback tag to the subject itself. Its
	// children are not referrers, and the tag schema requires an OCI index.
	if desc.MediaType != ocispec.MediaTypeImageIndex || desc.Digest == subject.Digest {
		return ocispec.Descriptor{}, nil, fmt.Errorf("invalid referrers index from referrers tag %s: %w", referrersTag, errdef.ErrNotFound)
	}
	// ... unchanged
}
```

and at `repository.go:2034`: `s.repo.referrersFromIndex(ctx, referrersTag, subject)`.

Worth a test for the write path too (attach against a registry that normalizes the tag must not DELETE the subject digest).

registry/remote/repository.go
```diff
@@ -1036,6 +1036,12 @@ func (r *Repository) referrersByTagSchema(ctx context.Context, desc ocispec.Desc
 		return err
 	}
 
+	// Some registries resolve the fallback tag to the subject itself. Its
+	// children are not referrers, and the tag schema requires an OCI index.
+	if indexDesc.MediaType != ocispec.MediaTypeImageIndex || indexDesc.Digest == desc.Digest {
+		return nil
+	}
+
```
TerryHowe: Drop this — the check moves into `referrersFromIndex`, so the write path at `repository.go:2034` gets it too.

```suggestion
```

registry/remote/repository.go
pentaoa: Addressed in 28f6b68. Validation now runs inside referrersFromIndex before decoding, with the subject passed by both read and write callers; invalid fallback descriptors wrap ErrNotFound. Added public PushReference regression cases for OCI/Docker subjects and a non-OCI index, asserting no DELETE and no subject children in the new index. A valid-index control still preserves existing referrers and garbage-collects the old index. The invalid cases reproduced the destructive behavior before the fix. Also corrected existing mock responses to send the OCI Content-Type before the body. Full make test passes (race + coverage); helper coverage is 83.3% and updateReferrersIndex is 88.2%.

TerryHowe (approved):
/lgtm

Issue #1191, opened 2026-05-27
Reference parsing considers zero-length tags and digests to be valid


Clockwork-Muse:
When parsing references, references with tag and/or digest delimiters but _without_ contents are considered valid.
That is, the following tests fail:
```golang
func TestParse(t *testing.T) {
	testcases := []string{
		"ghcr.io/repo:@", // Reference is empty
		"ghcr.io/repo:tag@", // Reference is empty, even though tag is set!
		"ghcr.io/repo:@sha256:94a00394bc5a8ef503fb59db0a7d0ae9e1110866e8aee8ba40cd864cea69ea1a", // Reference is the digest
	}
	for i, tc := range testcases {
		t.Run(fmt.Sprintf("t%d", i), func(t *testing.T) {
			_, err := registry.ParseReference(tc)
			require.Error(t, err)
		})
	}
}
```
The third test, with the full digest, is somewhat understandable (given digests override hash).
However, note that the second case, with the tag but the empty digest, ignores the tag entirely (resulting in an empty reference).  This is likely to be surprising behavior, and in some cases could lead to silent corruption if code ends up defaulting to the `latest` tag instead.

TerryHowe:
Fixed on `main` by #1404 (`fix(properties): reject dangling reference delimiters`).

Running the repro from the description against `main` (a83d17c):

```
"ghcr.io/repo:@"      -> invalid reference: missing digest
"ghcr.io/repo:tag@"   -> invalid reference: missing digest
"ghcr.io/repo:"       -> invalid reference: missing tag
"ghcr.io/repo@"       -> invalid reference: missing digest
"ghcr.io/repo:@sha256:94a00394bc5a8ef503fb59db0a7d0ae9e1110866e8aee8ba40cd864cea69ea1a" -> nil (parses as a digest reference)
```

The shared parser now returns `errdef.ErrInvalidReference` for a dangling `:` or `@`, so both `properties.NewReference` (and the `registry.ParseReference` wrapper over it) and `properties.NewReferenceList` reject them. That covers the second case called out here, where the tag was silently dropped and the reference came back empty — the case most likely to lead to a surprising `latest` fallback.

### The remaining case

[...]

PR #1404, opened 2026-09-07
fix(properties): reject dangling reference delimiters

pentaoa:
## Summary

Reject references ending in an empty tag or digest component, including `repo:`, `repo@`, `repo:@`, and `repo:tag@`. Previously these could be accepted with an empty reference or an unintended tag fallback.

Return `errdef.ErrInvalidReference` from the shared parser so both `properties.NewReference` (and its legacy `registry.ParseReference` wrapper) and `properties.NewReferenceList` reject these inputs.

Related to #1191. This deliberately preserves `repo:@sha256:<valid digest>`: existing `registry` tests explicitly accept this form because the digest takes precedence over the tag. It does not attempt to change that compatibility behavior.

## Tests

- Added regression cases for all four dangling-delimiter inputs through both single-reference and list parsing, asserting the error identity.
- Added coverage preserving an empty tag with a valid digest.
- Confirmed the dangling-delimiter regression cases fail before the fix.
- `make test` passes, including the full race-enabled suite, coverage, and encoding checks.

AI assistance was used to investigate, implement, and test this change.

### DIFF
--- a/registry/remote/properties/reference.go
+++ b/registry/remote/properties/reference.go
@@ -100,7 +100,10 @@ func NewReference(artifact string) (Reference, error) {
 		return Reference{}, fmt.Errorf("%w: missing registry or repository", errdef.ErrInvalidReference)
 	}
 
-	repository, digestStr, tag := splitRepository(path)
+	repository, digestStr, tag, err := splitRepository(path)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

Issue #1376, opened 2026-09-04
auth: scope hints are keyed per host, so namespaces on the same registry share one hint bucket


TerryHowe:
## Summary

#1370 narrowed scope hints from *global* to *per host*, closing #1362. The remaining narrowing asked for in #1348 — from *per host* to *per registry scope* (host + namespace) — is not covered by that PR and is tracked here.

Scope hints today are **keyed at the host but valued at the repository**. `scopesForHostContextKey` is a `string` of the host (`registry/remote/auth/scope.go:82`), so there is one bucket per host, while the strings inside it are already repository-granular (`repository:namespace1/app:pull,push`, via `ScopeRepository`). Registry-side authorization is therefore already namespace-granular. What is host-level is the *bucket*.

## Why the bucket granularity matters

The bucket decides two things:

1. **Which hints ride along on a request.** The whole bucket goes into a single token request — `q.Add("scope", scope)` per entry in the distribution flow (`registry/remote/auth/token.go:117`), space-joined into the form for OAuth2 (`token.go:166`).
2. **The token cache key** — `strings.Join(scopes, " ")` (`registry/remote/auth/client.go:292` and `:340`).

For a registry that partitions by namespace — `harbor.local/namespace1` and `harbor.local/namespace2` threaded through one context — that means:

- **Cross-namespace disclosure.** A token fetch during namespace2 work also asks for namespace1's repositories. Harbor issues a token for the subset it grants and drops the rest, so it is usually not an auth failure, but namespace1's repository names are disclosed on a request that has nothing to do with them. Same shape as #1362, narrowed to within-host.
- **Cache-key churn.** Every append changes the key for every subsequent request to that host, so a token cached during namespace1 work misses once namespace2 scopes land. Extra token round trips, which is the opposite of what a hint is for.
- **Unbounded growth.** Each `AppendScopesForHost` re-runs `CleanScopes` over the whole bucket, and `DistributionTokenFetcher` puts each scope in the query string, so a long-lived context spanning many namespaces heads toward URL-length limits.

In-tree this stays small — each repository method appends to a local context that dies with the call, and the widest case is two repositories on one host (blob mount appends `repoRef` then `fromRef`, `registry/remote/repository.go:1179-1184`). It bites the caller who threads one context across namespaces, which is exactly the `oras login example.com/myspace` shape #1348 is about.

## Proposal

[...]

PR #1405, opened 2026-09-07
refactor(auth)!: take properties.Reference in AppendRepositoryScope

### LINKED ISSUE #1385: auth: AppendRepositoryScope takes the deprecated registry.Reference
`registry.Reference` is deprecated in favour of `properties.Reference` (`registry/reference.go:52`), but it is still the parameter type of an exported function in the auth package:

```go
// registry/remote/auth/scope.go:73
func AppendRepositoryScope(ctx context.Context, ref registry.Reference, actions ...string) context.Context {
	if len(actions) == 0 {
		return ctx
	}
	scope := ScopeRepository(ref.Repository, actions...)
	return AppendScopesForHost(ctx, ref.Host(), scope)
}
```

This is the only use of the deprecated type in `registry/remote/auth`.

## The swap itself is trivial

The function touches exactly two things — `ref.Repository` and `ref.Host()` — and `properties.Reference` has both (`registry/remote/properties/reference.go:286` for `Host()`). Drop-in.

## But sequence it, or this signature breaks twice

Two callers and one design constraint make this less mechanical than it looks.

**1. The callers go through a structurally-matched interface.** Both call sites (`content.go:93` and `content.go:151`) get their `ref` from `interfaces.ReferenceParser.ParseReference`, which itself returns the deprecated type (`internal/interfaces/registry.go:23`). So either this issue converts at the call site, or it lands together with the exported-surface migration that changes `ParseReference`.

**2. `Repository` and `Resource` want the same signature.** The two fields this function reads — repository path and host — are exactly the two fields of the `Resource` type proposed in #1380, and #1376 proposes keying scope hints on `Resource` rather than on a host string. If `AppendRepositoryScope` is migrated to `properties.Reference` now and #1376 lands later, this exported signature breaks twice in one release window.

## Recommendation

Decide between:

- **(a)** Do the swap here, standalone, accepting a second break later if #1376 lands.
- **(b)** Fold this into #1376, so the signature changes once, straight from `registry.Reference` to whatever key type that issue settles on.

(b) is cheaper for downstream users, but leaves a deprecated type referenced in the auth package until #1376 is scheduled. Worth an explicit call rather than letting whoever picks up #1376 first decide by accident.

## Related

- #1376 — keying scope hints on a `Resource`; overlaps this signature directly.
- #1380 — the `Resource` type.
- #1386 / #1387 — the rest of the `registry.Reference` migration (internal and exported).

7487:
## What

- `auth.AppendRepositoryScope` now takes `properties.Reference` instead of the deprecated `registry.Reference`. It only reads `Repository` and `Host()`, which `properties.Reference` provides, so the body is unchanged.
- Remove the `appendRepositoryScope` adapter in `registry/remote/repository.go`, which existed only to convert back to the legacy type; its 17 call sites call `auth.AppendRepositoryScope` directly.
- With #1393 merged, `parseReferenceForScope` in `content.go` returns `properties.Reference` (converting only in the `LegacyReferenceParser` branch), so `oras.Tag` / `oras.TagN` pass it straight through with no call-site conversion.
- Migrate `TestAppendRepositoryScope` to `properties.NewReference` and add a row to the migration guide.

## Why

This is the only remaining use of the deprecated type in `registry/remote/auth`. This takes option (a) from the issue. On the sequencing concern with #1376: the parameter stays a *repository* reference, which is still the right input for a repository scope hint after the hint bucket is keyed on `Resource`. #1376 changes which helper this function calls internally (`AppendScopesForHost` -> a resource-keyed one), not what callers pass in, so this signature should not need to break a second time. If maintainers would rather fold it into #1376 per #1401, happy to close this.

Fixes #1385

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -117,6 +117,7 @@ function type explicit.
 | `auth.AppendScopes(ctx, scopes...)` | `auth.AppendScopesForHost(ctx, host, scopes...)` |
 | `auth.GetScopes(ctx)` | `auth.GetScopesForHost(ctx, host)` |
 | `auth.GetAllScopesForHost(ctx, host)` | `auth.GetScopesForHost(ctx, host)` |
+| `auth.AppendRepositoryScope(ctx, registry.Reference, actions...)` | `auth.AppendRepositoryScope(ctx, properties.Reference, actions...)` |
 
 Scope hints are host-specific in v3. Pass the host of the registry request,
 normally from `properties.Reference.Host()`, to the replacement functions. The
--- a/content.go
+++ b/content.go
@@ -32,6 +32,7 @@ import (
 	"github.com/oras-project/oras-go/v3/internal/syncutil"
 	"github.com/oras-project/oras-go/v3/registry"
 	"github.com/oras-project/oras-go/v3/registry/remote/auth"
+	"github.com/oras-project/oras-go/v3/registry/remote/properties"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


TerryHowe (approved):
/lgtm

7487:
Merged `main` to resolve the conflict with #1393. Since that PR migrated `ReferenceParser` to `properties.Reference`, `parseReferenceForScope` now returns `properties.Reference` and both `oras.Tag`/`oras.TagN` callers pass it straight to `auth.AppendRepositoryScope`; the `appendRepositoryScope` adapter in `registry/remote` is dropped as before. `go test ./...` passes locally.

TerryHowe:
Thanks!

PR #1406, opened 2026-09-07
refactor(remote)!: remove WithPolicyEnforcement middleware

### LINKED ISSUE #1374: Align WithPolicyEnforcement with Repository policy phases
## Problem

PR #1369 adds two-phase policy evaluation to Repository: reference-level requirements run before network I/O, while digest-dependent signature requirements run once a manifest digest is known. The separate WithPolicyEnforcement middleware still has its own operation routing and passes descriptor digests as reference strings, so the two policy entry points can drift and enforce different behavior for the same operation.

This already surfaced during #1369: middleware descriptor-based Fetch verified signatures while Repository.Fetch initially did not. Closed #1265 documents earlier middleware-specific operation semantics, but there is no open tracker for keeping both paths aligned.

## Proposed follow-up

- Share policy phase/routing logic where practical, or explicitly document intentional differences.
- Add parity tests for manifest descriptor operations, tag and digest references, blob operations, and push/Exists/Delete behavior.
- Preserve recursion protection when signature verification fetches artifacts through the same repository.

This is intentionally separate from #1369 so that bug fix stays focused.

7487:
Fixes #1374
Refs #1401 (Tier 5)

### What

Removes `WithPolicyEnforcement` and its `policyEnforcingRepository` / blob / manifest wrappers from `registry/remote`. `RepositoryMiddleware` and `Compose` stay.

### Why

This follows the evaluation on #1374 and the Tier 5 item in #1401: policy enforcement is wired through `Registry.Policy` (`ClientBuilder.PolicyEvaluator`, `builder.go:308`), which carries the two-phase evaluation from #1369. `WithPolicyEnforcement` is a second, independent routing of the same policy with no non-test callers, and it disagrees with `Repository` in ways that matter. Reproduced on current `main` with a recording `SignedByVerifier` behind the middleware:

```
descriptor Fetch   -> verifier saw Reference="sha256:0000…0001" Digest=""   (bare digest; DefaultSignedByVerifier.parseImageDigest rejects it)
Resolve("v1")      -> verifier saw Reference="v1"               Digest=""   (unqualified, no digest; the bug #1369 fixed in Repository)
Blobs().Fetch      -> signedBy evaluated on a blob               (Repository is reference-level only for blobs)
denying signedBy   -> Delete, Blobs().Resolve, Tags all returned nil        (fail-open carve-outs Repository does not have)
wrapped.(registry.Mounter) -> false                              (copy.go:402 assertion fails; cross-repo mount silently disabled)
```

It cannot be made correct in place: it operates on the `registry.Repository` interface, which cannot parse references against the repository or classify manifest media types, and both are required for the phase split. Keeping it means teaching the interface things only `*remote.Repository` knows, plus a parity test matrix, for an API nothing in tree uses. Removing it before v3 GA closes a fail-open enforcement path while removal is still cheap.

### Breaking change

`remote.WithPolicyEnforcement` is removed (v3 pre-release only; it never shipped in v2). Callers should set the evaluator on the registry instead:

```go
builder := remote.NewClientBuilder()
builder.PolicyEvaluator = evaluator
repo, err := remote.NewRepositoryWithProperties(props, builder)
// or on an existing repository: repo.Registry.Policy = evaluator
```

If the owners would rather keep the middleware, the alternative is the alignment work laid out in the #1374 evaluation (shared phase checker, `Mounter` passthrough, recursion guard, parity tests); happy to do that instead.

### Testing

- `go build ./...`, `go vet ./registry/remote/`
- `go test -race ./registry/remote/` (`TestCompose*` retained)


TerryHowe (approved):
/lgtm

Issue #1395, opened 2026-09-05
Performance audit: 10 findings in the copy engine, content stores, and registry client


TerryHowe:
## Summary

I ran a performance audit over the hot paths in `main` — the copy engine, the local content stores, and the registry client. Ten findings below. The top five are **measured** on this checkout with throwaway benchmarks (written, run, then deleted — no working-tree changes); findings 6–10 are **by inspection**, reasoning stated but unmeasured, and labeled as such.

The library is structurally sound. The DAG traversal, the dedup tracker, and the metadata caching proxy are all the right shapes. What it loses is concentrated in four places: a buffer optimization that never takes effect, a concurrency default tuned for a different workload, an allocation pattern that triples memory on every manifest read, and an OCI index that gets rewritten on every push.

**Environment:** `main` @ 9f50ec9, go1.26.2 darwin/arm64, Apple M1, APFS. Numbers are single-machine and single-run — treat ratios as sound and absolute times as indicative.

Happy to split any of these into separate issues and send PRs. Findings 1–4 are all small, contained changes.

---

## 1. The pooled 1 MiB copy buffer is never used — every copy runs in 32 KiB chunks

**High · measured**
`content/oci/storage.go:35,155` · `content/file/file.go:42,472,514,561` · `content/file/utils.go:107,365` · `internal/ioutil/io.go:38`

Both content stores keep a `sync.Pool` of 1 MiB buffers, with a comment explaining the choice:

> the buffer size should be larger than or equal to 128 KiB for performance considerations. we choose 1 MiB here so there will be less disk I/O.

Every call site then hands that buffer to `io.CopyBuffer` with an `*os.File` as the destination. `io.CopyBuffer` discards the caller's buffer whenever `dst` implements `io.ReaderFrom` or `src` implements `io.WriterTo` — and `*os.File` implements both. Control lands in `genericReadFrom`, which allocates its own 32 KiB buffer per call.

All four call sites in the library hit this: OCI blob ingest, file-store save, tar creation, and tar extraction. The pool is allocated, populated, and returned to — pure overhead, zero effect.

Direct probe, 4 MiB copy with a 1 MiB buffer supplied:

```
dst = *os.File             (library today)     reads=129   maxChunk=  32 KiB
dst = struct{io.Writer}{f} (ReaderFrom hidden) reads=  5   maxChunk=1024 KiB
```

[...]

PR #1418, opened 2026-09-10
perf(content): actually use the pooled 1 MiB copy buffer

jadhavgaurav:
## What

`content/oci` and `content/file` each keep a `sync.Pool` of 1 MiB buffers, with a comment explaining that a large buffer means less disk I/O. The buffer never reached a copy loop.

`io.CopyBuffer` drops the caller's buffer whenever `src` implements `io.WriterTo` or `dst` implements `io.ReaderFrom`. Every call site hands it an `*os.File` (both interfaces) or a `*tar.Reader` (`io.WriterTo`), so control went to the generic fallback, which allocates its own 32 KiB buffer per call. The pool was populated and returned to for nothing, and the real buffer was fresh garbage on every copy. Tar extraction is where it adds up: `writeFile` runs once per archive entry, so an archive with 10,000 files allocated 10,000 buffers and issued 32 times the intended number of writes.

## Change

`internal/ioutil.CopyWithBuffer` hides both sides behind plain `io.Reader` and `io.Writer` wrappers, the same technique `net/http` uses, so the supplied buffer is the one used. Four call sites now go through it:

- `internal/ioutil.CopyBuffer`, which covers OCI blob ingest (`content/oci/storage.go`) and file store save (`content/file/file.go`)
- `tarDirectory` and `writeFile` in `content/file/utils.go`

Wrapping only the destination is not enough at the two direct sites: `tarDirectory` writes to an `io.MultiWriter` and the fast path there is `*os.File.WriteTo` on the source, and `writeFile` reads from a `*tar.Reader`, which is also an `io.WriterTo`.

No zero-copy path is lost in the verified path: content has to stream through the digest verifier, so `ReadFrom` could never splice there anyway.

One behavior note: extracting a sparse tar entry now writes its zero bytes rather than taking the seek-based path in `archive/tar`. File content is identical, holes are not preserved. Tarballs produced by this package contain no sparse entries.

## Tests

`internal/ioutil/io_test.go` gains three cases that record the largest read or write a copy issues:

- destination implements `io.ReaderFrom` (`*os.File`)
- source implements `io.WriterTo` (`*os.File`)
- the verified `CopyBuffer` path

With a 128 KiB buffer and 512 KiB of content, all three report 32768 before the change and 131072 after. Verified by reverting only the body of `CopyWithBuffer` to a plain `io.CopyBuffer` call and rerunning.

## Checks

`make test` (vendor, encoding check, full suite with `-race` and coverage) passes: all packages ok, no failures. `go build ./...`, `go vet ./...` and `gofmt -l` are clean. `golangci-lint` is not installed locally; the config enables `govet`, `ineffassign` and the `gofmt` formatter, and the first and third were run directly.

## Scope

This is finding 1 of https://github.com/oras-project/oras-go/issues/1395. The other findings are left for separate pull requests, so this issue is not closed here.

### DIFF
--- a/content/file/utils.go
+++ b/content/file/utils.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/opencontainers/go-digest"
+	"github.com/oras-project/oras-go/v3/internal/ioutil"
 )
 
 // tarDirectory walks the directory specified by path, and tar those files with a new
@@ -104,7 +105,7 @@ func tarDirectory(ctx context.Context, root, prefix string, w io.Writer, modTime
 				}
 			}()
 
-			if _, err := io.CopyBuffer(tw, fp, buf); err != nil {
+			if _, err := ioutil.CopyWithBuffer(tw, fp, buf); err != nil {
 				return fmt.Errorf("failed to copy to %s: %w", path, err)
 			}
 		}
@@ -362,6 +363,6 @@ func writeFile(path string, r io.Reader, perm os.FileMode, buf []byte) (err erro
 		}
 	}()
 
-	_, err = io.CopyBuffer(file, r, buf)
+	_, err = ioutil.CopyWithBuffer(file, r, buf)
 	return err
 }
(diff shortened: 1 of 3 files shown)


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1422, opened 2026-09-12
ci: trigger v2 releases from release labels

### LINKED ISSUE #1420: ci: trigger v2 releases from a release label instead of the branch name
## Problem

The v2 release workflow is keyed on the head branch name:

```yaml
if: github.event.pull_request.merged == true && startsWith(github.head_ref, 'release/') && ...
```

and it derives the version to tag by stripping the `release/` prefix off that branch name.

That made sense when `release/vX.Y.Z` branches were maintained, but they no longer are: `v2` is always the latest v2, and a release simply tags its tip. Keying the release on a branch-name convention now has a few drawbacks:

- The branch name carries release semantics, so a contributor can trigger (or accidentally miss) a release purely by how they named a branch.
- It conflicts with the repo's conventional branch prefixes (`feat/`, `fix/`, `chore/`, ...).
- The version is interpolated straight from `github.head_ref` into the shell script, and a branch that doesn't parse silently produces a garbage version rather than failing.

## Requested change

Trigger the v2 release from a `release` label on the PR, and read the version from the PR title:

- Gate the job on `contains(github.event.pull_request.labels.*.name, 'release')` instead of the branch prefix (keeping the existing same-repo check that ignores forks).
- Parse the version out of the PR title (`release: v2.7.0`), passed via `env:` rather than interpolated into the script body.
- Fail loudly when a PR carries the `release` label but its title does not match `release: vX.Y.Z`, so a mislabelled PR never tags an unintended version.
- Update `RELEASES.md` for the label + title convention, and call out that the release branch must live on `oras-project/oras-go` — the workflow ignores PRs from forks, so a fork branch merges without tagging anything.
- Update the `act` fixture in `.github/act/release-event.json` to carry the label and the new title.

Nothing about tagging, GoReleaser, or the release body needs to change: the PR description stays the release notes verbatim.

The `release` label already exists in this repo ("Release proposal: merging tags and publishes").

## Implementation

Proposed in #1399.

Ricky-7-Yan:
## Summary

- gate the v2 release job on the existing `release` label while retaining the same-repository protection
- extract the release version from a `release: vX.Y.Z` PR title through an environment variable and fail before tagging when the title or semantic version is invalid
- document the label/title convention and same-repository branch requirement, and update the local `act` event fixture

## Validation

- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/release.yml`
- parsed `.github/workflows/release.yml` with PyYAML
- parsed `.github/act/release-event.json` with Python's `json` module
- exercised the Bash title/version parser with stable and `rc` versions plus invalid prefix, leading-zero, incomplete-version, trailing-text, and case variants
- `git diff --check`

`act` and GoReleaser are not installed in this environment, so I did not run the containerized release dry-run. No tag or release was created.

Fixes #1420

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,10 @@ on:
 
 jobs:
   release:
-    if: github.event.pull_request.merged == true && startsWith(github.head_ref, 'release/') && github.event.pull_request.head.repo.full_name == github.repository
+    if: >-
+      github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'release') &&
+      github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     permissions:
       contents: write
@@ -41,12 +44,17 @@ jobs:
           go-version-file: go.mod
           check-latest: true
 
-      - name: Extract version from branch name
+      - name: Extract version from PR title
         id: version
+        env:
+          TITLE: ${{ github.event.pull_request.title }}
         run: |
-          BRANCH="${{ github.head_ref }}"
-          VERSION="${BRANCH#release/}"
-          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          if [[ ! "$TITLE" =~ ^release:[[:space:]]+(v[^[:space:]]+)[[:space:]]*$ ]]; then
+            echo "Error: PR is labelled 'release' but its title is not 'release: vX.Y.Z'"
+            echo "Title was: $TITLE"
+            exit 1
+          fi
+          echo "version=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
 
       - name: Validate semantic version
         run: |
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,20 +1,25 @@
 # Releasing oras-go
 
-Releases are created via a GitOps workflow. Merging a `release/vX.Y.Z` branch
-into `v2` automatically tags the commit and publishes the GitHub Release.
+Releases are created via a GitOps workflow. Merging a pull request with the
+`release` label into `v2` automatically tags the commit and publishes the
+GitHub Release. The pull request title supplies the version to tag.
 
 ## Steps
 
 ### 1. Create a release branch
 
 The release branch needs at least one commit so GitHub will allow a PR to be
-opened. Use an empty commit as a lightweight marker:
+opened. It must live in `oras-project/oras-go`: the release workflow ignores
+pull requests from forks, so merging a fork branch will not publish a release.
+
+Use an empty commit as a lightweight marker. The branch may follow the usual
+repository naming conventions because its name does not control the release:
 
 ```bash
 git fetch upstream
-git checkout -b release/v2.7.0 upstream/v2
+git checkout -b chore/release-v2.7.0 upstream/v2
 git commit --allow-empty -s -m "chore: prepare release v2.7.0"
-git push origin release/v2.7.0
+git push upstream chore/release-v2.7.0
 ```
 
 The release does not need to contain the changes being released — those are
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -41,12 +44,17 @@ jobs:
...
-      - name: Extract version from branch name
+      - name: Extract version from PR title
         id: version
+        env:
+          TITLE: ${{ github.event.pull_request.title }}
         run: |
-          BRANCH="${{ github.head_ref }}"
-          VERSION="${BRANCH#release/}"
-          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          if [[ ! "$TITLE" =~ ^release:[[:space:]]+(v[^[:space:]]+)[[:space:]]*$ ]]; then
```
TerryHowe:
```suggestion
          if [[ ! "$TITLE" =~ ^release:[[:space:]]+(v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(alpha|beta|rc)\.[0-9]+)?)[[:space:]]*$ ]]; then
```

.github/workflows/release.yml
```diff
@@ -41,12 +44,17 @@ jobs:
...
+            echo "Error: PR is labelled 'release' but its title is not 'release: vX.Y.Z'"
+            echo "Title was: $TITLE"
+            exit 1
+          fi
+          echo "version=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
 
       - name: Validate semantic version
```
TerryHowe, line 59: With the stricter extraction regex suggested [above](https://github.com/oras-project/oras-go/pull/1422#discussion_r4042517555), this step becomes dead code. The capture group there is anchored on both sides, so `BASH_REMATCH[1]` can only ever hold a string matching

```
v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(alpha|beta|rc)\.[0-9]+)?
```

which is character-for-character the pattern on the line below. The `if !` can never be true, so `exit 1` is unreachable and the step always passes.

Worth deleting rather than leaving inert, because `VERSION="${{ steps.version.outputs.version }}"` is itself the injection site: the runner substitutes the raw text before bash parses the line, so a `$(...)` payload in the PR title executes at assignment, one line before the regex that was meant to reject it. Tightening extraction defuses that, but removing the step eliminates the interpolation entirely instead of leaving it to depend on an upstream guard staying strict.

Suggest dropping lines 58-66 - the blank line plus this whole step - leaving version extraction to flow straight into "Create and push tag".

(Posted as a comment rather than a suggestion: the step body falls outside this file's diff hunk, so GitHub won't anchor an applicable suggestion across it.)

RELEASES.md
```diff
@@ -24,8 +29,9 @@ all prior work on the branch.
 
 ### 2. Open a pull request
 
-Open a PR from `release/v2.7.0` targeting the `v2` branch. Write the release
-notes directly in the PR description using the format from prior releases:
+Open a PR titled `release: v2.7.0` targeting the `v2` branch and add the
+`release` label. Write the release notes directly in the PR description using
+the format from prior releases:
```
TerryHowe: Worth spelling out that the label has to go on *before* the merge.

The job triggers on `pull_request: types: [closed]`, and a webhook payload is a snapshot taken at event time. Merge first and label second, and the `closed` event has already fired with an empty `labels` array - the `if` on `release.yml:26` evaluates false and nothing runs. Adding the label afterwards fires a `labeled` event, which this workflow doesn't subscribe to.

The failure is silent by construction: a job that never triggers can't report anything, so there's no red X to catch it. The PR just merges green and no tag appears. And a merged PR can't be reopened and re-merged, so there's no retry - recovery is a whole new empty-commit release PR.

This is a genuine change in failure mode from what it replaces. The `release/vX.Y.Z` branch name was fixed at branch creation and couldn't be forgotten halfway through; a label is a separate manual action at a separate moment. Nothing automates it today either - `.github/workflows/` on `main` has no proposer job that would apply the label for you.

```suggestion
Open a PR titled `release: v2.7.0` targeting the `v2` branch and add the
`release` label **before merging** - the release workflow triggers on the pull
request `closed` event, and that payload snapshots the labels as they were at
merge time, so labelling afterwards does nothing.

A release PR merged without the label tags nothing and cannot be re-triggered:
recovery is a fresh release PR.

Write the release notes directly in the PR description using the format from
prior releases:
```

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1423, opened 2026-09-14
auth: CredentialFunc and TokenParams should take a Resource instead of a host string

### LINKED ISSUE #1382: auth: CredentialFunc and TokenParams should take a Resource instead of a host string

srikary12:
Fixes #1382


srikary12:
@TerryHowe If this is good. I'll start working on #1383

registry/remote/auth.go
```diff
@@ -74,16 +76,19 @@ func Logout(ctx context.Context, store credentials.Store, registryName string) e
...
+	return func(ctx context.Context, res properties.Resource) (credentials.Credential, error) {
+		lookup := ServerAddressFromHostname(res.Host())
+		if lookup == "" {
 			return credentials.EmptyCredential, nil
 		}
-		return store.Get(ctx, hostport)
+		if res.Path != "" {
+			lookup = strings.TrimSuffix(lookup, "/") + "/" + res.Path
+		}
+		return store.Get(ctx, lookup)
```
TerryHowe: `host/path` never matches a non-hierarchical store (the default), and `https://index.docker.io/v1/` is a store key, not a path root — both return empty today.

```suggestion
		host := ServerAddressFromHostname(res.Host())
		if host == "" {
			return credentials.EmptyCredential, nil
		}
		// ServerAddressFromHostname may map to a sentinel server address
		// (docker.io), which is a store key rather than a path root.
		if res.Path != "" && host == res.Host() {
			// Most-specific to least-specific, so namespaced credentials
			// resolve over any store, not only the hierarchical file store.
			for path := res.Path; path != ""; {
				cred, err := store.Get(ctx, host+"/"+path)
				if err != nil {
					return credentials.EmptyCredential, err
				}
				if cred != credentials.EmptyCredential {
					return cred, nil
				}
				index := strings.LastIndex(path, "/")
				if index < 0 {
					break
				}
				path = path[:index]
			}
		}
		return store.Get(ctx, host)
```

registry/remote/builder.go
```diff
@@ -247,15 +247,15 @@ func (b *ClientBuilder) buildCredentialFunc(props *properties.Registry) credenti
...
 		// Credential specified in properties, for its own registry only.
-		if props.Credential != credentials.EmptyCredential && strings.EqualFold(reg, host) {
+		if props.Credential != credentials.EmptyCredential && strings.EqualFold(resource.Host(), host) {
 			return props.Credential, nil
 		}
 
 		// Fall back to credential store if available
 		if b.CredentialStore != nil {
-			cred, err := b.CredentialStore.Get(ctx, reg)
+			cred, err := b.CredentialStore.Get(ctx, resource.String())
```
TerryHowe:
```suggestion
			cred, err := NewCredentialFunc(b.CredentialStore)(ctx, resource)
```

registry/remote/credentials/credential.go
```diff
@@ -67,8 +42,8 @@ func StaticCredentialFunc(registry string, cred Credential) CredentialFunc {
 		// reference: https://github.com/moby/moby/blob/v24.0.0-beta.2/registry/config.go#L25-L48
 		registry = "registry-1.docker.io"
 	}
-	return func(_ context.Context, hostport string) (Credential, error) {
-		if hostport == registry {
+	return func(_ context.Context, resource properties.Resource) (Credential, error) {
+		if resource.Registry == registry {
```
TerryHowe: Needed once main is merged: `requestResource` there canonicalizes to `docker.io`, which never equals `registry-1.docker.io`.

```suggestion
		if resource.Host() == registry {
```

registry/remote/credentials/credential_test.go
```diff
@@ -88,7 +94,7 @@ func TestStaticCredential_DockerIORedirect(t *testing.T) {
 	}
 
 	// Test that docker.io itself doesn't match (because it gets redirected)
-	cred, err = credFunc(ctx, "docker.io")
+	cred, err = credFunc(ctx, testResource("docker.io"))
```
TerryHowe, line 98: Assertion below needs to flip to `expectedCred` — `docker.io` is the canonical form the client now passes.

registry/remote/credentials/credential.go
```diff
@@ -18,46 +18,21 @@ package credentials
...
-func (c *Credential) IsEmpty() bool {
-	if *c == EmptyCredential {
-		return true
-	}
-	return false
-}
-
 // CredentialFunc represents a function that resolves the credential for the
-// given registry (i.e. host:port).
 //
```
TerryHowe, line 33:
```suggestion
// CredentialFunc represents a function that resolves the credential for the
// given resource.
//
```

registry/remote/auth/token.go
```diff
@@ -27,11 +27,16 @@ import (
...
 
 // TokenParams contains parameters for token acquisition.
 type TokenParams struct {
-	// Registry is the registry hostname (i.e. host:port).
+	// Resource is the registry resource being accessed, including any namespace
+	// or repository path when available.
+	Resource properties.Resource
+	// Registry is kept as a backwards-compatible alias for the host portion of
+	// the resource. Token fetchers should prefer Resource.
 	Registry string
```
TerryHowe: Nothing reads it. `auth/token_test.go` sets it in 9 places.

```suggestion
```

registry/remote/auth/client.go
```diff
@@ -416,7 +415,8 @@ func (c *Client) fetchBearerToken(ctx context.Context, registry, realm, service
 		fetcher = NewCompositeTokenFetcher(c.Client, c.Header, c.ClientID, false)
 	}
 	params := TokenParams{
-		Registry: registry,
+		Resource: resource,
+		Registry: resource.Registry,
```
TerryHowe:
```suggestion
```

registry/remote/auth/client.go
```diff
@@ -320,7 +319,7 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
 		resp.Body.Close()
 
 		token, err := cache.Set(ctx, host, SchemeBasic, "", func(ctx context.Context) (string, error) {
```
TerryHowe: #1383: cache key is still host-only, so this PR makes namespaced credentials cross-use here. Verified two namespaces with distinct creds — the second request reuses the first's token.

TerryHowe (changes requested):
Conflicts with main after #1398: `auth/resource.go` (add/add) and `auth/client.go`. Take main's `resource.go`; in `client.go` keep the bound resource and use `host := resource.Host()`.

TerryHowe:
I think if you can increase the code coverage to get CI to pass, we are good to go. Thanks!

TerryHowe (approved):
/lgtm

PR #1425, opened 2026-09-17
fix(remote): accept alternate manifest pull digest algorithms

### LINKED ISSUE #949: Handle mismatched algorithm between `Docker-Content-Digest` and requested digest
The [`distribution-spec v1.1.1`](https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#pull) defines the behavior of the `Docker-Content-Digest` header for pulling manifests as follows:

> The `Docker-Content-Digest` header, if present on the response, returns the canonical digest of the uploaded blob which MAY differ from the provided digest.
If the digest does differ, it MAY be the case that the hashing algorithms used do not match.
See [Content Digests](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests) <sup>[apdx-3](#appendix)</sup> for information on how to detect the hashing algorithm in use.
Most clients MAY ignore the value, but if it is used, the client MUST verify the value matches the returned manifest.
If the `<reference>` part of a manifest request is a digest, clients SHOULD verify the returned manifest matches this digest.

This means that a registry may accept a `HEAD` or `GET` request with a digest using `sha512`, yet return a `Docker-Content-Digest` header computed with canonical digest algorithm `sha256`. Currently, oras-go v2 does not support this scenario and fails validation if the algorithms do not match.

We see two potential solutions:

1. If the algorithm used in the `Docker-Content-Digest` header differs from the one in the request, simply ignore the header for digest verification.
2. If the response includes content, validate it using the algorithm specified in the `Docker-Content-Digest` header. This approach introduces extra compute cost in addition to the validation between the content and the requested digest.

LarryHu0217:
Fixes #949.

Allow manifest pulls when Docker-Content-Digest uses a different
algorithm from the requested digest. Preserve the requested digest
and verify GET response content against it.

Push validation remains unchanged.

Tests cover Resolve, Fetch, FetchReference, chunked responses,
corrupted content, malformed headers, and digest mismatches.

Validation:

- Full race test suite passed
- go vet ./registry/remote passed
- New verification helper coverage: 90.9%


TerryHowe (commented):
Reviewed against a local checkout; ran the suite plus some throwaway probes. One crash-level issue and a few scope/consistency points below.

registry/remote/repository.go
```diff
@@ -2160,6 +2173,26 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
 	}, nil
 }
 
+// verifyPullContentDigest permits a different canonical digest algorithm on
+// manifest pulls, but verifies the returned bytes against the requested digest.
+// Pushes deliberately continue to use verifyContentDigest's strict comparison.
+func (s *manifestStore) verifyPullContentDigest(resp *http.Response, expected digest.Digest) error {
+	canonical, err := digest.Parse(resp.Header.Get(headerDockerContentDigest))
+	if err != nil || canonical.Algorithm() == expected.Algorithm() {
```
TerryHowe, line 2181: `expected.Algorithm().FromBytes(body)` below panics when the algorithm isn't linked in (`go-digest`'s `Algorithm.Hash()` panics). `manifestStore.Fetch` passes `target.Digest` through without `digest.Parse`, and descriptors routinely come from registry-supplied index JSON, so a remote unknown-algorithm digest crashes where `main` returned an error:

```
pr-1425: panic: sha1 not available (make sure it is imported)
main:    invalid response; digest mismatch in Docker-Content-Digest: ...
```

```suggestion
	if err != nil || !expected.Algorithm().Available() || canonical.Algorithm() == expected.Algorithm() {
```

registry/remote/repository.go
```diff
@@ -2160,6 +2173,26 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
...
+		return verifyContentDigest(resp, expected)
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(limitReader(resp.Body, s.repo.maxMetadataBytes()))
+	if err != nil {
+		return err
+	}
+	if expected.Algorithm().FromBytes(body) != expected {
+		return fmt.Errorf("%s %q: %w", resp.Request.Method, resp.Request.URL, content.ErrMismatchedDigest)
+	}
```
TerryHowe, line 2191: `verifyContentDigest` doesn't wrap `ErrMismatchedDigest` for the same failure, so `errors.Is(err, content.ErrMismatchedDigest)` now depends on whether the registry's algorithm happened to match. This also drops the received/expected values every sibling error carries.

```suggestion
	if actual := expected.Algorithm().FromBytes(body); actual != expected {
		return fmt.Errorf(
			"%s %q: invalid response; content digest mismatch: received %q when expecting %q; %w",
			resp.Request.Method, resp.Request.URL,
			actual, expected, content.ErrMismatchedDigest,
		)
	}
```

registry/remote/repository.go
```diff
@@ -2143,6 +2146,16 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
 		contentDigest = serverHeaderDigest
 	}
 
+	if len(refDigest) > 0 && len(serverHeaderDigest) > 0 && contentDigest.Algorithm() != refDigest.Algorithm() {
```
TerryHowe, line 2149: `len(serverHeaderDigest) > 0` leaves the header-less case unfixed. The header is optional, and `calculateDigestFromResponse` hardcodes `digest.FromBytes` (sha256), so a sha512 ref against a registry that omits it still fails — and the message blames a header that was never sent:

```
GET ".../manifests/sha512:6174...": invalid response; digest mismatch in
Docker-Content-Digest: received "sha256:bc58..." when expecting "sha512:6174..."
```

Passing `refDigest.Algorithm()` into `calculateDigestFromResponse` would cover it.

registry/remote/repository.go
```diff
@@ -1567,8 +1567,8 @@ func (s *manifestStore) Fetch(ctx context.Context, target ocispec.Descriptor) (r
 	// Content-Length is not validated here because some registries (e.g. Harbor)
 	// return different values between HEAD (used by Resolve) and GET responses,
 	// typically due to transparent compression. Integrity is guaranteed by
-	// verifyContentDigest below.
-	if err := verifyContentDigest(resp, target.Digest); err != nil {
+	// digest verification below.
+	if err := s.verifyPullContentDigest(resp, target.Digest); err != nil {
```
TerryHowe, line 1571: Blob pulls keep strict `verifyContentDigest` (L1127/L1170/L1216/L1510), so sha512 blob pulls still fail. The spec text quoted in #949 is the generic Pull section — "canonical digest of the uploaded blob" — not manifest-specific. Worth stating in the description whether blobs are deliberately deferred.

registry/remote/repository.go
```diff
@@ -1737,6 +1737,9 @@ func (s *manifestStore) FetchReference(ctx context.Context, reference string) (d
 			// policy is evaluated below against the fetched descriptor, so
 			// skip the redundant evaluation in Resolve.
 			desc, err = s.Resolve(withPolicyChecked(ctx), reference)
+			if err == nil {
+				err = s.verifyPullContentDigest(resp, desc.Digest)
```
TerryHowe, line 1741: This also tightens the *same-algorithm* case, which isn't mentioned in the description. A tag re-pointed between the HEAD and the GET now errors, where `main` silently returned a descriptor that doesn't describe the body. Good hardening, but please call it out and add a test for it.

TerryHowe (commented):
Follow-up review on `9dc3dc3e`. The three things I flagged last round are fixed — the `Available()` guard, the wrapped mismatch error with received/expected values, and threading the ref algorithm into `calculateDigestFromResponse`. Tests pass for me locally, including `-race`.

A few things remain. The HEAD comment is the one I'd want settled before this merges — it loosens digest pinning in a way the description doesn't mention. The rest are smaller.

registry/remote/repository.go
```diff
@@ -2160,9 +2177,30 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
...
+// verifyPullContentDigest permits a different canonical digest algorithm on
+// manifest pulls, but verifies the returned bytes against the requested digest.
+// Pushes deliberately continue to use verifyContentDigest's strict comparison.
+func (s *manifestStore) verifyPullContentDigest(resp *http.Response, expected digest.Digest) error {
+	canonical, err := digest.Parse(resp.Header.Get(headerDockerContentDigest))
+	if err != nil || !expected.Algorithm().Available() || canonical.Algorithm() == expected.Algorithm() {
+		return verifyContentDigest(resp, expected)
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(limitReader(resp.Body, s.repo.maxMetadataBytes()))
```
TerryHowe, line 2189: `limitReader` is a plain `io.LimitReader` ([utils.go:99](https://github.com/oras-project/oras-go/blob/main/registry/remote/utils.go#L99)), so it truncates without error. A manifest larger than `MaxMetadataBytes` (default 4 MiB) therefore hashes to the wrong value, and this reports it as corrupted content:

```
GET ".../manifests/sha512:6174...": invalid response; content digest mismatch:
received "sha512:9f2c..." when expecting "sha512:6174..."; mismatched digest
```

A valid 5 MiB index served correctly by the registry fails here whenever the algorithms differ. Two problems:

1. `manifestStore.Fetch` used to stream manifests of unbounded size; on this path it no longer can.
2. `content.ErrMismatchedDigest` is a security signal — "the registry served poisoned content" — and callers that alert or blocklist on it will misfire on what is actually a size limit.

Reading `maxMetadataBytes+1` and returning `errdef.ErrSizeExceedsLimit` when over would keep the two failures distinguishable.

registry/remote/repository.go
```diff
@@ -2143,6 +2150,16 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
...
 
+	if len(refDigest) > 0 && len(serverHeaderDigest) > 0 && contentDigest.Algorithm() != refDigest.Algorithm() {
+		// A registry may report a canonical digest using a different algorithm.
+		// Preserve the requested identity, verifying it against the body for GET.
+		if httpMethod == http.MethodGet {
+			if err := s.verifyPullContentDigest(resp, refDigest); err != nil {
+				return ocispec.Descriptor{}, err
+			}
+		}
+		contentDigest = refDigest
```
TerryHowe, line 2161: On the HEAD path this adopts `refDigest` with nothing having verified it. Against a server that answers HEAD 200 with `Docker-Content-Digest: sha256:<digest of X>`:

```go
desc, err := repo.Resolve(ctx, "sha512:<digest of entirely different content>")
// err == nil, desc.Digest == that sha512 digest
ok, _ := repo.Exists(ctx, desc) // true
```

On `main` this is a digest mismatch error. As written it defeats digest pinning for `Resolve`, and through `Exists` it can make `oras.Copy` skip pushing a manifest to a destination that doesn't hold it and report success.

A HEAD has no body, so there is nothing to verify against — which argues for keeping the relaxation on the GET path only and letting `Fetch` establish identity. If accepting the unverified digest is deliberate, the doc comment should say so explicitly.

registry/remote/repository.go
TerryHowe, line 1741: This reads as verification but is inert in exactly the case the chunked branch exists for. `verifyPullContentDigest` starts with `digest.Parse(resp.Header.Get(headerDockerContentDigest))`; with no header `Parse` fails, so it delegates to `verifyContentDigest`, which returns `nil` for an empty header (L2222).

So a server whose HEAD advertises the right digest and whose chunked GET returns a different body with no digest header passes straight through — the caller gets the HEAD's descriptor paired with the wrong bytes.

The behavior predates this PR, but the description says GET content is verified against the requested digest, which isn't true here. For `ContentLength == -1` the body should be hashed against `desc.Digest` whether or not the header is present.

registry/remote/repository.go
TerryHowe, line 1571: Worth a doc-comment note: when the algorithms differ, `verifyPullContentDigest` buffers the whole manifest into a `bytes.Reader`, so `Fetch` stops streaming. Callers fetching many manifests concurrently hold up to `MaxMetadataBytes` per in-flight fetch. Same root cause as the `limitReader` comment below.

registry/remote/repository.go
```diff
@@ -2160,9 +2177,30 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
...
+	if err != nil || !expected.Algorithm().Available() || canonical.Algorithm() == expected.Algorithm() {
+		return verifyContentDigest(resp, expected)
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(limitReader(resp.Body, s.repo.maxMetadataBytes()))
+	if err != nil {
+		return err
+	}
+	if actual := expected.Algorithm().FromBytes(body); actual != expected {
+		return fmt.Errorf("%s %q: invalid response; content digest mismatch: received %q when expecting %q; %w",
```
TerryHowe, line 2194: Thanks for taking the earlier suggestion. One leftover: `verifyContentDigest` — the same-algorithm path this function delegates to — still doesn't wrap `content.ErrMismatchedDigest`, so `errors.Is(err, content.ErrMismatchedDigest)` is true or false for the same logical failure depending on which algorithm the registry reported. Wrapping it there too would settle it; I'm fine with that being a separate PR.

registry/remote/repository.go
```diff
@@ -2160,9 +2177,30 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
...
+// Pushes deliberately continue to use verifyContentDigest's strict comparison.
+func (s *manifestStore) verifyPullContentDigest(resp *http.Response, expected digest.Digest) error {
+	canonical, err := digest.Parse(resp.Header.Get(headerDockerContentDigest))
+	if err != nil || !expected.Algorithm().Available() || canonical.Algorithm() == expected.Algorithm() {
+		return verifyContentDigest(resp, expected)
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(limitReader(resp.Body, s.repo.maxMetadataBytes()))
+	if err != nil {
+		return err
```
TerryHowe, line 2191: Bare `return err` loses the method/URL context every other error in this file attaches (including `calculateDigestFromResponse`'s equivalent read), so a mid-body network failure surfaces as an opaque `unexpected EOF`.

```suggestion
		return fmt.Errorf("%s %q: failed to read response body: %w", resp.Request.Method, resp.Request.URL, err)
```

registry/remote/repository.go
```diff
@@ -2160,9 +2177,39 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
 	}, nil
 }
 
+// verifyPullContentDigest permits a different canonical digest algorithm on
+// manifest pulls, but verifies the returned bytes against the requested digest.
+// Pushes deliberately continue to use verifyContentDigest's strict comparison.
+// Alternate algorithms and headerless chunked responses buffer at most
+// MaxMetadataBytes before returning the verified body to the caller.
```
TerryHowe: The relaxation is GET-only, which I agree with — a HEAD has no body, so there's nothing to verify an alternate algorithm against. But that leaves `Resolve` and `Exists` rejecting exactly the digests `Fetch` now accepts, and nothing in the code or the PR description says so. Against a sha256 registry:

```go
repo.Resolve(ctx, sha512Digest.String())
// HEAD ".../manifests/sha512:6174...": invalid response; digest mismatch in
// Docker-Content-Digest: received "sha256:bc58..." when expecting "sha512:6174..."

repo.Exists(ctx, sha512Desc)   // false, same error
```

Identical on `main` and on this branch, so it's a scope boundary rather than a regression — but it means #949 is half-fixed, and a caller who reads "accept alternate manifest pull digest algorithms" will be surprised when `Exists` disagrees with `Fetch`. Please make the boundary explicit here, and add a line to the PR description too:

```suggestion
// verifyPullContentDigest permits a different canonical digest algorithm on
// manifest pulls, but verifies the returned bytes against the requested digest.
// Pushes deliberately continue to use verifyContentDigest's strict comparison.
// Alternate algorithms and headerless chunked responses buffer at most
// MaxMetadataBytes before returning the verified body to the caller.
//
// The relaxation applies to GET only. A HEAD response carries no body, so
// Resolve cannot verify an alternate algorithm and still reports a digest
// mismatch; resolving a manifest by a digest the registry does not use, or
// calling Exists with such a descriptor, continues to fail. Only Fetch and
// FetchReference accept an alternate algorithm.
```

This is additive, so it applies cleanly alongside the suggestion below whichever order you take them in.

registry/remote/repository.go
```diff
@@ -2160,9 +2177,39 @@ func (s *manifestStore) generateDescriptor(resp *http.Response, ref properties.R
...
+	} else if err != nil && err != io.EOF {
+		return fmt.Errorf("%s %q: failed to read response body: %w", resp.Request.Method, resp.Request.URL, err)
+	}
+	if actual := expected.Algorithm().FromBytes(body); actual != expected {
+		return fmt.Errorf("%s %q: invalid response; content digest mismatch: received %q when expecting %q; %w",
+			resp.Request.Method, resp.Request.URL, actual, expected, content.ErrMismatchedDigest)
+	}
+	resp.Body = io.NopCloser(bytes.NewReader(body))
+	return nil
+}
```
TerryHowe: `headerlessChunked` makes **every** header-less chunked response take the buffering path, not just the alternate-algorithm ones — so an ordinary sha256 pull is now capped at `MaxMetadataBytes`. A 5 MiB index served chunked with no `Docker-Content-Digest` header:

```
base 2db49b1: Fetch OK, read 5242906 bytes
PR   1c3ec9a: GET ".../manifests/sha256:7e98...": size exceeds limit   (errdef.ErrSizeExceedsLimit)
```

`ContentLength == -1` is not exotic: Go's transport sets it to `-1` whenever it transparently gunzips, so the same manifest behind a gzip-capable proxy fails identically — and the comment you added above the `Fetch` call site already names transparent compression as something Harbor does. `manifestStore.Fetch` never enforced `MaxMetadataBytes` (unlike `deleteWithIndexing`, which calls `limitSize` first), so this is a new ceiling on a path that was deliberately unbounded. `Tag` inherits it via `Fetch`.

Closing the header-less hole was the right call — it just doesn't need buffering. When no header is present there is no alternate algorithm in play: the descriptor's identity is already `expected`, and we only need to *detect* corruption, which a streaming verifier does without holding the body. Buffering then stays confined to the case that genuinely needs the whole body to establish a different identity.

```suggestion
func (s *manifestStore) verifyPullContentDigest(resp *http.Response, expected digest.Digest) error {
	headerStr := resp.Header.Get(headerDockerContentDigest)
	if !expected.Algorithm().Available() {
		return verifyContentDigest(resp, expected)
	}
	if headerStr == "" {
		if resp.ContentLength != -1 {
			return verifyContentDigest(resp, expected)
		}
		// A chunked response carries no digest to compare against, so verify
		// the body as the caller reads it. Manifests fetched this way were
		// never bounded by MaxMetadataBytes, so do not buffer them.
		resp.Body = newVerifyReadCloser(resp.Body, expected, resp.Request.Method, resp.Request.URL)
		return nil
	}
	canonical, err := digest.Parse(headerStr)
	if err != nil || canonical.Algorithm() == expected.Algorithm() {
		return verifyContentDigest(resp, expected)
	}
	defer resp.Body.Close()
	body, err := io.ReadAll(limitReader(resp.Body, s.repo.maxMetadataBytes()))
	if err != nil {
		return fmt.Errorf("%s %q: failed to read response body: %w", resp.Request.Method, resp.Request.URL, err)
	}
	var extra [1]byte
	if n, err := io.ReadFull(resp.Body, extra[:]); n > 0 {
		return fmt.Errorf("%s %q: %w", resp.Request.Method, resp.Request.URL, errdef.ErrSizeExceedsLimit)
	} else if err != nil && err != io.EOF {
		return fmt.Errorf("%s %q: failed to read response body: %w", resp.Request.Method, resp.Request.URL, err)
	}
	if actual := expected.Algorithm().FromBytes(body); actual != expected {
		return fmt.Errorf("%s %q: invalid response; content digest mismatch: received %q when expecting %q; %w",
			resp.Request.Method, resp.Request.URL, actual, expected, content.ErrMismatchedDigest)
	}
	resp.Body = io.NopCloser(bytes.NewReader(body))
	return nil
}

// verifyReadCloser hashes a response body as it is read and reports a digest
// mismatch on the read that ends the stream, keeping the body unbuffered.
type verifyReadCloser struct {
	io.ReadCloser
	verifier digest.Verifier
	expected digest.Digest
	method   string
	url      *url.URL
	checked  bool
}

func newVerifyReadCloser(rc io.ReadCloser, expected digest.Digest, method string, url *url.URL) io.ReadCloser {
	return &verifyReadCloser{
		ReadCloser: rc,
		verifier:   expected.Verifier(),
		expected:   expected,
		method:     method,
		url:        url,
	}
}

func (v *verifyReadCloser) Read(p []byte) (int, error) {
	n, err := v.ReadCloser.Read(p)
	if n > 0 {
		if _, werr := v.verifier.Write(p[:n]); werr != nil {
			return n, werr
		}
	}
	if err == io.EOF && !v.checked {
		v.checked = true
		if !v.verifier.Verified() {
			return n, fmt.Errorf("%s %q: invalid response; content digest mismatch: expecting %q; %w",
				v.method, v.url, v.expected, content.ErrMismatchedDigest)
		}
	}
	return n, err
}
```

`expected.Verifier()` panics on an unavailable algorithm, which is why the `Available()` guard moves to the top and returns early.

I applied this locally and ran the suite. Results:

```
5 MiB chunked, no header, sha256:   streamed 5242906 bytes, read err = <nil>
5 MiB gzipped, no header, sha256:   streamed 5242906 bytes, read err = <nil>
corrupted body, no header:          read err = GET "...": invalid response; content
                                    digest mismatch: expecting "sha256:bc58..."; mismatched digest
                                    (errors.Is ErrMismatchedDigest = true)
```

One test needs adjusting, since the error now surfaces on `Read` rather than from the function: `TestManifestPullBodyLimits/headerless_corrupt` (`manifest_digest_algorithm_test.go:109`). The `headerless valid` case still passes. Everything else in `./registry/remote/...` is green, `-race` included.

Worth deciding explicitly whether the alternate-algorithm path should keep buffering at all. Deferring its verification to the stream too would drop the ceiling entirely and match how `blobStore.Fetch` already returns an unverified body for callers to wrap — but it means `generateDescriptor` hands back a descriptor whose digest nothing has checked yet, which matters for callers that keep the descriptor and discard the reader. Buffering is the stronger choice there; I'd keep it and just document the bound, which is what the doc-comment suggestion above covers.

TerryHowe (commented):
Third pass on `1c3ec9a4`. I've resolved the four threads this commit addressed — the size-limit/corruption split, the unverified `refDigest` on the HEAD path, the inert header-less chunked check, and the missing method/URL context on read errors. All four look right to me.

Left open deliberately: the two asking for description changes (blob pulls staying strict, and the same-algorithm tag-repointing tightening — the test landed, the description still doesn't mention it), the `Fetch` doc-comment note, and `verifyContentDigest` not wrapping `ErrMismatchedDigest`, which we agreed is a separate PR.

Two suggestions below. The first is the one I'd want settled before merge: the new header-less chunked branch caps manifests that used to stream, in the plain sha256 case that has nothing to do with this feature. The second is documentation — `Resolve`/`Exists` still reject what `Fetch` now accepts, and that boundary should be written down.

PR #1426, opened 2026-09-17
perf(content): avoid allocating 3x the content size in ReadAll

ctxloop:
Finding 3 from #1395.

`bytes.Buffer.ReadFrom` asks for `bytes.MinRead` of spare capacity before every read, not only the first. `ReadAll` sizes its buffer to exactly `desc.Size`, so the last iteration cannot satisfy that by reslicing: `growSlice` allocates `2*cap` and copies the whole buffer across, and the read it made room for returns `(0, io.EOF)` from the `LimitedReader`. Each call therefore allocates roughly three times the content size to end up holding one copy.

The fix is one term: `make([]byte, 0, initialCap+bytes.MinRead)`.

## Results

`benchstat`, 6 runs each, go1.26.8 darwin/arm64 (Apple M4). Every row p=0.002.

```text
           sec/op      B/s       B/op
4 KiB     -19.95%   +24.93%   -58.07%
1 MiB     -11.37%   +12.83%   -66.38%
8 MiB     -14.42%   +16.85%   -66.63%
geomean   -15.32%   +18.10%   -63.90%
```

The benchmark lands in its own commit ahead of the fix, so the baseline is reproducible: check out the first commit for `before`, the second for `after`.

## Two notes

**`allocs/op` is omitted on purpose.** With GC enabled it reads as -1 / -3 / -5 across the three sizes, but under `GOGC=off` the delta is exactly -1 everywhere (14 -> 13). The additional allocations appear to be GC-induced allocations counted by `testing`, so only sec/op and B/op are quoted above.

**Content above the 32 MiB cap is unaffected.** At 40 MiB, B/op moves from 100,665,503 to 100,681,946 (+0.016%, size-class rounding) with no measurable change in sec/op. The buffer has to grow regardless at that point, so the headroom neither helps nor hurts.

## Safety

The 32 MiB cap on the initial allocation and the `VerifyReader` size and digest checks are untouched, so a forged `desc.Size` still cannot drive an unbounded allocation — the cap now admits 512 additional bytes out of 32 MiB.

The boundaries this touches are already covered by `TestReadAll_EmptyContent`, `TestReadAll_OversizedDescriptor` and `TestReadAll_LargeContent` (33 MiB); all pass, and `go test -race ./...` is green across the repo.

Refs #1395


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #1428, opened 2026-09-18
perf(graph): bound goroutine fan-out in IndexAll

jadhavgaurav:
## Summary

`oci.New` loads `index.json` and calls `graph.Memory.IndexAll` for each top-level manifest. `IndexAll` recurses through `syncutil.Go(ctx, nil, fn, successors...)` with a nil limiter, so `LimitRegion` returns nil and `region.Start()` becomes a no-op. Every node in the graph gets its own goroutine, each opening a file, reading it, and parsing JSON, with nothing bounding the fan-out. A store with a few hundred manifests fans out thousands of concurrent file opens competing for the same disk.

This adds a `semaphore.Weighted`-backed `indexLimiter` field to `graph.Memory`, created once in `NewMemory` and reused across every `IndexAll` call on that store (so the bound holds across the whole store, not just one top-level manifest), and threads it through both `syncutil.Go` call sites in place of `nil`.

Bounding the fan-out with a real, finite pool of slots introduces a new correctness requirement: a node that holds its slot while waiting on its successors, instead of releasing it first, deadlocks any chain deeper than the concurrency limit, since the parent sits on a slot its own descendants need to make progress. `region.End()` is called before recursing to release the slot, matching the pattern `copyGraph` already uses in `copy.go` for the same reason (see the comment at `copy.go:274` and the issue's own callout of that pattern under "What's already right").

`defaultIndexConcurrency` is set to 64, separate from `copy.go`'s `defaultConcurrency` (3): that constant bounds network round trips, this one bounds local disk I/O (open + read + parse a file), so a higher value is appropriate.

## Testing

Two new tests in `internal/graph/memory_test.go`:
- `TestMemory_IndexAllConcurrencyBound` builds a manifest with `4 * defaultIndexConcurrency` direct successors and wraps the fetcher to track the maximum number of concurrent in-flight `Fetch` calls, asserting it never exceeds `defaultIndexConcurrency`.
- `TestMemory_IndexAllDeepChainNoDeadlock` builds a linear chain deeper than `defaultIndexConcurrency` and asserts `IndexAll` returns within a bounded timeout, rather than deadlocking.

Both were verified to fail in the expected, targeted way when only the relevant half of the fix is reverted (source only, test file untouched):
- Reverting just the `nil` -> `m.indexLimiter` change: `TestMemory_IndexAllConcurrencyBound` fails, reporting 256 concurrent fetches against a limit of 64.
- Reverting just the `region.End()` call: `TestMemory_IndexAllDeepChainNoDeadlock` times out after 10s, reproducing the deadlock the release-before-recurse ordering prevents.

Checks run locally (this checkout has a pre-existing, unrelated `go.mod`/`vendor/modules.txt` mismatch from a recent dependency bump, so `-mod=mod` was used instead of regenerating `vendor/`, which would have pulled in an unrelated diff):
- `golangci-lint` is not installed locally; per the repo's own config it enables only `govet`, `ineffassign`, and the `gofmt` formatter, and govet/gofmt were run directly above.

## Scope

This addresses finding 6 only. The outer `loadIndex` loop (`content/oci/readonlyoci.go`) and `gcIndex` (`content/oci/oci.go`) still call `IndexAll` once per top-level manifest sequentially; the issue also suggests parallelizing that outer loop under the same limiter. I left that out of this PR: `graph.Memory`'s internal state is already safe for concurrent access (see the locking in `index()`), but parallelizing the callers touches `content/oci/readonlyoci.go` and `content/oci/oci.go`'s tagging logic in a second package, which felt like a separate, independently reviewable change from bounding the fan-out itself.

Part of #1395 (finding 6).


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

Issue #338, opened 2022-09-30
Allow chunked blob push


shizhMSFT:
The current implementation of `content/oci` and `registry/remote` only supports monolithic blob upload where the `descriptor` must be known in advance.

If we want to support chunked blob push, we should have the implementations to implement a new interface like
```go
type UnsafePusher interface {
	// UnsafePush pushes the content, matching the specified field in the expected descriptor.
	// - The `Digest` of the expected descriptor is recalculated if empty
	// - The `Size` of the expected descriptor is recalculated if less than zero
	// The recalculated descriptor is returned upon a successful push.
	UnsafePush(ctx context.Context, expected ocispec.Descriptor, content io.Reader) (ocispec.Descriptor, error)
}
```

This issue is to track the discussion on whether we should support chunked blob push with security considerations.
- Since the digest might be missing, the integrity of the content might not be guaranteed. That is, the pushed content might be different from expected.
- Since the size might be missing, arbitrary length or even infinite length of content might be written to the target without notice.

sparr:
For possibly useful reference, containerd's content package implements a [`Writer`](https://pkg.go.dev/github.com/containerd/containerd/content#Writer) that accepts multiple writes followed by a single `Commit`.

chlins:
In AI, we often deal with very large models and datasets. Chunked transfer would be really important for these scenarios for better reliability.

Wwwsylvia:
Related issue on upstream: https://github.com/opencontainers/distribution-spec/issues/574

PR #1434, opened 2026-09-18
feat(remote): support chunked blob push

jakobmoellerdev:
Adds opt-in chunked blob push to `registry/remote`. Today `Repository.Push` uploads blobs monolithically (POST session, then PUT). This supports the OCI Distribution Spec [chunked upload](https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#pushing-a-blob-in-chunks) (POST session → PATCH chunks with `Content-Range` → PUT close), so clients can bound per-request body size.

New field:

```go
// MaxChunkSize is the target size in bytes for each PATCH chunk when
// pushing a blob in chunks. Zero or negative (default) = monolithic.
MaxChunkSize int64
```

- Default (`<= 0`) keeps the current monolithic behavior.
- A blob that fits in one chunk stays monolithic.
- `OCI-Chunk-Min-Length` from the session response raises the floor for non-final chunks.
- If the session can't be opened before any content is read, `Push` falls back to the monolithic path.

Reuses the existing request path (`do`, `buildRepositoryBlobUploadURL`, `errutil.ParseErrorResponse`) and shares the upload `Location` handling with the monolithic path (keeping the port-443 work-around and the cross-host/scheme-downgrade guards).


TerryHowe:
Related to https://github.com/oras-project/oras-go/issues/338 probably

registry/remote/chunked.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	if computed := up.digester.Digest(); computed != expected.Digest {
+		s.cancelUpload(ctx, up.location)
+		return fmt.Errorf("chunked blob push: content digest %s does not match expected digest %s", computed, expected.Digest)
+	}
+	if up.offset != expected.Size {
+		s.cancelUpload(ctx, up.location)
+		return fmt.Errorf("chunked blob push: uploaded %d bytes, expected %d", up.offset, expected.Size)
+	}
```
TerryHowe, line 107: **Critical.** Three defects in one spot, verified locally against the PR branch:

1. `digest.Canonical` (L124) hardcodes sha256, so any non-sha256 descriptor uploads the whole blob then fails: `content digest sha256:2cf24dba... does not match expected digest sha512:9b71d224...`. `repository.go:2181` and `content.NewVerifyReader` both honour `desc.Digest.Algorithm()`.
2. `make([]byte, up.chunk)` (L134) is sized by the registry's `OCI-Chunk-Min-Length`, which `parseChunkMinLength` only checks for negativity. A header of `1<<62` panics the caller: `runtime error: makeslice: len out of range`. A merely large value OOMs.
3. The reader is never bounded by `expected.Size` — a 5-byte descriptor against a 5000-byte reader pushes all 5000 bytes before the post-hoc size check fires. The monolithic path short-circuits this via the `ContentLength != expected.Size` check.

```suggestion
	up, err := s.openUploadSession(ctx, chunkSize)
	if err != nil {
		// The session never opened; no bytes consumed, fallback is safe.
		return fmt.Errorf("%w: %w", errChunkedUploadNotStarted, err)
	}
	// Digest with the descriptor's own algorithm, not the canonical one.
	algo := expected.Digest.Algorithm()
	if !algo.Available() {
		s.cancelUpload(ctx, up.location)
		return fmt.Errorf("chunked blob push: unsupported digest algorithm %q", algo)
	}
	up.digester = algo.Digester()
	// Never buffer more than the blob itself, whatever the registry advertised.
	if up.chunk > expected.Size {
		up.chunk = expected.Size
	}

	// Bound the read at the declared size so an over-long reader cannot push
	// more bytes than the descriptor claims.
	if err := s.uploadChunks(ctx, up, io.LimitReader(content, expected.Size)); err != nil {
		return err
	}
```

Capping `up.chunk` at `expected.Size` also settles the "a lone PATCH is strictly worse than monolithic" case when min-length raises the chunk past the blob size. Applied as written this builds clean, fixes all three, and leaves the PR's existing chunked tests green.

registry/remote/chunked.go
```diff
@@ -0,0 +1,241 @@
...
+
+	return s.closeUploadSession(ctx, up, expected.Digest)
+}
+
+// openUploadSession performs the POST that starts a blob upload and returns the
+// initialised session state. A non-sha256 expected digest advertises its
+// algorithm via the digest-algorithm query parameter. The returned session has
```
TerryHowe: No `digest-algorithm` query parameter is set anywhere in this function.

```suggestion
// initialised session state. The returned session has its chunk size raised to
// any registry-advertised OCI-Chunk-Min-Length.
```

registry/remote/chunked.go
```diff
@@ -0,0 +1,241 @@
...
+		_, _ = hasher.Write(buf[:n])
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodPatch, up.location.String(), bytes.NewReader(buf[:n]))
+		if err != nil {
+			if up.consumed {
+				s.cancelUpload(ctx, up.location)
+			}
+			return fmt.Errorf("chunked blob push: failed to build PATCH request: %w", err)
+		}
+		req.Header.Set("Content-Type", "application/octet-stream")
```
TerryHowe, line 183: The monolithic path deliberately reuses the session POST's `Authorization` header. This drops that, so with a caller-supplied `auth.Client` that has no `Cache`, `Client.cache()` returns `noCache{}` (`auth/client.go:203`) and every PATCH costs a 401 plus a token fetch — 2N extra round trips on exactly the large blobs chunking is for. Worth carrying the header across, same as `completePushAfterInitialPost`.

registry/remote/chunked.go
```diff
@@ -0,0 +1,241 @@
...
+		req.Header.Set("Content-Range", fmt.Sprintf("%d-%d", up.offset, up.offset+int64(n)-1))
+		req.ContentLength = int64(n)
+		up.consumed = true
+
+		resp, err := s.repo.do(req)
+		if err != nil {
+			return fmt.Errorf("chunked blob push: PATCH failed after %d bytes: %w", up.offset, err)
+		}
```
TerryHowe, line 192: Every other post-consumption failure path cancels; a transport error doesn't, leaving the session dangling.

```suggestion
		resp, err := s.repo.do(req)
		if err != nil {
			s.cancelUpload(ctx, up.location)
			return fmt.Errorf("chunked blob push: PATCH failed after %d bytes: %w", up.offset, err)
		}
```

registry/remote/chunked.go
```diff
@@ -0,0 +1,241 @@
...
+		return fmt.Errorf("chunked blob push: registry returned digest %q, expected %q", returned, finalDigest.String())
+	}
+	return nil
+}
+
+// cancelUpload issues a best-effort DELETE to release an in-progress upload
+// session. Per the OCI spec, clients SHOULD ignore any failures.
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#canceling-a-blob-upload
+func (s *blobStore) cancelUpload(ctx context.Context, location *url.URL) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, location.String(), http.NoBody)
```
TerryHowe, line 255: Best-effort cancel is a no-op when the failure *was* ctx cancellation or deadline expiry — precisely when the session needs releasing. Needs a `"time"` import.

```suggestion
func (s *blobStore) cancelUpload(ctx context.Context, location *url.URL) {
	// Detach from ctx: cancellation or deadline expiry is exactly when the
	// session most needs releasing.
	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 10*time.Second)
	defer cancel()
	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, location.String(), http.NoBody)
```

TerryHowe (commented):
Reviewed by checking out the branch and building/testing locally. The `resolveUploadLocation` extraction is a genuine improvement — the port-443 work-around and both GHSA guards now apply to every hop, not just the first — and the `errChunkedUploadNotStarted`/`consumed` design is the right way to make fallback safe.

Two blocking issues, both confirmed with throwaway tests: non-sha256 descriptors can never succeed on the chunked path, and a registry response header can panic the client. Details inline, with suggestions that build clean and keep the existing chunked tests green.

registry/remote/chunked.go
jakobmoellerdev: Applied as suggested and added unit tests covering non-sha256 digests (SHA-512), huge min chunk lengths, and overlong readers.

registry/remote/chunked.go
jakobmoellerdev: Updated doc comment as suggested.

registry/remote/chunked.go
jakobmoellerdev: Carried the `Authorization` header across all PATCH and PUT requests, covered by `TestRepository_Push_ChunkedCarriesAuthorization`.

registry/remote/chunked.go
jakobmoellerdev: Applied as suggested; upload session is now cancelled on transport errors, covered with a test.

registry/remote/chunked.go
jakobmoellerdev: Applied with `context.WithoutCancel` and a 10s timeout, verified with a test.

registry/remote/chunked.go
TerryHowe: Following up on my own suggestion — carrying the header turns out to be the wrong lever.

`auth.Client.Do` returns early whenever `Authorization` is already set (`auth/client.go:268`): no cache lookup, and no 401 challenge/re-auth. So every PATCH and the closing PUT now lose token refresh, on exactly the long-running uploads chunking exists for. Registry tokens are commonly ~5 min.

Repro — test registry that rotates its token mid-upload:

```
chunked blob push: unexpected PATCH status 401 after 256 bytes: PATCH ".../uploads/session": response status code 401: Unauthorized
```

Dropping the two `req.Header.Set("Authorization", up.auth)` lines makes it pass.

My original concern still stands, so please don't just revert. Suggest: keep the header, but on a 401 from a PATCH, clear `up.auth` and retry that chunk once — the chunk is still in `buf`, so the retry is free. Worth noting `auth.DefaultClient` does have a cache, so the no-`Cache` case is opt-in.

jakobmoellerdev: Good catch — implemented exactly that lever. The session `Authorization` header is still carried across PATCH/PUT (keeping the no-`Cache` round-trip savings from the original concern), but a 401 now drops `up.auth` and retries the request once. The retry has no cached header, so `auth.Client.Do` runs its full challenge/re-auth path and refreshes the token; the fresh token is captured back into `up.auth` for the remaining chunks. The PATCH retry reuses the buffered chunk, and the closing PUT gets the same treatment.

Added `TestRepository_Push_ChunkedReauthenticatesOnTokenRotation`: a registry that rotates its bearer token just before the first PATCH. It fails pre-fix with the exact `unexpected PATCH status 401 after 0 bytes` you reported and passes with the retry in place.

Also rebased onto latest upstream `main`.

TerryHowe: Verified on `4b9ec68`. `sendChunkRequest` drops `up.auth` on a 401 and replays the rebuilt request with no `Authorization`, so `auth.Client.Do` runs its full challenge path and refreshes the token, and the refreshed header is carried forward. `TestRepository_Push_ChunkedReauthenticatesOnTokenRotation` exercises a real mid-upload rotation and asserts the chunk is not lost and the session is not cancelled. Resolving.

registry/remote/chunked.go
```diff
@@ -0,0 +1,322 @@
...
+		retry.Header.Set("Content-Type", "application/octet-stream")
+		return retry, nil
+	})
+	if err != nil {
+		return fmt.Errorf("chunked blob push: PUT failed: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusCreated {
+		return fmt.Errorf("chunked blob push: unexpected PUT status %d: %w", resp.StatusCode, errutil.ParseErrorResponse(resp))
+	}
```
TerryHowe, line 284: A failed closing PUT leaves the upload session orphaned. Three exits here return without cancelling — the `sendChunkRequest` error and the non-201 status — while every other post-consumption path in this file cancels. `pushChunked`'s own doc says "later failures return a wrapped error and best-effort cancel the session", so this is the one place that breaks that contract, and it is the point where the registry is holding the entire blob.

There is a second, quieter bug in the same function: by the time the build-request failure at the top calls `s.cancelUpload(ctx, up.location)`, `up.location` already carries `?digest=...` from the lines just above it, so the DELETE does not address the session. Capturing the bare session URL first fixes both.

The `Docker-Content-Digest` mismatch below is deliberately left alone — that path is only reachable after a 201, so the blob is committed and the session is already gone.

Verified locally on `4b9ec68`: `gofmt` clean, `go build ./...` clean, `-run Chunk` passes, and a scratch assertion on the existing `chunkedUploadRegistry` (which records DELETEs) shows `cancelled` flipping from `false` to `true` for `TestRepository_Push_ChunkedPUTStatusError`. Worth asserting that in `TestRepository_Push_ChunkedPUTStatusError` and `..._ChunkedPUTTransportError`, which currently only check the error text.

```suggestion
	// Keep the bare session URL for cancellation: the commit URL below carries
	// the digest query, which has no meaning on a DELETE.
	sessionURL := *up.location

	q := up.location.Query()
	q.Set("digest", finalDigest.String())
	up.location.RawQuery = q.Encode()

	req, err := http.NewRequestWithContext(ctx, http.MethodPut, up.location.String(), http.NoBody)
	if err != nil {
		s.cancelUpload(ctx, &sessionURL)
		return fmt.Errorf("chunked blob push: failed to build PUT request: %w", err)
	}
	req.ContentLength = 0
	req.Header.Set("Content-Type", "application/octet-stream")

	resp, err := s.sendChunkRequest(up, req, func() (*http.Request, error) {
		retry, rerr := http.NewRequestWithContext(ctx, http.MethodPut, up.location.String(), http.NoBody)
		if rerr != nil {
			return nil, rerr
		}
		retry.ContentLength = 0
		retry.Header.Set("Content-Type", "application/octet-stream")
		return retry, nil
	})
	if err != nil {
		s.cancelUpload(ctx, &sessionURL)
		return fmt.Errorf("chunked blob push: PUT failed: %w", err)
	}
	defer resp.Body.Close()
	if resp.StatusCode != http.StatusCreated {
		respErr := errutil.ParseErrorResponse(resp)
		s.cancelUpload(ctx, &sessionURL)
		return fmt.Errorf("chunked blob push: unexpected PUT status %d: %w", resp.StatusCode, respErr)
	}
```

registry/remote/chunked.go
```diff
@@ -0,0 +1,322 @@
...
+	if returned := resp.Header.Get("Docker-Content-Digest"); returned != "" && returned != finalDigest.String() {
+		return fmt.Errorf("chunked blob push: registry returned digest %q, expected %q", returned, finalDigest.String())
+	}
+	return nil
```
TerryHowe, line 288: This reimplements `verifyContentDigest` (`registry/remote/repository.go:1684`), which already does exactly this job and is used by six call sites — including the monolithic blob push this path is the alternative to, at `repository.go:1390`.

Three things the local copy loses:

- **No `digest.Parse`.** A malformed or truncated header is reported as a digest *mismatch* rather than `invalid response header`, which sends you looking for a corrupted upload instead of a misbehaving registry.
- **The header name is hardcoded** instead of using the `headerDockerContentDigest` constant declared at `repository.go:58`.
- **The error shape diverges** from every other digest-verification failure in the package — no method, no URL.

Verified on `4b9ec68` in a scratch worktree: `gofmt` clean, `go build ./...` clean, full `./registry/remote/` package green.

One caveat — applying this alone turns `TestRepository_Push_ChunkedRegistryDigestMismatch` red, because it asserts on the bespoke string. The assertion at `chunked_test.go:464` needs:

```go
if !strings.Contains(err.Error(), "digest mismatch in Docker-Content-Digest") {
```

The mismatch is still caught either way; only the message changes. With that one line updated the whole package passes.

Separately, the monolithic path ends with `s.checkOCISubjectHeader(resp)` before verifying. It is a practical no-op for a blob PUT, so I have left it out of the suggestion, but if you want the two close paths byte-identical it belongs here too — your call.

```suggestion
	return verifyContentDigest(resp, finalDigest)
```

registry/remote/chunked.go
```diff
@@ -0,0 +1,322 @@
...
+			return fmt.Errorf("chunked blob push: unexpected PATCH status %d after %d bytes: %w", resp.StatusCode, up.offset, respErr)
+		}
+		next, err := resolveUploadLocation(resp, req)
+		resp.Body.Close()
+		if err != nil {
+			s.cancelUpload(ctx, up.location)
+			return fmt.Errorf("chunked blob push: invalid PATCH location after %d bytes: %w", up.offset, err)
+		}
+		up.location = next
+		up.offset += int64(n)
```
TerryHowe, line 211: The `Range` header the registry returns on the 202 is ignored — `up.offset` advances by the full `n` on the assumption that the registry stored everything that was sent.

That assumption is the one thing a chunked protocol exists to not make. If a registry commits only part of a chunk and reports so via `Range`, every subsequent `Content-Range` is off by the shortfall. The failure mode is bad: the upload keeps going, the registry keeps accepting or rejecting chunks at the wrong offsets, and it surfaces as a digest mismatch at the closing PUT after the entire blob has been streamed — pointing at content corruption rather than at the offset drift that actually happened.

Reading the header back and comparing it against `up.offset + int64(n)` turns that into an immediate, accurate error at the chunk that actually diverged.

I am not posting this as an applicable suggestion, because the right handling is a design call I would rather leave to you:

1. **Resync** `up.offset` to the registry-reported end and re-send the remainder. Most robust, but it means holding the unsent tail of the chunk, which changes the buffer contract in `uploadChunks`.
2. **Fail fast** with a clear error naming the reported and expected offsets. Much simpler, and no worse than today, which fails later with a misleading message.

There is also a parsing wrinkle worth knowing before you pick: the distribution spec shows `Range: 0-<end>`, but some registries emit the HTTP-conventional `bytes=0-<end>`. A strict parser that errors on anything it cannot read would reject those registries outright, so whichever option you take, an unparsable `Range` should be treated as absent rather than fatal.

registry/remote/repository.go
```diff
@@ -150,6 +150,22 @@ type Repository struct {
...
+	// If zero or negative (the default), blobs are pushed monolithically in a
+	// single request, preserving the previous behavior. If positive, a blob
+	// larger than the effective chunk size is uploaded in chunks per the OCI
+	// Distribution Spec: an upload session is opened, the content is streamed in
+	// PATCH requests, and the session is closed with a final PUT. A
+	// registry-advertised OCI-Chunk-Min-Length raises this floor for all but the
+	// final chunk. If the chunked upload cannot be started (e.g. the registry
+	// does not support it), the push transparently falls back to a monolithic
+	// upload before any content is consumed.
+	//
```
TerryHowe, line 165: This reads as though `MaxChunkSize` governs blob pushes generally, but it only reaches `blobStore.Push`. `blobStore.Mount` (`repository.go:1213`) hands off to `completePushAfterInitialPost` directly at `repository.go:1270` when the registry does not implement the mount endpoint, so it never passes the chunking branch and stays monolithic no matter what this is set to.

That is a defensible scope for a first cut — mount fallback is the uncommon path — but it is exactly the case someone sets `MaxChunkSize` for: a large blob that could not be mounted and now has to be uploaded in full. Worth saying so here rather than leaving it to be discovered.

Verified on `4b9ec68`: `gofmt` clean, `go build ./...` clean.

```suggestion
	// upload before any content is consumed.
	//
	// MaxChunkSize applies to [Repository.Push] only. Mount falls back to a
	// monolithic upload when the registry does not implement the mount
	// endpoint, regardless of this setting.
	//
```

registry/remote/chunked.go
```diff
@@ -0,0 +1,365 @@
...
+	if !algo.Available() {
+		s.cancelUpload(ctx, up.location)
+		return fmt.Errorf("chunked blob push: unsupported digest algorithm %q", algo)
+	}
+	up.digester = algo.Digester()
+
+	// Never buffer more than the blob itself, whatever the registry advertised.
+	if up.chunk > expected.Size {
+		up.chunk = expected.Size
+	}
```
TerryHowe: A registry-controlled `OCI-Chunk-Min-Length` can raise `up.chunk` to the whole blob; the clamp then makes this one `make([]byte, blobSize)` and a single unbounded PATCH. Confirmed with `MaxChunkSize=4096` and `OCI-Chunk-Min-Length: 1<<40` on an 8 MiB blob: 1 PATCH, `Content-Range: 0-8388607`.

```suggestion
	// A chunk at least as large as the blob means a lone PATCH plus the closing
	// PUT, which is strictly worse than a monolithic POST/PUT, and it would size
	// the read buffer at the whole blob however large the registry's advertised
	// minimum. Nothing is consumed yet, so release the session and fall back.
	if up.chunk >= expected.Size {
		s.cancelUpload(ctx, up.location)
		return fmt.Errorf("%w: effective chunk size %d is not smaller than blob size %d", errChunkedUploadNotStarted, up.chunk, expected.Size)
	}
```

registry/remote/chunked.go
```diff
@@ -0,0 +1,365 @@
...
+	location *url.URL
+	// chunk is the effective chunk size, raised to any registry-advertised
+	// OCI-Chunk-Min-Length.
+	chunk int64
+	// offset is the number of bytes uploaded so far (the next Content-Range
+	// start).
+	offset int64
+	// consumed reports whether at least one PATCH has been issued; after that
+	// the reader cannot be rewound, so a monolithic fallback is impossible.
+	consumed bool
```
TerryHowe: Dead once the guards below go.

```suggestion
```

registry/remote/chunked.go
```diff
@@ -0,0 +1,365 @@
...
+			if up.consumed {
+				s.cancelUpload(ctx, up.location)
+			}
+			return fmt.Errorf("chunked blob push: failed to build PATCH request: %w", err)
+		}
+		req.Header.Set("Content-Type", "application/octet-stream")
+		req.Header.Set("Content-Range", fmt.Sprintf("%d-%d", up.offset, up.offset+int64(n)-1))
+		req.ContentLength = int64(n)
+		up.consumed = true
+		resp, err := s.sendChunkRequest(up, req, func() (*http.Request, error) {
```
TerryHowe: `consumed` tracks whether the *reader* can be rewound, but the session is already open on entry, so both guards leak it. Confirmed: `Size: 11` with a reader erroring on its first `Read` gives methods `[POST]`, no DELETE.

```suggestion
// running digest, byte offset, and push location as it goes. The session is
// already open on entry, so every failure path cancels it.
func (s *blobStore) uploadChunks(ctx context.Context, up *chunkedUpload, content io.Reader) error {
	hasher := up.digester.Hash()
	buf := make([]byte, up.chunk)
	for {
		n, readErr := io.ReadFull(content, buf)
		if readErr == io.EOF {
			// Clean end of stream on a chunk boundary.
			return nil
		}
		if readErr != nil && readErr != io.ErrUnexpectedEOF {
			s.cancelUpload(ctx, up.location)
			return fmt.Errorf("chunked blob push: failed to read content after %d bytes: %w", up.offset, readErr)
		}
		if n == 0 {
			return nil
		}
		_, _ = hasher.Write(buf[:n])

		req, err := http.NewRequestWithContext(ctx, http.MethodPatch, up.location.String(), bytes.NewReader(buf[:n]))
		if err != nil {
			s.cancelUpload(ctx, up.location)
			return fmt.Errorf("chunked blob push: failed to build PATCH request: %w", err)
		}
		req.Header.Set("Content-Type", "application/octet-stream")
		req.Header.Set("Content-Range", fmt.Sprintf("%d-%d", up.offset, up.offset+int64(n)-1))
		req.ContentLength = int64(n)
		resp, err := s.sendChunkRequest(up, req, func() (*http.Request, error) {
```

PR #1436, opened 2026-09-22
perf(status): avoid allocating a channel on every TryCommit call

jadhavgaurav:
## What

`Tracker.TryCommit` (`internal/status/tracker.go:41`) built its notification channel with `make(chan struct{})` as an argument to `LoadOrStore`, so the channel was allocated on every call, including the calls where the descriptor was already committed and the freshly made channel was thrown away immediately.

`copyGraph` hits the throwaway path at least twice per node:

- once when a worker goroutine finds a node already claimed by another worker (`copy.go:233`)
- once in the parent's wait loop, which calls `TryCommit` on every successor purely to fetch its already-existing notification channel (`copy.go:279`)

Neither of these calls needs a new channel; they only need to read the existing one.

## Change

`TryCommit` now calls `sync.Map.Load` first. If the entry already exists, it returns that channel and `false` without allocating anything. The channel allocation and `LoadOrStore` call only happen on an actual miss, which is the genuine first-commit case.

## Tests

`internal/status/tracker_test.go` gains two benchmarks:

- `BenchmarkTracker_TryCommit_Hit`: repeated calls on an already-committed descriptor, the throwaway case described above.
- `BenchmarkTracker_TryCommit_Miss`: each call uses a brand-new descriptor, the legitimate allocation case the fix must not touch.

```
before: TryCommit_Hit    50.2 ns/op    160 B/op   2 allocs/op
after:  TryCommit_Hit    21.7 ns/op      0 B/op   0 allocs/op

before: TryCommit_Miss  374.2 ns/op    265 B/op   3 allocs/op
after:  TryCommit_Miss  399.1 ns/op    265 B/op   3 allocs/op
```

The already-committed path drops to zero allocations and roughly halves in time. The first-commit path is unchanged within noise.

Verified by reverting only the body of `TryCommit` back to the original single-`LoadOrStore` form (source only, `git apply -R` on a saved patch, test file untouched) and rerunning: the benchmark reproduces the original numbers exactly (160 B/op, 2 allocs/op on the hit path), then reapplying the patch restores the fix.

Existing `TestTracker_TryCommit` still passes and covers the commit/already-committed/notify behavior end to end.

## Checks

Run locally (this checkout has the same pre-existing, unrelated `go.mod`/`vendor/modules.txt` mismatch noted in #1418 and #1428, so `-mod=mod` was used instead of regenerating `vendor/`):

- `golangci-lint` is not installed locally; per the repo's own config it enables only `govet`, `ineffassign`, and the `gofmt` formatter, and govet/gofmt were run directly above.

## Scope

This addresses finding 9 of https://github.com/oras-project/oras-go/issues/1395 (the `TryCommit` half; finding 10, `redirectSafeClient` in `registry/remote/auth/client.go`, is a separate, unrelated call site left for its own PR). Part of #1395; this issue is not closed here.


internal/status/tracker_test.go
```diff
@@ -16,8 +16,10 @@ limitations under the License.
 package status
 
 import (
+	"fmt"
 	"testing"
 
+	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 )
```
TerryHowe, line 24: Both new imports become unused with the suggestion below.

```suggestion
import (
	"testing"

	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
)
```

internal/status/tracker_test.go
```diff
@@ -69,3 +71,42 @@ func TestTracker_TryCommit(t *testing.T) {
...
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		if _, committed := tracker.TryCommit(descs[i]); !committed {
+			b.Fatal("TryCommit() should commit on a new descriptor")
+		}
+	}
+}
```
TerryHowe, line 112: Fixed descriptor set + a fresh `Tracker` per iteration: bounds memory and makes ns/op independent of `b.N`. Keying on `Size` works because `descriptor.FromOCI` keys on `{MediaType, Digest, Size}`, and it drops both imports plus 1024 string allocations from setup. Added a parallel case since both callers fan out over goroutines.

```suggestion
// benchTrackerNodes is the number of distinct descriptors each TryCommit
// benchmark iteration walks. Fixing it keeps the reported ns/op independent
// of b.N and bounds the memory the benchmark holds.
const benchTrackerNodes = 1024

// benchDescriptors builds benchTrackerNodes distinct descriptors. Only the
// fields descriptor.FromOCI keys on need to differ.
func benchDescriptors() []ocispec.Descriptor {
	descs := make([]ocispec.Descriptor, benchTrackerNodes)
	for i := range descs {
		descs[i] = ocispec.Descriptor{Size: int64(i)}
	}
	return descs
}

// BenchmarkTracker_TryCommit_Miss measures first commits, where a channel
// allocation is required. This is the legitimate allocation that must not
// be removed.
func BenchmarkTracker_TryCommit_Miss(b *testing.B) {
	descs := benchDescriptors()

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		tracker := NewTracker()
		for _, desc := range descs {
			if _, committed := tracker.TryCommit(desc); !committed {
				b.Fatal("TryCommit() should commit on a new descriptor")
			}
		}
	}
}

// BenchmarkTracker_TryCommit_Hit measures repeat calls on already-committed
// descriptors, which must not allocate a channel only to throw it away. This
// is the shape of copyGraph's wait loop in copy.go, which calls TryCommit on
// every already-committed successor purely to fetch its notification channel.
func BenchmarkTracker_TryCommit_Hit(b *testing.B) {
	descs := benchDescriptors()
	tracker := NewTracker()
	for _, desc := range descs {
		if _, committed := tracker.TryCommit(desc); !committed {
			b.Fatal("TryCommit() first call should commit")
		}
	}

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for _, desc := range descs {
			if _, committed := tracker.TryCommit(desc); committed {
				b.Fatal("TryCommit() should not commit on an already-committed descriptor")
			}
		}
	}
}

// BenchmarkTracker_TryCommit_Parallel measures concurrent repeat calls, the
// way copyGraph and Memory.IndexAll actually call TryCommit.
func BenchmarkTracker_TryCommit_Parallel(b *testing.B) {
	descs := benchDescriptors()
	tracker := NewTracker()
	for _, desc := range descs {
		if _, committed := tracker.TryCommit(desc); !committed {
			b.Fatal("TryCommit() first call should commit")
		}
	}

	b.ReportAllocs()
	b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
		for i := 0; pb.Next(); i++ {
			tracker.TryCommit(descs[i%len(descs)])
		}
	})
}
```

TerryHowe (commented):
`_Miss` ns/op scales with `b.N` — 409 → 480 → 638 ns/op at `-benchtime=200ms` / `1s` / `3s`, since the `sync.Map` grows to `b.N` entries. So the `374 → 399` in the description compares two different workloads. With `b.N` pinned (`-benchtime=1000000x`) the miss path is consistently slower, not noise — the extra `Load` acquires `m.mu` a second time once the read table is amended:

```
Hit    72.0 ns  160 B  2 allocs  ->  33.1 ns    0 B  0 allocs
Miss    361 ns  265 B  3 allocs  ->   425 ns  265 B  3 allocs   (+18%)
```

That same benchmark also peaks at 2.8 GB RSS at `-benchtime=3s` and grows without bound.

Fixed-work version in the two suggestions below (apply together — the first alone won't compile). Drift drops to 4.8% across the same 15x `b.N` range, peak RSS to 13 MB, and it keeps the result that actually justifies the change: 2048 -> 0 allocs on the hit path.

Worth reframing the description around allocations rather than time. Over a `copyGraph`-shaped walk (1024 nodes, one miss to claim + one hit in the parent wait loop) I measured 5487 -> 3439 allocs and 436 -> 272 KB, with wall-clock a wash at ~220 us. The parallel hit path is the strongest number here and isn't currently covered: 42 ns/2 allocs -> 6 ns/0 allocs at 8 goroutines.

`tracker.go` itself looks correct to me — identical `!exists` semantics, the race handled by leaving `LoadOrStore` in place, and nothing ever deletes from `t.status` so a `Load` hit can't go stale. I also measured an `RWMutex` + typed map alternative: faster serially on both paths, but 80-95 ns/op on parallel hits vs 6 — so `sync.Map` is the right structure and this is the right fix for it.

TerryHowe (approved):
/lgtm

PR #1437, opened 2026-09-22
perf(auth): cache the redirect-safe client instead of rebuilding it per send

jadhavgaurav:
## What

`redirectSafeClient` (`registry/remote/auth/client.go:151`) copies the underlying `http.Client` struct and builds a fresh `CheckRedirect` closure on every call. `send` invoked it on every single request, so one `Do` call — which can issue up to three `send`s (a cached-token attempt, the challenge round trip, and a retry with a freshly fetched token) — paid that cost three times, and a copy operation issuing many `Do` calls paid it many times over. The wrapper only depends on the `Client`'s configured `http.Client` and is otherwise stateless with respect to the request being sent, so none of that work needs to be repeated between calls.

## Change

`Client` now builds the wrapper once, on the first request sent through it, and reuses it for every later request via `sync.Once`.

Adding a `sync.Once` field to `Client` is only safe as long as nothing copies a `Client` by value after it has been used — but `Login` (`registry/remote/auth.go`) already did exactly that, assigning `*client` into a local variable to get an independent copy it can attach a login-scoped `CredentialFunc` to without mutating the caller's original client. `go vet`'s `copylocks` check catches this correctly, since it can't prove that copying a `Client` whose `sync.Once` may have already fired is safe. `Client` gains an exported `Clone` method that copies only the exported configuration fields — the same pattern `Repository.clone` already uses in this package for the same reason — and `Login` now calls it instead of copying the struct directly. The clone does not inherit the original's cached wrapper; it builds its own the first time it sends a request.

## Tests

`client_test.go` gains a benchmark and two correctness tests:

- `BenchmarkClient_send`: measures `send`'s own overhead using a stub `RoundTripper` instead of a loopback server, so the numbers aren't dominated by real network I/O.
- `TestClient_send_ConcurrentFirstUse`: 64 goroutines call `send` on a single freshly-constructed `*Client` at once, so the lazy construction of the cached wrapper races across all of them. Run with `-race`, this exercises the new caching under real concurrent first use.
- `TestClient_Clone`: verifies `Clone` copies the exported fields, that mutating the clone's `CredentialFunc` doesn't affect the original, and that a clone does not inherit an already-built cached wrapper.

`auth_test.go` gains `TestLogin_customClientNotMutated`, exercising the branch of `Login` that reuses a caller-supplied `*auth.Client` (previously untested) and asserting the original client's `CredentialFunc` is unchanged afterward — the property `Login`'s doc comment already claims and that `Clone` now has to uphold.

```
before: BenchmarkClient_send   552 B/op   10 allocs/op
after:  BenchmarkClient_send   488 B/op    8 allocs/op
```

-2 allocs/op, matching exactly the two removed per-call allocations: the copied `http.Client` struct and the `CheckRedirect` closure. Verified by reverting only `client.go`'s change (source only, `git apply -R` on a saved patch, test file untouched) and rerunning the benchmark: it reproduces the original 552 B/op, 10 allocs/op, then reapplying the patch restores the fixed numbers.

## Checks

Run locally (this checkout has the same pre-existing, unrelated `go.mod`/`vendor/modules.txt` mismatch noted in #1418, #1428, and #1436, so `-mod=mod` was used instead of regenerating `vendor/`):

- `golangci-lint` is not installed locally; per the repo's own config it enables only `govet`, `ineffassign`, and the `gofmt` formatter. `govet`/`gofmt` were run directly above; `ineffassign` was run via `GOFLAGS=-mod=mod go run github.com/gordonklaus/ineffassign@latest ./...` and is clean.

## Scope

This addresses finding 10 of https://github.com/oras-project/oras-go/issues/1395, the last of the ten findings. Part of #1395; this issue is not closed here since it tracks the audit as a whole, not any single finding.


registry/remote/auth/client.go
```diff
@@ -77,6 +78,10 @@ type Client struct {
 	// is already available in the DefaultClient.
 	// It is also possible to use a custom client. For example, github.com/hashicorp/go-retryablehttp
 	// is a popular HTTP client that supports retries.
+	//
+	// Client is read once, on the first request sent through this Client, to
+	// build the redirect-safe wrapper used for every subsequent request.
+	// Changing Client after the first request has no effect.
```
TerryHowe, line 84: Also frozen: in-place mutation of the `*http.Client` this points to. Verified on this branch — a `Transport` swap after the first `send` is silently ignored, same as reassigning the field.

```suggestion
	// The Client field and the *http.Client it points to are both read once,
	// on the first request sent through this Client, to build the
	// redirect-safe wrapper used for every subsequent request. After that
	// first request, neither reassigning Client nor mutating the fields of
	// the http.Client it points to (Transport, Timeout, Jar, CheckRedirect)
	// has any effect. Configure the client fully before first use, or derive
	// a separately-configured one with Clone.
```

registry/remote/auth/client.go
```diff
@@ -118,6 +123,36 @@ type Client struct {
...
+	// redirectSafeClientValue caches the client returned by
+	// redirectSafeClient(c.client()). The wrapper is stateless with respect
+	// to the request being sent, so it only needs to be built once per
+	// Client rather than on every send.
+	redirectSafeClientValue *http.Client
+}
+
+// Clone returns a shallow copy of c's exported configuration, safe to modify
+// independently of c — for example to attach a different CredentialFunc
+// before use.
```
TerryHowe, line 140: It's shallow, so `Header`, `Cache` and `Client` stay shared — only field assignment is independent.

```suggestion
// Clone returns a shallow copy of c's exported configuration. Assigning to the
// clone's fields does not affect c — for example to attach a different
// CredentialFunc before use — but the Header map and the Cache and Client
// values it references are shared with c, so mutating those through the clone
// is visible to c as well.
```

registry/remote/auth/client_test.go
```diff
@@ -4564,3 +4567,164 @@ func TestClient_fetchBearerToken_RealmRedirect(t *testing.T) {
...
+	}
+}
+
+// TestClient_Clone verifies that Clone copies a Client's exported
+// configuration but not its internal, once-only cache state, and that the
+// clone is independent of the original: mutating the clone's fields must not
+// affect the original, which is the property Login relies on.
+func TestClient_Clone(t *testing.T) {
+	original := &Client{
+		Client:   &http.Client{Timeout: 5},
```
TerryHowe, line 4632: `Timeout: 5` is 5ns; only the pointer identity matters here.

```suggestion
		Client:   &http.Client{},
```

registry/remote/auth/client_test.go
```diff
@@ -4564,3 +4567,164 @@ func TestClient_fetchBearerToken_RealmRedirect(t *testing.T) {
...
+	}
+	if original.redirectSafeClientValue == nil {
+		t.Fatal("original did not build its cached redirect-safe client")
+	}
+
+	freshClone := original.Clone()
+	if freshClone.redirectSafeClientValue != nil {
+		t.Error("Clone() copied the original's cached redirect-safe client instead of leaving it unbuilt")
+	}
+}
```
TerryHowe, line 4694: Guards `Clone` against a field added later being silently dropped. Verified: removing `TokenFetcher:` from `Clone` fails this while `TestClient_Clone` still passes.

```suggestion
}

// TestClient_Clone_copiesEveryExportedField fails when an exported field is
// added to Client but not to Clone, which would otherwise be a silent drop:
// Login would quietly stop carrying that field onto its local client. It sets
// every exported field to a non-zero value, clones, and asserts none came back
// zero.
func TestClient_Clone_copiesEveryExportedField(t *testing.T) {
	original := &Client{
		Client: &http.Client{},
		Header: http.Header{"X-Test": {"v"}},
		CredentialFunc: func(context.Context, properties.Resource) (credentials.Credential, error) {
			return credentials.EmptyCredential, nil
		},
		Cache:        DefaultCache,
		ClientID:     "original-id",
		TokenFetcher: NewCompositeTokenFetcher(nil, nil, "", false),
	}

	originalValue := reflect.ValueOf(original).Elem()
	cloneValue := reflect.ValueOf(original.Clone()).Elem()
	clientType := originalValue.Type()
	for i := range clientType.NumField() {
		field := clientType.Field(i)
		if !field.IsExported() {
			continue
		}
		if originalValue.Field(i).IsZero() {
			t.Fatalf("test needs updating: original.%s is zero, so Clone cannot be checked for it", field.Name)
		}
		if cloneValue.Field(i).IsZero() {
			t.Errorf("Clone() did not copy exported field %s", field.Name)
		}
	}
}
```

TerryHowe (commented):
A few suggestions. All four applied together locally: `gofmt`, `go vet` and `go test ./registry/remote/... -race` clean.

TerryHowe (approved):
/lgtm

